PR #11, opened 2022-01-22
Feature/cli with cmd line parser

Ma233:
Related issue: https://github.com/BNSnet/bns-node/issues/7

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,5 @@
 use bns_node::discoveries::http::remote_handler;
 use std::net::SocketAddr;
-use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -15,10 +15,19 @@ use webrtc::ice_transport::ice_candidate::RTCIceCandidate;
 use webrtc::peer_connection::math_rand_alpha;
 use webrtc::peer_connection::peer_connection_state::RTCPeerConnectionState;
 
+#[derive(Parser, Debug)]
+#[clap(about, version, author)]
+struct Args {
+    #[clap(long, short = 'h', default_value = "0.0.0.0:60000")]
```
Ma233, line 21: We'd better use another short name here to avoid conflict with `--help`.

Ma233 (commented):
Please tell me if there should be more configs for the cli.

PR #17, opened 2022-01-28
inge transport and channel

RyanKung:


RyanKung:
TODO: 
1. mv callbacks from main to transports/default_callback.rs
2. CallbackTrait
3. Wasm impl

PR #38, opened 2022-02-18
Refine: wasm transport, pure rust signing module and lockless mpmc

RyanKung:


RyanKung:
1. rm mutex on transport::signaler
2. use async_channel instead of tokio::mpsc
3. fixed wasm transport

PR #42, opened 2022-02-21
Refactor/split service and drive by axum

Ma233:


src/main.rs
```diff
@@ -81,7 +55,11 @@ async fn run(localhost: &str, key: SecretKey, stun: &str) {
 async fn main() -> Result<()> {
     let args = Args::parse();
     Logger::init(args.log_level)?;
+
+    // 放到 Args::parse 里
```
RyanKung, line 59: Comments should be English

src/config.rs
```diff
@@ -1,25 +0,0 @@
-use serde_derive::Deserialize;
```
RyanKung, line 1: Can we share same Config struct between toml parser and cl-arg parser?

Ma233: Clap does not support that. Try https://direnv.net/ with environment variables.

PR #43, opened 2022-02-21
Swarm and Channel related

RyanKung:


RyanKung:
1. Should not clone Swarm, pass Arc<Swarm> instead.
2. Refined Channel trait.
3. Redirect msg to channel for testing.

bns-core/src/channels/default.rs
```diff
@@ -12,7 +12,7 @@ pub struct AcChannel {
     receiver: Arc<Receiver<Events>>,
 }
 
-#[async_trait(?Send)]
+#[async_trait]
 impl Channel for AcChannel {
     type Sender = Arc<Sender<Events>>;
     type Receiver = Arc<Receiver<Events>>;
```
Ma233, line 18: https://docs.rs/async-channel/latest/async_channel/struct.Sender.html
https://docs.rs/async-channel/latest/async_channel/struct.Receiver.html

> Senders can be cloned and shared among threads. When all senders associated with a channel are dropped, the channel becomes closed.
> Receivers can be cloned and shared among threads. When all receivers associated with a channel are dropped, the channel becomes closed.

Sender / Receiver 可以直接 clone，不需要 Arc

Ma233: 我看了下 crossbeam_channel 的文档，好像也是可以直接 clone 的

PR #48, opened 2022-02-27
Chore/fixups in swarm and default transport

Ma233:


bns-core/src/transports/default.rs
```diff
@@ -319,13 +273,7 @@ impl IceTransportCallback<AcChannel> for DefaultTransport {
...
     }
 
-    async fn on_peer_connection_state_change_callback(
-        &self,
-    ) -> Box<
-        dyn FnMut(RTCPeerConnectionState) -> Pin<Box<dyn Future<Output = ()> + Send + 'static>>
-            + Send
-            + Sync,
-    > {
+    async fn on_peer_connection_state_change_callback(&self) -> OnPeerConnectionStateChangeHdlrFn {
```
RyanKung, line 276: Should be same as the def in trait. Maybe we can create a new type-alias in types::ice_transport.rs file.

bns-core/src/transports/default.rs
```diff
@@ -337,18 +285,19 @@ impl IceTransportCallback<AcChannel> for DefaultTransport {
...
+    async fn on_data_channel_callback(&self) -> OnDataChannelHdlrFn {
+        let channel = self.get_data_channel().await.unwrap();
+        let signaler = self.signaler();
+
         box move |d: Arc<RTCDataChannel>| {
-            let recv = Arc::clone(&d);
+            let channel = Arc::clone(&channel);
+            let signaler = Arc::clone(&signaler);
+
             Box::pin(async move {
```
RyanKung, line 296: Why not on_open(self.some_handler_fn), why remove `callback` fn from trait?

bns-core/src/transports/default.rs
Ma233: @RyanKung Done with feature `#![feature(associated_type_defaults)]`

bns-core/src/transports/default.rs
Ma233: Since the `IceTransportCallback` trait cannot constraint the usage place of `on_open_callback` and `on_message_callback`. I currently place those logics in `on_data_channel_callback` to avoid ambiguous invokes.

PR #50, opened 2022-03-02
stabilize dht through data_channel

Raytlty:
use stabilize periodly, find successor and  update finger_table


bns-core/src/types/ice_transport.rs
```diff
@@ -21,7 +22,7 @@ pub trait IceTransport<Ch: Channel> {
     type ConnectionState;
     type Msg;
 
-    fn new(signaler: Arc<Ch>) -> Self;
+    fn new(signaler: Arc<Ch>, procedures: Arc<Procedures>) -> Self;
```
RyanKung, line 25: Transport is just ONLY for handling connection. Concrete logic should NOT attach to it.

bns-core/src/dht/message.rs
```diff
@@ -0,0 +1,30 @@
+use crate::channels::default::AcChannel;
+use crate::dht::chord::Chord;
+use crate::dht::chord::ChordAction;
+use std::sync::{Arc, Mutex as StdMutex};
+use webrtc::data_channel::RTCDataChannel;
+
+#[derive(Clone, Debug)]
+pub struct Procedures {
+    pub routing: Arc<StdMutex<Chord>>,
```
RyanKung, line 9: `Procedures` or `Successor` is NOT real role, they are virtual and should not include `Chord`, BTW, `Chord` is not routing, it's an algorithm helper and to manage statement for RING.

All function associated to `Procedures` or `Successor` should be implemented under Swarm.

bns-core/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 #![feature(associated_type_defaults)]
 #![feature(async_closure)]
 #![feature(box_syntax)]
+#![allow(dead_code)]
```
RyanKung, line 4: Pls, Never do this, if a PR should fail CI, let it failed.

bns-core/src/types/channel.rs
```diff
@@ -21,3 +23,17 @@ pub trait Channel {
     async fn send(&self, e: Events) -> Result<()>;
     async fn recv(&self) -> Result<Events>;
 }
+
+// impl Serialize & Deserialize ChordAction to Vec<u8>
+//
+impl Events {
+    pub fn from_action(action: ChordAction) -> Events {
```
RyanKung, line 30: Use `std::From` instead. And it should be a `TryFrom`.

RyanKung (changes requested):
Cant merge yet

bns-core/src/dht/message.rs
Raytlty: `it's an algorithm helper and to manage statement for RING`
why chord cant use for ROUTING

Raytlty:
close this branch, will push another checkout from features/response-msg

PR #51, opened 2022-03-02
Features/some modify for bns ext

RyanKung:


RyanKung:
1. impl `fn rand` for SecretKey.
2. `std::time` will cause panic on wasm runtime, replace with `chrono`.
3. A Wrapper for web-sys::Sdp types, because of `Send + Sync` is not impl on RtcSDP.

PR #54, opened 2022-03-03
response msg

RyanKung:


RyanKung:
Once a Peer receive a `DHTMessage`, it can cause two situations:

1. Found some did, and return with `DHTResponse`.

```
#[derive(Deserialize, Serialize, Debug, Clone)]
pub struct DHTResponse {
    act: RemoteAction,
    req_sig: Vec<u8>,
    sdp: Option<Encoded>,
}
```

DHT response is for handling `ResponseMsg` From DHT Peer. It simply wrapped a `DHTAction` with addition info 
such as `req_sig` and `sdp` for transactions management and handshake connection.



3. Not Found, continue to send `DHTRelayMessage` to next peer.

```
#[derive(Deserialize, Serialize, Debug, Clone)]
pub struct DHTRelayMessage {
    act: RemoteAction,
    req_sig: Vec<u8>,
    from: vec<Address>
}

#[derive(Deserialize, Serialize, Debug, Clone)]
pub struct DHTRelayResponse {
    act: DHTResponse,
    from: Vec<Address>
}
```

Consider Peer A send `DHTRelayMessage` `msg_a` with push `self.id` to `msg_a`.from to B, 
then B send `DHTRelayMessage` `msg_b` to C and C got some value, it's returning
`ResponseMsg<RelayMsg>` resp_c to B

Once B received resp_c, B should send `ResponseMsg<RelayMsg>` to resp_c.act.from[-1],
Once B received resp_c, B should send `ResponseMsg<RelayMsg>` to resp_c.act.from[-2],
Once B received resp_c, B should send `ResponseMsg<RelayMsg>` to resp_c.act.from[-3],

So `DHTRelayMessage` has two associated method: 

`record`: record self.id to relay chain.
`prev`: pop and get previous hop for msg routing back

RyanKung:
`DHTRelayMsg`, `DHTRelayResp` is ugly, needs to be refined.

Ma233:
@RyanKung @Raytlty 

Implement MessageHandler, Message and MSRP for message exchange.
Some memory optimization for address parameter passing.
Add concurrent safely get_or_set for memory storage.

bns-core/src/routing/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod did;
```
RyanKung, line 1: I dont think Did is belongs to `routing`

Ma233: @RyanKung Changed `routing` to `dht`.

PR #60, opened 2022-03-16
Featuress/stabilize

Raytlty:
Add stabilize and fix_finger to background services


src/service/stabilization.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::Result;
```
RyanKung, line 1: Mv it to `bns-core`

bns-core/src/ecc.rs
```diff
@@ -135,6 +137,24 @@ fn do_verify(message: &str, address: &Address, sig_bytes: &SigBytes) -> Result<b
     Ok(&pubkey.address() == address)
 }
 
+pub fn encrypt_with_sha1(inputs: &str) -> String {
```
RyanKung, line 140: hash function is not `encrypt` function.

RyanKung (approved):

Ma233 (approved):

PR #72, opened 2022-03-23
Refactor/channel enhancement

Ma233:
1. If the transport only sends events, then only a sender is needed.
2. Abstract channel to use it in message handler.

### DIFF
--- a/bns-core/src/types/channel.rs
+++ b/bns-core/src/types/channel.rs
@@ -1,4 +1,3 @@
-use anyhow::Result;
 use async_trait::async_trait;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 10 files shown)


bns-core/src/types/ice_transport.rs
```diff
@@ -9,16 +8,16 @@ use web3::types::Address;
 
 #[cfg_attr(feature = "wasm", async_trait(?Send))]
 #[cfg_attr(not(feature = "wasm"), async_trait)]
-pub trait IceTransport<Ch: Channel> {
+pub trait IceTransport {
```
RyanKung, line 11: Sender should be a constraint type. For an ideal implementation, all associated type should be  rewritten with some `trait type`. And for current situation, I think `EventSender` should have its trait binding, otherwise, `Channel` trait is making no sense.

bns-core/src/types/channel.rs
```diff
@@ -7,15 +6,16 @@ pub enum Event {
     ReceiveMsg(Vec<u8>),
 }
 
+pub trait ChannelSender {}
```
RyanKung, line 9: maybe

```
pub trait ChSender {
    async send()...
}
```

PR #77, opened 2022-03-24
Bugfix for wasm

RyanKung:


RyanKung:
for Chrome, username_fragment can be None

RyanKung:
1. Fix `IceCandidate` may lose info about `mid`.
2. Added help function `cargo wasm`, and `cargo wasm_test`
3. A test scaffold for `wasm_bindgen`

PR #83, opened 2022-03-28
Features/more wasm tests

RyanKung:


RyanKung:
1. Rename `stun_server` to `ice_server`
2. `UsernameFragment` is not supported by `web-sys`, force hack it.
3. Default value of `sdp_mid` should be `0`
4. Re org tests of wasm.

PR #84, opened 2022-03-29
Features/more wasm tests

RyanKung:


RyanKung:
Add a raw `web-rtc` test (copy from https://rustwasm.github.io/docs/wasm-bindgen/examples/webrtc_datachannel.html), for `Browser Compatibility Testing`

PR #86, opened 2022-03-31
add bns-core error handle

Raytlty:
error refine
1. remove anyhow in bns_core lib, and use thiserror instead
2. bns-node can still use anyhow in application level


googolmo:
Cool

PR #98, opened 2022-04-09
update test_message_handler unittest

Raytlty:
add test_message_handler_unittes alpha version

### DIFF
--- a/bns-core/src/dht/chord.rs
+++ b/bns-core/src/dht/chord.rs
@@ -70,7 +70,7 @@ impl Chord {
                         // for a existed value v
                         // if id is more close to self.id than v
                         if id - self.id < v - self.id {
-//                        if id < v || id > -v {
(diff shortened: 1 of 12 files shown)


RyanKung:
image

N1 is more close to N8 than N2, and N63 is more close to N8 than N1

So the aloghrithm should be:

```python
In [2]: from klefki.algebra.fields import PrimeField

In [3]: class Chord(PrimeField):
   ...:     P = 64
   ...:

In [4]: Chord(1) - Chord(8)
Out[4]: Chord::0x39

In [5]: Chord(2) - Chord(8)
Out[5]: Chord::0x3a

In [6]: Chord(63) - Chord(8)
Out[6]: Chord::0x37
```



```python
In [18]: Chord(42)- Chord(8) < Chord(7) - Chord(8)
Out[18]: True

In [19]: Chord(42) < Chord(7)
Out[19]: False
```

PR #114, opened 2022-04-20
Node should sync it's vnode with successor

RyanKung:


RyanKung:
ref: https://github.com/BNSnet/bns-node/issues/87

PR #147, opened 2022-05-22
Features/more unitest and bugfix

RyanKung:


RyanKung:
related: https://github.com/RingsNetwork/rings-node/issues/145

PR #150, opened 2022-05-23
add stabilize into daemon

Raytlty:


bin/daemon.rs
```diff
@@ -133,9 +141,10 @@ async fn run_jobs(args: &RunArgs) -> anyhow::Result<()> {
     };
 
     let ice_servers = ice_servers.join(";");
-    let swarm = Arc::new(Swarm::new(&ice_servers, key.address(), session));
+    let swar = Arc::new(Swarm::new(&ice_servers, key.address(), session));
```
googolmo, line 144: typo ?

RyanKung:
`run_stablization` should be moved to `rings-core`, and contains a WASM implementation.

PR #154, opened 2022-05-26
fix msg receive in wasm

googolmo:


RyanKung:
try: https://crates.io/crates/console_log

PR #161, opened 2022-06-02
new structure of relay

Ma233:
The new relay fields:
```rust
{
    // A push only stack. Record routes when sending message.
    path: Vec<Did>,

    // Move this cursor to flag the top of stack when reporting.
    path_end_cursor: usize,

    // The next node to handle the message. Current handler will pick transport by this field.
    // When and only when located at the end of the message propagation, the next_hop is none.
    next_hop: Some<Did>,

    // The final destination of the message. May be customized when sending. Cannot change when reporting.
    // It may help handler to find out next_hop in some situations.
    destination: Did,
}
```

To **send** a message, you must provide both next_hop and destination.
```rust
    pub async fn send_message(
        &self,
        next_hop: &Did,
        destination: &Did,
        msg: Message,
    ) -> Result<()> {...}
```

To **send** or **report** a message, you can also use send_relay_message. Usually, use in relay processing.
```rust
    pub async fn send_relay_message(&self, msg: MessageRelay<Message>) -> Result<()> {...}
```

### DIFF
--- a/rings-core/src/dht/stabilization.rs
+++ b/rings-core/src/dht/stabilization.rs
@@ -42,10 +42,11 @@ impl Stabilization {
             Message::NotifyPredecessorSend(NotifyPredecessorSend { id: chord.id }),
             &self.swarm.session_manager,
             OriginVerificationGen::Origin,
-            None,
-            None,
-            None,
             MessageRelayMethod::SEND,
+            None,
+            None,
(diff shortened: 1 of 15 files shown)


RyanKung (approved):

RyanKung:
GREAT！

PR #167, opened 2022-06-06
update redis cache

Raytlty:


rings-core/src/storage/cache.rs
```diff
@@ -0,0 +1,109 @@
+use crate::err::{Error, Result};
```
RyanKung, line 1: 1. I think it is *not* cache, it is persistence/semi-persistence storage, maybe the path should be `storage/persistence`
2. `RedisStorage` is only for Native build, it should be flagged as `#[cfg(not(features==`wasm`))`

rings-core/src/storage/cache.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn max_size(&self) -> Result<Option<u64>>;
+}
+
+impl RedisCache {
+    /// Create a new `RedisCache`.
+    pub fn new(url: &str) -> Result<RedisCache> {
+        let mut parsed = Url::parse(url)?;
+        // If the URL has a password set, mask it when displaying.
+        if parsed.password().is_some() {
+            let _ = parsed.set_password(Some("bns-redis-cache"));
```
RyanKung, line 42: r/bns/rings

rings-core/src/storage/cache.rs
```diff
@@ -0,0 +1,109 @@
...
+        let cache_key = key.to_string();
+        let mut con = self.connect()?;
+        let _: () = redis::pipe()
+            .atomic()
+            .set(&cache_key, &entry)
+            .expire(&cache_key, 60)
+            .query(&mut con)?;
+        Ok(())
+    }
+
```
RyanKung, line 86: We need a method to reduce storage size.

rings-core/src/storage/cache.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Returns the current cache size. This value is aquired via
+    /// the Redis INFO command (used_memory).
+    fn current_size(&self) -> Result<Option<u64>> {
```
googolmo, line 89: This function could NOT be used outside, Because the type of `K` is not known.

rings-core/Cargo.toml
```diff
@@ -40,6 +40,9 @@ thiserror = "1"
 flate2 = { version = "1.0.22" }
 async-recursion = "1.0.0"
 itertools = "0.10.3"
+redis = { version = "0.21" }
```
googolmo, line 43: These two dependencies should set  `optional = true`, and only for default feature.

Cargo.toml
```diff
@@ -60,6 +60,8 @@ http = { version = "0.2.6" }
 
 # client
 tokio = { version = "1.13.0", features = ["full"], optional = true }
+tempfile = { version = "3.2", optional = true }
```
googolmo, line 63: These two dependencies can be removed.

rings-core/src/storage/mod.rs
```diff
@@ -2,6 +2,14 @@ mod memory;
 pub mod persistence;
 
 pub use memory::MemStorage;
+#[cfg(feature = "default")]
```
googolmo, line 5: I think maybe use `#[cfg(not(feature = "wasm"))]` is better.

rings-core/src/storage/persistence/redis.rs
```diff
@@ -0,0 +1,396 @@
...
+use redis::Value;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use serde_json::{self};
+
+use crate::err::Error;
+use crate::err::Result;
+
+/// Persistence Storage Operations
+#[cfg(not(features = "wasm"))]
```
googolmo, line 24: Maybe this line can be removed.

RyanKung (changes requested):
Redis support should be providing as a feature.
```
#[cfg(features = "redis")]
```

PR #169, opened 2022-06-08
Add Persistence storage

googolmo:
use IndexedDB as storage for WASM package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ docs
 .DS_Store
 
 pkg/
+temp/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -591,6 +591,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-epoch"
+version = "0.9.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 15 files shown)


rings-core/src/storage/indexed_db.rs
```diff
@@ -0,0 +1,198 @@
+#![warn(missing_docs)]
```
RyanKung, line 1: We can organize wasm & native modules just like

file: `src/stroage/persistence.rs`

```rust

#[cfg(feature = "wasm")]
mod storage {
    struct PersStorage {}

}

#[cfg(not(feature = "wasm"))]
mod storage {
    struct PersStorage {}

}
```

RyanKung: or 

```
#[cfg(feature = "wasm")]
pub use wasm::IndexDBStroage as Stroage;
#[cfg(not(feature = "wasm"))]
pub use default::RedisStroage as Stroage;

#[cfg(feature = "wasm")]
mod wasm {
    struct PersStorage {}

}

#[cfg(not(feature = "wasm"))]
mod default {
    struct RedisStorage {}
}
```

RyanKung (approved):

PR #171, opened 2022-06-08
TransportManager do not need persistent storage

Ma233:


RyanKung (approved):

RyanKung (changes requested):
Just rename `Storage` trait to `TMemStorage` is fine, or remove `Storage` trait and ` impl l<K, V>  MemStorage<K, V> `

RyanKung (approved):

PR #224, opened 2022-07-16
Features/bugfix for conn

RyanKung:


RyanKung:
1. Fixed `IceServer` list.
2. Support mDNS query and gather.
3. Support setting external_ip address.
4. Handshake info should always contain candidate info.

bin/node.rs
```diff
@@ -222,7 +226,16 @@ async fn daemon_run(
     )?;
     let sig = key.sign(&auth.to_string()?).to_vec();
     let session = SessionManager::new(&sig, &auth, &temp_key);
-    let swarm = Arc::new(Swarm::new(stuns, key.address(), session.clone()));
+    let swarm = if let Some(eip) = external_ip {
```
googolmo, line 229: Maybe here can call swarm::new_with_external_ip(...,external_ip) directly.

Ma233 (approved):
LGTM

bin/node.rs
Ma233: We may unify external interface of core mod in another pr. Then we can add a `SwarmBuilder` for the `Swarm` interface.

PR #225, opened 2022-07-18
support external ip and conf path

RyanKung:


bin/daemon.rs
```diff
@@ -146,7 +152,12 @@ async fn run_jobs(args: &RunArgs) -> anyhow::Result<()> {
     };
 
     let ice_servers = ice_servers.join(";");
-    let swarm = Arc::new(Swarm::new(&ice_servers, key.address(), session));
+    let swarm = Arc::new(Swarm::new_with_external_address(
+        &ice_servers,
+        key.address(),
+        session,
+        args.external_ip,
```
googolmo, line 159: `args.external_ip.clone()`

googolmo (approved):

PR #243, opened 2022-07-25
add build_version for browser

googolmo:
## How to use
```typescript
import { internal_info } from 'rings-node'
const info = internal_info()
console.log(`build_version: ${info.build_version}`)
```


Ma233 (approved):
LGTM

PR #247, opened 2022-07-25
Try to fill finger table after remove did

Ma233:


rings-core/src/dht/finger.rs
```diff
@@ -71,9 +71,30 @@ impl FingerTable {
...
+            let fix_id = self
+                .finger
+                .iter()
+                .skip(end_idx)
+                .take(1)
+                .collect::<Vec<_>>()
+                .first()
+                .unwrap_or(&&None)
+                .as_ref()
+                .copied();
```
Ma233, line 94: Update with next did of last equaled did in finger table. If cannot get that, use `None`.

RyanKung (approved):

PR #254, opened 2022-07-31
Bugfix/already connect

RyanKung:
1. Do not register transport when ConnectNodeReport
 2. Impl Drop for WASMTransport, connection should be closed before drop.


RyanKung:
1. Do not register transport when `ConnectNodeReport`
2. Impl Drop for WASMTransport, connection should be closed before drop.

googolmo (approved):

PR #256, opened 2022-08-07
Features/sig with ed25519

RyanKung:


RyanKung:
1. Support Session auth with ed25519 / EdDSA
2. Session key sign msg should only use pureECDSA.
3. Support deserialize pubkey from trezor style b58 (also used by solana) and monero style b58

rings-core/src/ecc/signers.rs
```diff
@@ -106,4 +131,31 @@ mod test {
         assert_eq!(pubkey.address(), address);
         assert!(eip712::verify(msg, &address, &sig));
     }
+
+    #[test]
+    fn test_verify_ed25519() {
```
RyanKung, line 136: For Fe msg fmt, plz check this test @gaowhen

googolmo (approved):

PR #265, opened 2022-08-22
Add dummy transport

Ma233:


googolmo (approved):

RyanKung (changes requested):
1. Transport trait should not change
2. Too much unittest is removed.

RyanKung (approved):

PR #272, opened 2022-09-17
add feature node and set it to default

googolmo:
if you want to edit feature == browser, you can just set default = "browser" in Cargo.toml.

how to build ring binary

run

```cargo build-node```


Ma233 (approved):
LGTM

PR #274, opened 2022-09-18
vnode storage api

googolmo:


RyanKung (approved):

src/processor.rs
```diff
@@ -352,6 +355,27 @@ impl Processor {
...
+    /// fetch virtual node from DHT
+    pub async fn fetch(&self, id: &Did) -> Result<()> {
+        self.msg_handler
+            .fetch(id)
+            .await
+            .map_err(error::Error::VNodeError)
+    }
+
+    /// store virtual node on DHT
+    pub async fn store(&self, vnode: vnode::VirtualNode) -> Result<()> {
```
RyanKung, line 373: Maybe we should not export type `VirtualNode` to JS here, if we have to re-implement it.

googolmo: This function is not use by js directly.

PR #276, opened 2022-09-20
Add validator for message handler

Ma233:
* Add validator on message handler.
* Remove lock wrapper and mut func of callback and validator. Just pass them in when create message handler.

### DIFF
--- a/rings-core/src/message/handlers/mod.rs
+++ b/rings-core/src/message/handlers/mod.rs
@@ -49,7 +49,7 @@ pub trait MessageCallback {
 #[cfg_attr(feature = "wasm", async_trait(?Send))]
 #[cfg_attr(not(feature = "wasm"), async_trait)]
 pub trait MessageValidator {
-    async fn store_vnode(
+    async fn validate(
         &self,
         handler: &MessageHandler,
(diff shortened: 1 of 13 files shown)


rings-core/src/swarm.rs
```diff
@@ -155,6 +148,14 @@ impl Swarm {
         &self.session_manager
     }
 
+    pub fn message_handler(
```
RyanKung, line 151: Maybe `create_message_handler`? `message_handler` looks like a `getter`

rings-core/src/message/handlers/mod.rs
```diff
@@ -47,17 +46,34 @@ pub trait MessageCallback {
     async fn builtin_message(&self, handler: &MessageHandler, ctx: &MessagePayload<Message>);
 }
 
+#[cfg_attr(feature = "wasm", async_trait(?Send))]
+#[cfg_attr(not(feature = "wasm"), async_trait)]
+pub trait MessageValidator {
+    async fn store_vnode(
```
RyanKung, line 52: `MessageValidator` is not only for `vnode`.

RyanKung (approved):

PR #311, opened 2022-11-01
Features/warmup with documents

Raytlty:
1. remove eip712 signer because this is PersonSign kind sign.
2. mark `sha1` algorithm is not safe.[reason](https://sha-mbles.github.io/)
3. simple doc without doctest


rings-core/src/prelude.rs
```diff
@@ -1,4 +1,4 @@
-//! prelude
+//! Re-exports of important traits, types, and functions used with rin-core.
```
googolmo, line 1: Typo rings–core

googolmo (changes requested):

rings-core/src/ecc/signers.rs
```diff
@@ -33,24 +36,82 @@ pub mod default {
...
+        if let Ok(p) = recover(msg, sig) {
+            p.address() == *address
+        } else {
+            false
+        }
+    }
+}
+
+/// PersonSign mod.
+pub mod person_sign {
```
RyanKung, line 92: rename to EIP191

RyanKung: ref: https://eips.ethereum.org/EIPS/eip-191

rings-core/src/ecc/signers.rs
```diff
@@ -116,7 +180,7 @@ mod test {
     }
 
     #[test]
-    fn test_eip712_sign() {
```
RyanKung, line 119: we should keep both `test_eip712_*`, and `test_eip191_*`

RyanKung:
As I understand, the original EIP712 is EIP191, and the current person_sign method is actually EIP191.
So, should we support SIGN as EIP712? Or EIP191 is good enough?

rings-core/src/message/handlers/custom.rs
```diff
@@ -16,6 +16,7 @@ use crate::transports::manager::TransportManager;
 #[cfg_attr(feature = "wasm", async_trait(?Send))]
 #[cfg_attr(not(feature = "wasm"), async_trait)]
 impl HandleMsg<MaybeEncrypted<CustomMessage>> for MessageHandler {
+    ///
```
RyanKung, line 19: ？

rings-core/src/message/types.rs
```diff
@@ -10,100 +10,124 @@ use crate::err::Error;
 use crate::err::Result;
 
 #[derive(Debug, PartialEq, Eq, Deserialize, Serialize, Clone)]
+/// MessageType use to ask for connection, send to remote with transport_uuid and handshake_info.
```
RyanKung, line 13: Coding style

```rust
/// Comment
#[derive]
pub struct foo {
}
```

RyanKung (approved):

googolmo (approved):

PR #331, opened 2023-01-04
Trim data when reach the maximum length in VirtualNode

Ma233:


RyanKung (approved):

rings-core/src/dht/vnode.rs
```diff
@@ -160,10 +162,68 @@ impl VirtualNode {
...
             return Err(Error::VNodeDidNotEqual);
         }
+
+        let trim_num = max(
+            0,
+            (self.data.len() + other.data.len()) as i64 - VNODE_DATA_MAX_LEN as i64,
+        ) as usize;
+
+        let mut data = self.data.iter().skip(trim_num).cloned().collect::<Vec<_>>();
+        for e in other.data {
```
googolmo, line 172: [vec.extend_from_slice() ?](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice)

PR #341, opened 2023-02-01
Add measure with period counter

Ma233:


rings-core/src/measure.rs
```diff
@@ -0,0 +1,173 @@
+use std::sync::Mutex;
```
RyanKung, line 1: #![warn(missing_docs)]

RyanKung (approved):

PR #354, opened 2023-03-06
Automatically determine the next hop when sending a message

Ma233:


src/processor.rs
```diff
@@ -802,6 +801,7 @@ mod test {
         ) {
             let msg = handler.decrypt_msg(msg).unwrap();
             let text = unpack_text_message(&msg).unwrap();
+            dbg!(&text);
```
RyanKung, line 804: ：）

RyanKung:
Also, I think all three functions should be included: 

`send_direct_message`, `send_relay_message`, `and send_message`. The reasons are:

 1. There are scenarios where even if a direct connection is established, it may be necessary to pretend not to know and use relay to ensure privacy. 
 
2. There are scenarios where only direct interaction is allowed and third-party forwarding is not allowed, to ensure privacy. 

3. There are scenarios (most scenarios) where it doesn't matter how the message is delivered. The default can be set to send_message, but the other two interfaces should still be kept.

RyanKung (approved):

PR #392, opened 2023-04-04
CorrectChord: New stabilization algorithm for Correctness

RyanKung:
FACT

Related Issue: https://github.com/RingsNetwork/rings-node/issues/388

### DIFF
--- a/core/src/message/payload.rs
+++ b/core/src/message/payload.rs
@@ -141,7 +141,7 @@ where T: Serialize + DeserializeOwned
         Self::new(data, session_manager, OriginVerificationGen::Origin, relay)
     }
 
-    /// Check that the payload is expired
+    /// Checks whether the payload is expired.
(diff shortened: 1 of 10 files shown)


core/src/message/payload.rs
```diff
@@ -29,19 +29,22 @@ use crate::err::Result;
 use crate::session::SessionManager;
 use crate::utils::get_epoch_ms;
 
+/// Encode with gzip
```
Ma233, line 32: Please add comment in another pull request. (With enhanced grammar...)

Ma233 (changes requested):

core/src/message/payload.rs
RyanKung: Code commenting should be done as byproduct. We have a lot of codes needs to be comment, thus If we wait a pull request for them, we will never do it.

Ma233 (approved):

PR #407, opened 2023-04-23
Unified handshake procedure

Ma233:


core/src/types/ice_transport/mod.rs
```diff
@@ -32,6 +29,13 @@ pub struct IceCandidate {
     pub username_fragment: Option<String>,
 }
 
+/// Trickle payload contain webrtc sdp and candidates.
+#[derive(Debug, PartialEq, Eq, Deserialize, Serialize, Clone)]
+pub struct Trickle {
```
RyanKung, line 34: Maybe `HandshakeInfo` is better here. And, it's not trickle, actually the implementation is opposite to trickle, as trickle uses a negotiated approach to create connections, but we will wait for gathering to complete.

This is because, while handshaking over DHT, negotiate is too expensive.

core/src/transports/manager.rs
```diff
@@ -115,3 +142,125 @@ impl TransportManager for Swarm {
...
+
+        let offer_msg = ConnectNodeSend {
+            transport_uuid: trans.id.to_string(),
+            offer,
+        };
+
+        Ok((trans, offer_msg))
+    }
+
+    async fn answer_remote_transport(
```
RyanKung, line 166: remote_transport?, remote_node?

RyanKung (approved):

core/src/transports/manager.rs
```diff
@@ -115,3 +142,125 @@ impl TransportManager for Swarm {
...
     }
 }
+
+#[cfg_attr(feature = "wasm", async_trait(?Send))]
+#[cfg_attr(not(feature = "wasm"), async_trait)]
+impl TransportHandshake for Swarm {
+    type Transport = Arc<Transport>;
+    type Payload = MessagePayload<Message>;
+
+    async fn prepare_transport_offer(&self) -> Result<(Self::Transport, ConnectNodeSend)> {
```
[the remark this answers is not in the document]
RyanKung: And documentation?

RyanKung (approved):

PR #422, opened 2023-05-26
Wrap `wasm_bindgen` with `wasm_export`

RyanKung:


RyanKung:
Explanation

1. Why not using wasm_bindgen_macro_support: 

The purpose of this Pull Request (PR) is to avoid wrapping structs solely for the purpose of using the wasm_bindgen macro. Since wasm_bindgen does not compile in the not(target_arch=wasm32) scenario, a wasm_export macro is introduced here. In a non-wasm environment, the wasm_export macro does nothing.

2. Why not use wasm_bindgen_macro_support: 

- Although wasm_bindgen directly uses wasm_bindgen_macro_support::expand for wrapping, it involves complex processes such as using __wasm_bindgen_class_marker and a series of parsing work for class types. These processes can be quite intricate. Therefore, in this case, we choose to directly utilize wasm_bindgen instead of wasm_bindgen_macro_support. One side effect of this choice is that we often need to use both wasm_bindgen and wasm_export together, depending on whether the exported object is an extended value or a class.

PR #436, opened 2023-06-14
refactor!: init SessionManager by builder pattern

Ma233:

* Incorporate the SessionManager creation into core by providing `SessionManagerBuilder`.
* Simplify the `SessionManager` struct by removing its nested structure.
* Also remove ambiguous and reduplicate defined structs, such as `UnsignedInfo`, `SignerMode`, `Signer`, `AuthorizedInfo `. Now we only have `Session`, `SessionManager`, `SessionManagerBuilder` and an enum `Authorizer`.
* Change the name, `DEFAULT` signer to `Secp256k1`. Unified naming formats for all signers.

The creation of client is complicated since we support multiple types of signer.
And we have to construct an `UnsignedInfo` then dump it for user to sign.
Then we accept these two parameters (usigned_info and sig) separately.

The entire creation process will be expressed through `SessionManagerBuilder`. Including:
* parse authorizer_entity and authorizer_type
* generate random session key
* pack session content for signing
* set signature back

Yes, it's  a breaking change since we removed `UnsignedInfo`.
Wasm user should now create session by `SessionManagerBuilder` and then pass it to `Client` creation method.
Here is an example for secp256k1 type session:
```rust
    let key = SecretKey::random();
    let authorizer_entity = Did::from(key.address()).to_string();
    let authorizer_type = "secp256k1".to_string();

    let mut session_manager_builder = SessionManagerBuilder::new(authorizer_entity, authorizer_type);
    let sig = key.sign(&session_manager_builder.pack_session());
    session_manager_builder = session_manager_builder.sig(sig.to_vec());

    let client = browser::Client::new_client_with_storage_internal(
        session_manager_builder,
        stuns,
        None,
        storage,
    )
    .await;
```

Closes #issue


core/src/message/payload.rs
```diff
@@ -70,24 +70,26 @@ pub enum OriginVerificationGen {
...
     pub data: T,
-    /// The transaction ID of payload
+    /// The transaction ID of payload.
+    /// Remote peer should use same tx_id when response.
     pub tx_id: uuid::Uuid,
-    /// Address from payload authorizer
+    /// The did of last sender.
     pub addr: Did,
-    /// Relay messages
+    /// Guide message passing on rings network.
```
RyanKung, line 86: What is "guide message"

RyanKung: I think the naming can be similar to EVM. 

msg.tx / msg.sender / msg.origin

core/src/session.rs
```diff
@@ -1,17 +1,19 @@
...
-//! To avoid frequent signing, and keep the private key safe
-//! - ECDSA Session is based on secp256k1, which create a temporate secret key with one time signing auth
-//! - To create a ECDSA Session, we should generate the unsign_info with our pubkey (Address)
-//! - `SessionManager::gen_unsign_info(addr, ..)`, it will returns the msg needs for sign, and a temporate private key
-//! - Then we can sign the auth message via some web3 provider like metamask or just with raw private key, and create the SessionManger with
-//! - SessionManager::new(sig, auth_info, temp_key)
+//! This module provides a mechanism for node A to verify that the message received was sent by node B.
+//! It also allows node A to obtain the public key for sending encrypted messages to node B.
+//!
+//! If we have the private key of node B, we can easily implement it. Just regular actions about cryptograph.
```
RyanKung, line 8: Maybe just rm this two line

```text
//! If we have the private key of node B, we can easily implement it. Just regular actions about cryptograph. 
//! But, considering security factors, asking user to provide private key is not practical.
``

This is too much, We just need to explain that Rings should never "own" a sk.

File: core/src/session.rs [hunk 3]
```diff
@@ -1,17 +1,19 @@
...
+//! But, considering security factors, asking user to provide private key is not practical.
+//! On the contrary, we generate a temporary private key and let user sign it.
```
RyanKung, line 10: temporary private key  -> delegated private key / proxy private key

core/src/session.rs
```diff
@@ -24,281 +26,258 @@ use crate::err::Error;
...
 
-/// Authorizor with unique did and pubkey.
-#[derive(Deserialize, Serialize, PartialEq, Eq, Debug, Clone)]
-pub struct Authorizer {
-    /// Sha3 of a public key
-    pub did: Did,
-    // for ecdsa, it's hash of pubkey, and it can be revealed via `ecdsa.recover`
-    /// for ed25519' it's pubkey
-    pub pubkey: Option<PublicKey>,
+/// SessionManager holds the [Session] and its temporary private key.
```
RyanKung, line 54: temporary private key -> delegated private key / proxy private key

core/src/session.rs
```diff
@@ -24,281 +26,258 @@ use crate::err::Error;
...
     /// Timestamp when session created
     ts_ms: u128,
-    /// Did of session.
-    session_id: Did,
+    /// Signature to verify that the session was signed by the authorizer.
+    sig: Vec<u8>,
 }
 
-/// Session contain signature which sign with `Signer`, so need AuthorizedInfo as well.
+/// we support both EIP191 and raw ECDSA singing format.
```
RyanKung, line 90: r/singing/signing

RyanKung: Wrong doc, I think, maybe:

```
We will support as many protocols/algorithms as we can. For now, these include xxx, xxx, xxx. We welcome issues and PRs for additional implementations.
```

core/src/session.rs
```diff
@@ -24,281 +26,258 @@ use crate::err::Error;
...
-}
-
-/// Session with temp secretKey.
-#[derive(Debug, Clone)]
-struct SessionWithKey {
-    /// Session
-    session: Session,
-    /// The private key for session.
-    session_key: SecretKey,
+pub enum Authorizer {
```
RyanKung, line 92: I'm just wondering, since EIP191/BIP137/Secp256k1 are ECDSA, and Ed25519 is EdDSA, should we separate them? Sub enum or somehow?

core/src/session.rs
```diff
@@ -24,281 +26,258 @@ use crate::err::Error;
...
-        } else {
-            false
-        }
+        let now = utils::get_epoch_ms();
+        now > self.ts_ms + self.ttl_ms as u128
     }
 
     /// Verify session.
-    pub fn verify(&self) -> bool {
+    pub fn verify_self(&self) -> Result<()> {
```
RyanKung, line 198: I don't think present logic with Type is a good idea.

core/src/session.rs
```diff
@@ -24,281 +26,258 @@ use crate::err::Error;
...
-        let info = AuthorizedInfo {
-            signer,
-            authorizer,
-            session_id: key.address().into(),
-            ttl_ms: ttl.unwrap_or(Ttl::Some(DEFAULT_SESSION_TTL_MS)),
-            ts_ms: utils::get_epoch_ms(),
-        };
-        Ok((info, key))
+    /// Verify message.
+    pub fn verify(&self, msg: &str, sig: impl AsRef<[u8]>) -> Result<()> {
```
RyanKung, line 222: Same issue, should not use Type for presenting first-order logic.

RyanKung (approved):

PR #442, opened 2023-06-21
feat: support jsonrpc in browser

RyanKung:

feature

The interface implementation of Cli and Browser are separated and different.

Allow `browser::client` handle jsonrpc requests.

New method for `browser::client::Client`:

```rust
    pub fn request(
        &self,
        method: String,
        params: JsValue,
        opt_id: Option<String>,
    ) -> js_sys::Promise {
        ...
   }
```

Nop

Closes #issue

### DIFF
--- a/node/src/jsonrpc/handler.rs
+++ b/node/src/jsonrpc/handler.rs
@@ -2,7 +2,7 @@
 //! JSON-RPC handler for both feature=browser and feature=node.
 //! We support running the JSON-RPC server in either native or browser environment.
 //! For the native environment, we use jsonrpc_core to handle requests.
-//! For the browser environment, we utilize a [MessageHandler] to process the requests.
+//! For the browser environment, we utilize a Simple MessageHandler to process the requests.
 
 use core::future::Future;
 use std::pin::Pin;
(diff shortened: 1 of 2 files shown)


node/src/jsonrpc/handler.rs
```diff
@@ -0,0 +1,244 @@
...
+use crate::prelude::jsonrpc_core::Params;
+use crate::prelude::jsonrpc_core::Result;
+use crate::prelude::jsonrpc_core::Value;
+use crate::prelude::rings_rpc::method::Method;
+
+/// Type of handler function
+#[cfg(feature = "node")]
+pub type MethodFnBox = Box<
+    dyn Fn(Params, RpcMeta) -> Pin<Box<dyn Future<Output = Result<Value>> + Send>> + Send + Sync,
+>;
```
Ma233, line 29: Can also use `BoxFuture` and `LocalBoxFuture` here. Like what we did in transport default implementation.

Ma233 (approved):
Others LGTM

node/src/jsonrpc/handler.rs
RyanKung: Yes I know, I just don't want to import them just for two type alias. And I think this is more clearly.

PR #454, opened 2023-07-12
feat!: construct Processor with Config and Builder

### LINKED ISSUE #447: The implementation of the Node signature is incorrect.
### Describe the Bug

In Node/Processor:

```rust
impl Processor {
    /// Generate Signature for Authorization
    pub fn generate_signature(secret_key: &SecretKey) -> String {
        let message = format!("rings-node: {}", secret_key.address().into_token());
        let (signature, _recovery_id) = libsecp256k1::sign(
            &libsecp256k1::Message::parse(&keccak256(message.as_bytes())),
            secret_key,
        );
        base64::encode(signature.serialize())
    }

    /// verify signature
    /// will throw error when signature is illegal
    pub fn verify_signature(signature: &[u8], public_key: &PublicKey) -> Result<bool> {
        let message = format!("rings-node: {}", public_key.address().into_token());
        Ok(libsecp256k1::verify(
            &libsecp256k1::Message::parse(&keccak256(message.as_bytes())),
            &libsecp256k1::Signature::parse_standard_slice(
                &base64::decode(signature).map_err(|_| Error::DecodeError)?,
            )
            .map_err(|_| Error::DecodeError)?,
            &TryInto::<libsecp256k1::PublicKey>::try_into(*public_key)
                .map_err(|_| Error::DecodeError)?,
        ))
    }
    ...
}
```

The `generate_signature` function actually generates a signature for "rings-node: <public key>", which can be reused for replaying RPC calls.

### Expected Behavior

The message for signing should include `timestamp` and `request_id`.

### Actual Behavior

The signature is fixed, and the request/response can be replayed.

Ma233:

In order to reduce the differences in `Processor` initialization between native and browser, this PR provides `ProcessorBuilder` and `ProcessorConfig`.

`ProcessorConfig` should be a yaml string from user.
And `ProcessorBuilder::from_config` will accept that string to initialize `ProcessorBuilder`.
Then the programming context may set storage/measure/message_callback by its respective methods.
Finally use `ProcessorBuilder::build` method to construct a `Processor`.

This PR also defined session_manager serialization method `SessionManager::dump` for saving it in ProcessorConfig.

This PR also fixed the unsafe signature of rpcserver endpoint by using `SessionManager` to sign each request. See also #447.

User use variant ways to construct Swarm and Stabilization then make them `into` Processor in tests, rings bin and browser Client.

Use ProcessorConfig for configuration, use ProcessorBuilder for runtime setup.

Yes, it changed the initialization parameters of browser Client.

Closes #447


node/src/processor.rs
```diff
@@ -132,31 +169,6 @@ impl Processor {
 }
 
 impl Processor {
-    /// Generate Signature for Authorization
-    pub fn generate_signature(secret_key: &SecretKey) -> String {
```
RyanKung, line 136: This change is related to issue #447.
The implementation incorrect, but we should not remove it for not, it may cause security issues.

node/bin/rings.rs
```diff
@@ -367,19 +361,24 @@ struct InspectCommand {
...
-fn get_value<V>(value: Option<V>, default_value: V) -> V {
-    value.unwrap_or(default_value)
-}
-
 #[allow(clippy::too_many_arguments)]
 async fn daemon_run(args: RunCommand) -> anyhow::Result<()> {
-    let c = config::Config::read_fs(args.config_args.config)?;
+    let mut c = config::Config::read_fs(args.config_args.config)?;
 
-    let key = get_value(args.ecdsa_key, c.ecdsa_key);
```
RyanKung, line 378: "Accept a secret key directly, generated delegatedKey, then drop it."
-- I don't think the origin impl is that bad. It's fine if the sk is not holding by Client.
-- Just use & drop.

RyanKung: We can support both way for Client initialization, by sk directly, and dumped DKEY(SessionManager)

node/src/processor.rs
RyanKung: maybe simply sign with DKEY/DTOKEN(SessionManager)

node/src/browser/client.rs
```diff
@@ -83,11 +85,10 @@ pub struct Client {
 impl Client {
     /// Creat a new client instance.
     pub fn new_client(
-        session_manager: SessionManager,
-        stuns: String,
+        config: String,
```
RyanKung, line 88: what is this config..... 
ProcessorConfig ?

node/src/browser/client.rs
```diff
@@ -142,8 +120,25 @@ impl Client {
...
+            .await
+            .map_err(error::Error::Storage)?;
+
+        let ms = PersistenceStorage::new_with_cap_and_path(50000, measure_path)
+            .await
+            .map_err(error::Error::Storage)?;
+        let measure = PeriodicMeasure::new(ms);
+
+        let processor = Arc::new(
+            ProcessorBuilder::from_config(config)?
```
RyanKung, line 136: How to generate the Config?

RyanKung: Ok, I got it from unittest

```rust
    let config = serde_yaml::to_string(&ProcessorConfig {
        ice_servers: "stun://stun.l.google.com:19302".to_string(),
        external_address: None,
        session_manager: sm.dump().unwrap(),
        stabilize_timeout: 200,
    })
```
1. How to generate session manager (DKEY).
2. I think JSON is better than YAML here, for browser case. We can support both YAML and JSON.

RyanKung: 3. Support serialized config is great, but I think It's also necessary to support Config Struct.

core/src/swarm.rs
```diff
@@ -100,8 +100,8 @@ impl SwarmBuilder {
         self
     }
 
-    pub fn measure(mut self, implement: MeasureImpl) -> Self {
-        self.measure = Some(implement);
+    pub fn measure(mut self, implement: Option<MeasureImpl>) -> Self {
+        self.measure = implement;
```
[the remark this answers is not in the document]
RyanKung: BTW, you should add document here, we should add document for any new or modified code

node/src/processor.rs
Ma233: Agree that. Fixing now.

node/bin/rings.rs
Ma233: The generation progress is in init command. Here is just behavior that is easily misunderstood, it replace the key of config with the command line arguments, so that I removed this.

node/src/browser/client.rs
Ma233: Sure it's `ProcessorConfig`.
The original `Config` was moved to native module and should only be used in native mode.
The Client need indeed enhancement. Here I just temporarily make it work.

core/src/swarm.rs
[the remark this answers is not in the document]
Ma233: Agree to avoid using Option as input param (especially in build pattern). Fixing now.

node/src/browser/client.rs
Ma233: 1. To generate the SessionManager, use the exported SessionManagerBuilder.
2. JSON is also available for YAML loader.
3. I'm trying to make the method as less as possible.

core/src/session.rs
```diff
@@ -118,6 +118,19 @@ impl TryFrom<(String, String)> for Authorizer {
     }
 }
 
+// A SessionManager can be converted to a string using JSON and then encoded with base58.
+// To load the SessionManager from a string, use `SessionManager::from_str`.
+impl FromStr for SessionManager {
```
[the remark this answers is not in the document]
RyanKung: It's wrap of key, not wrap of session. The field session is used as a proof of the sk.
So, `SessionizeKey` if you want to undeline it can be verified by session. (I dont think it's a good idea, but it's ok)
Or, Delegatedkey if you want to underline how and what it's desined for.

core/src/session.rs
```diff
@@ -279,6 +292,13 @@ impl SessionManager {
     pub fn authorizer_did(&self) -> Did {
         self.session.authorizer_did()
     }
+
+    /// Dump session_manager to string, allowing user to save it in a config file.
+    /// It can be restored using `SessionManager::from_str`.
+    pub fn dump(&self) -> Result<String> {
+        let s = serde_json::to_string(&self).map_err(|_| Error::SerializeError)?;
```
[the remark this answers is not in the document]
RyanKung: Ok, I agree

node/bin/rings.rs
RyanKung: Get!

node/src/browser/client.rs
RyanKung: Ok

core/src/session.rs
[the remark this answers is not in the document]
Ma233: `DelegatedKey` sounds more understandable. 😆

Ma233:
@RyanKung Two commits are appended. PR description updated.

RyanKung (approved):
Looks great

PR #459, opened 2023-07-26
feat: limit connection by measurement

RyanKung:

- [X] Docs have been added / updated (for bug fixes / features)

feature

related issue https://github.com/RingsNetwork/rings-node/issues/399

In the current scenario, the DHT may attempt to establish a connection to did_a, but this can fail due to incorrect configuration or network issues. The existing implementation might repeatedly try to establish the connection, potentially causing the system's file descriptor limit to be reached very quickly.

1. A judgement for node behaviour, make sure node is behaviour `good`.
2. A measurement of node behavior, ensuring that the difference between disconnect and connect counts exceeds the THRESHOLD.
3. Throw a new error type: NodeBehaviourBad.
4. Record behaviour on function `connect`, `connect_via` and `disconnect`.

No

After this modify, if a node is behaviour no good, it will printout errors.

Screenshot 2023-07-26 at 6 51 21 PM

Closes #issue

### DIFF
--- a/core/src/measure.rs
+++ b/core/src/measure.rs
@@ -53,6 +53,10 @@ pub trait ConnectBehaviour<const THRESHOLD: i16>: Measure {
     async fn good(&self, did: Did) -> bool {
         let conn = self.get_count(did, MeasureCounter::Connect).await;
         let disconn = self.get_count(did, MeasureCounter::Disconnected).await;
+	tracing::debug!(
+	    "[ConnectBehaviour] in Threadhold: {:}, connect: {:}, disconn: {:}, delta: {:}",
+	    THRESHOLD,
+	    conn, disconn, conn - disconn);
         ((conn - disconn) as i16) < THRESHOLD
     }
 }
--- a/core/src/measure.rs
+++ b/core/src/measure.rs
@@ -53,10 +53,13 @@ pub trait ConnectBehaviour<const THRESHOLD: i16>: Measure {
     async fn good(&self, did: Did) -> bool {
         let conn = self.get_count(did, MeasureCounter::Connect).await;
         let disconn = self.get_count(did, MeasureCounter::Disconnected).await;
-	tracing::debug!(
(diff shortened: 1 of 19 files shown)


core/src/measure.rs
```diff
@@ -29,3 +33,61 @@ pub trait Measure {
...
+        let conn = self.get_count(did, MeasureCounter::Connect).await;
+        let disconn = self.get_count(did, MeasureCounter::Disconnected).await;
+        tracing::debug!(
+            "[ConnectBehaviour] in Threadhold: {:}, connect: {:}, disconn: {:}, delta: {:}",
+            THRESHOLD,
+            conn,
+            disconn,
+            conn - disconn
+        );
+        ((conn - disconn) as i16) < THRESHOLD
```
Ma233, line 63: To ensure adaptability across different orders of magnitude, it may be more effective to calculate using the admission base and ratio rather than simply taking the difference.

Ma233 (approved):
Others LGTM

core/src/measure.rs
RyanKung: Agree, this algorithm is just a very basically impl for current situation, we needs ratio based, time weighted based and so on. And futher more, we allow users to compose those impls together.

derive/src/derives.rs
```diff
@@ -0,0 +1,53 @@
...
+    impl<const T: i16> MessageRecvBehaviour<T> for #name {}
+    #[cfg_attr(feature = "node", async_trait)]
+    #[cfg_attr(feature = "browser", async_trait(?Send))]
+    impl<const T: i16> ConnectBehaviour<T> for #name {}
+    };
+
+    #[cfg(not(feature = "core_crate"))]
+    quote! {
+    use rings_core::measure::measure::MessageRecvBehaviour;
+    use rings_core::measure::measure::MessageSendehaviour;
```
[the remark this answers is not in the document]
RyanKung: fixed

PR #480, opened 2023-10-03
chore: remove web3 dependency and fix Did display prefix

Ma233:

* Remove web3 package dependency. Use specific ethereum-type and tiny-keccak packages.
* Fix the display format for Did. Now it has a `0x` prefix when display.

The web3 is a huge package. Rings-core crate do not need all the features of it.

It is replaced by the sub-dependencies ethereum-type and tiny-keccak.

No breaking change.

Closes #issue

### DIFF
--- a/core/src/ecc/mod.rs
+++ b/core/src/ecc/mod.rs
@@ -4,6 +4,7 @@ use std::fmt::Write;
 use std::ops::Deref;
 use std::str::FromStr;
 
+use ethereum_types::H160;
 use hex;
 use rand::SeedableRng;
 use rand_hc::Hc128Rng;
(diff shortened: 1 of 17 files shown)


core/src/ecc/mod.rs
```diff
@@ -196,7 +206,7 @@ impl Serialize for SecretKey {
     }
 }
 
-fn public_key_address(pubkey: &PublicKey) -> Address {
```
RyanKung, line 199: I like the type alias here, If you wanna use ethereum_type::H160, just create a type alias here.

RyanKung (approved):

PR #501, opened 2023-12-05
feature: move session_sk to a separate file in config

Ma233:

This PR moves the session_sk dump from the config file to a separate session_sk file. It then uses the path of that file as a replacement.
Additionally, this PR adds a `new-session` subcommand in the rings bin, allowing users to refresh their session_sk or generate a new one.
However, it is important to note that users still need to restart the rings node after updating the session_sk file.

Cannot generate a config file with template engine since the session_sk need rings bin to generate.
Cannot update session_sk without touch config file when its expired.

The `session_sk` config is now a file path. Simply use `rings new-session` to generate it when needed.

No breaking change.

Closes #issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1985,6 +1985,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "home"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


RyanKung (approved):
lgtm

PR #575, opened 2024-04-25
feat!: differentiate swarms based on network_id

Ma233:

- [ ] Tests for the changes have been added (for bug fixes/features)

This PR introduces a `network_id` field in `ConnectNodeSend` message. It's used to prevent nodes from different networks from connecting to each other.

There is a breaking change in config file.
User must specify network_id in config file now.

Closes #issue

### DIFF
--- a/crates/core/src/message/handlers/connection.rs
+++ b/crates/core/src/message/handlers/connection.rs
@@ -60,6 +60,10 @@ impl HandleMsg<QueryForTopoInfoReport> for MessageHandler {
 #[cfg_attr(not(feature = "wasm"), async_trait)]
 impl HandleMsg<ConnectNodeSend> for MessageHandler {
     async fn handle(&self, ctx: &MessagePayload, msg: &ConnectNodeSend) -> Result<()> {
+        if msg.network_id != self.transport.network_id {
+            return Ok(());
+        }
+
         if self.dht.did != ctx.relay.destination {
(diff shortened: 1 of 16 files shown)


crates/core/src/message/handlers/connection.rs
```diff
@@ -60,6 +60,10 @@ impl HandleMsg<QueryForTopoInfoReport> for MessageHandler {
 #[cfg_attr(not(feature = "wasm"), async_trait)]
 impl HandleMsg<ConnectNodeSend> for MessageHandler {
     async fn handle(&self, ctx: &MessagePayload, msg: &ConnectNodeSend) -> Result<()> {
+        if msg.network_id != self.transport.network_id {
+            return Ok(());
```
Ma233, line 64: Should the current node directly not respond when receiving a handshake request from a different network node?

RyanKung: Yes, response of denial of connection may be expensive in some case, especially the signature/verify phase.

Ma233:
@RyanKung Hi, it's a breaking change in config file. We may should upgrade the middle version number.

RyanKung (approved):

RyanKung:
> @RyanKung Hi, it's a breaking change in config file. We may should upgrade the minor version number.

Yep, version number will be upgrade on PR https://github.com/RingsNetwork/rings/pull/578

PR #593, opened 2026-06-16
refactor(rpc): drop protobuf/prost, keep JSON-RPC

RyanKung:
## What

Remove the protobuf/prost dependency from `crates/rpc`. **The wire protocol is unchanged — still JSON-RPC.**

The RPC messages were generated from `rings_node.proto` via prost, but the wire format has always been JSON-RPC (`serde_json`): protobuf binary `encode`/`decode` was never called anywhere, and the `.proto` `service` blocks were never wired to codegen. So prost only produced a set of serde structs, at the cost of a build script, `prost`/`prost-build-config` deps and proto3 quirks.

## Changes (only `crates/rpc`)

- `protos/rings_node.rs`: prost-generated structs → hand-written `serde` structs with **identical field names/types**, so the emitted/accepted JSON is byte-for-byte the same.
- Removed `build.rs`, `rings_node.proto`, `build_config.yaml`.
- `Cargo.toml`: dropped `prost`, `prost-build-config`, `[build-dependencies]`.
- One incidental rustfmt normalization in `jsonrpc.rs` (the old `build.rs` used to auto-run `cargo +nightly fmt`).

`Method`, the handler dispatch, the JSON-RPC client, cli, FFI and the browser provider are all untouched.

## Verification

- `cargo check` — native (lib + bins + examples), `ffi` feature, browser/wasm (`wasm32-unknown-unknown`)
- `cargo clippy` (rpc + node)


RyanKung (commented):
Blocking issue found in the JS/npm packaging path.

This PR deletes `crates/rpc/src/protos/rings_node.proto`, but the root npm package still depends on that file:

- `package.json` keeps `proto_generate` pointing `pbjs` at `crates/rpc/src/protos/rings_node.proto`, and `prepare` still runs `npm run proto_generate`.
- `index.ts` still exports `./rings_node_proto`, which is the generated output from that deleted proto path.

I reproduced this on PR head `2bb9f8f`:

```text
$ npm run proto_generate
Error: ENOENT: no such file or directory, open 'crates/rpc/src/protos/rings_node.proto'
```

So the Rust `rings-rpc` crate can build without prost, but the npm package prepare/publish path is now broken. Please either keep a proto source for the JS generator, or migrate/remove the npm `proto_generate` + `rings_node_proto` export path as part of this refactor.

Validation I ran:

```text
cargo check -p rings-rpc
cargo check -p rings-rpc --no-default-features --features wasm --target wasm32-unknown-unknown
npm run proto_generate  # fails as above
```

RyanKung:
Fixed in a2fca18. The npm package no longer touches proto:

- Removed the `proto_generate` script and its step in `prepare` (kept `mkdir -p dist` so `prepare` still works).
- Dropped the `export * from "./rings_node_proto"` line in `index.ts`.
- Removed the `protobufjs` / `protobufjs-cli` dependencies.
- Bumped the npm package version to 0.7.1 to match the workspace.

`npm run prepare` now runs wasm_pack → copy → tsc with no proto step.

RyanKung (commented):
Follow-up review on PR head `a2fca18`:

The previous JS/npm packaging blocker is fixed. The new commit removes the proto generation step from `prepare`, removes the `rings_node_proto` export from `index.ts`, drops the protobuf JS deps, and bumps the npm package to `0.7.1`.

Validation I ran:

```text
npm run prepare  # passes
cargo check -p rings-rpc  # passed on the previous head; a2fca18 only changes package.json/index.ts
cargo check -p rings-rpc --no-default-features --features wasm --target wasm32-unknown-unknown  # same
```

Non-blocking packaging hygiene note: `package.json` still publishes the whole `dist/` directory, and `prepare` now does `mkdir -p dist` without cleaning it first. In my current workspace, old ignored `dist/rings_node_proto.{js,d.ts}` files from the previous generator were still present, so `env npm_config_cache=/private/tmp/rings-node-npm-cache npm pack --dry-run --json` included those stale files in the tarball even after the fix. The GitHub release workflow uses a clean checkout before `npm install && npm pack`, so that path should not inherit stale files, but cleaning `dist` in `prepare` would make local/manual packaging deterministic too.

No additional blocking issue found in this pass.

RyanKung:
Addressed the non-blocking hygiene note in ac33e29: `prepare` now runs `rm -rf dist && mkdir -p dist` before copying, so stale generated files (e.g. old `rings_node_proto.*`) can't leak into a local `npm pack`. The release workflow already uses a clean checkout, so this only hardens local/manual packaging.

Issue #594, opened 2026-06-16
RFC: Unify the extension/protocol layer across native and browser


RyanKung:
## Summary

Redesign the `rings-node` extension/protocol layer so that **the abstraction, code
structure, and constraints are identical across `native` and `browser`**. Replace the
closed `BackendMessage` enum and the two divergent dispatch paradigms (native = wasmer
wasm module, browser = JS callbacks) with a single Rust `Extension` trait, a `Ctx`
capability handle, and namespace-based routing. Dynamically-loaded extensions
(wasm / JS) become optional leaf *adapters* that implement the same trait.

This is a **clean break**: the wire format changes, the protocol version is bumped, and
old nodes are **not** supported.

> **Revision (addressing review):** this version tightens the boundary definition —
> envelope codec, unknown-namespace behavior, clean-break rollout, the scope of the
> "ctx is a plain reference" claim, and the full set of public surfaces to migrate
> (RPC request types, CLI/browser APIs, FFI callback ABI), not just `backend/*`.

## Motivation

`rings-core` already cleanly owns the P2P core (DHT / swarm / messaging). The mess is
in the layer above it, `crates/node`, which bundles unrelated concerns: node glue
(`provider`, `processor`), the control plane (`rpc_impl`, `native/endpoint`,
`native/cli`), three runtimes (`native` / `browser` / `ffi`), and a grab-bag of
applications (TCP tunnel, HTTP proxy, SNARK, wasm extensions).

The sharpest pain is the **extension/protocol model**.

### 1. Applications are hard-coded into a closed enum

`crates/node/src/backend/types/mod.rs`:

```rust
#[non_exhaustive]
pub enum BackendMessage {
    Extension(Bytes),
    ServiceMessage(ServiceMessage),   // hard-codes TcpDial/TcpClose/TcpPackage/Http*
    PlainText(String),
    #[cfg(feature = "snark")] SNARKTaskMessage(...),
}
```

Adding/removing an application protocol means editing this core enum, adding a feature
flag, and recompiling the whole node across the native/browser/ffi matrix.

### 2. `native` and `browser` are two different paradigms behind one enum

| | native (`backend/native/extension.rs`) | browser (`backend/browser.rs`) |
|---|---|---|
| An extension is | a **wasm module** loaded by wasmer | a **JS callback** |
| Registration | `ExtensionConfig.paths` | `behaviour.on("Extension", jsFn)` |
[...]

PR #595, opened 2026-06-16
feat(backend): unified extension abstraction (step 1 of RFC #594)

RyanKung:
First step of the extension-layer unification described in #594.

## What

Adds a single, **target-identical** extension abstraction in `crates/node/src/backend/ext.rs`:

- **`Envelope { namespace, payload }`** — namespaced message envelope (bincode on the P2P transport) that will replace the closed `BackendMessage` enum. Inner `payload` codec is the extension's own business.
- **`Ctx`** — capability handle over `Processor` (`did()`, `send()`); extensions never see `Provider`/`Processor` directly.
- **`Extension` trait** + **`Extensions`** namespace router; unknown namespaces are logged at `debug` and dropped (non-fatal).
- **`Provider::ctx()`** to build a `Ctx`.

## Scope

**Additive only.** `BackendMessage` and the existing native/browser dispatch are untouched — nothing is wired to the new types yet. This lands the shared seam so subsequent PRs can port built-ins and remove the enum (RFC rollout steps 3–5).

The only target divergence is the `Send`/`?Send` bound on async trait methods, handled with the crate's usual `cfg_attr` pair.

## Verification

- `cargo check -p rings-node` (native)
- `cargo check -p rings-node --no-default-features --features browser_default --target wasm32-unknown-unknown`

Refs #594

### DIFF
--- a/crates/node/src/backend/ext.rs
+++ b/crates/node/src/backend/ext.rs
@@ -1,0 +1,137 @@
+#![warn(missing_docs)]
+//! Unified extension/protocol abstraction shared by `native` and `browser`.
+//!
+//! This is the seam that replaces the closed `BackendMessage` enum and the two
+//! divergent dispatch paradigms. A protocol is an [`Extension`]; the node routes
+//! inbound [`Envelope`]s to extensions by `namespace`; an extension talks back to the
+//! node only through the [`Ctx`] capability handle.
+//!
+//! The abstraction, structure and constraints are identical on both targets; the only
+//! divergence is the `Send` / `?Send` bound on async trait methods (browser futures
+//! are not `Send`), expressed with the same `cfg_attr` pair used throughout the crate.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use bytes::Bytes;
+use rings_core::dht::Did;
+use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


RyanKung (commented):
I would treat this PR as superseded by #596's pure-core/effect-shell model rather than merge it as the final abstraction.

The additive boundary is useful, but the base API here is still effectful: `Extension::on_message(&self, ctx, from, payload) -> async Result<()>`, and `Ctx::send` performs IO directly. That makes protocol authors write imperative handlers, which is the abstraction we are trying to move away from.

Given the design principle is functional abstraction, the stack should expose a pure transition, roughly `Protocol::step(Ctx<State>, Event) -> Transition<State>`, with `Effect` as data and `Interpreter` as the only IO boundary. If #595 remains in the stack, I would either update it to that pure base or explicitly mark it as an obsolete/mechanical precursor to #596.

RyanKung:
Superseded by #596. The #596 branch already contains this commit; its effectful `Extension::on_message`/`Ctx::send` base was rewritten there into the pure `Protocol`/`Effect`/`Interpreter` model. Collapsing the stack: #596 now targets master and carries the whole change.

PR #596, opened 2026-06-17
feat: route Envelope to extension registry (RFC #594)

RyanKung:
**Draft / WIP** — stacked on #595 (base: `feature/extension-abstraction`). Reviewing live while I implement.

Step 2 of #594: wire namespace routing, then port built-ins on top of it.

## Done in this WIP

- `Backend` holds an `Extensions` registry + `register_extension()`.
- Transitional `BackendMessage::Envelope(Envelope)` variant (built-ins migrate to it, then it becomes the only wire shape and `BackendMessage` is deleted).
- `Backend::on_backend_message` routes `Envelope` → `Extensions::dispatch` (sender via `MessageVerificationExt::signer`); other variants keep the legacy path; unknown namespaces dropped.

## Not yet (incoming commits)

- Port built-ins to `Extension` impls — **PlainText first** as the reference, then HTTP/TCP proxy and SNARK.
- Migrate public surfaces (`Processor::send_backend_message`, RPC request type, CLI, browser provider, FFI callback ABI).
- Remove `BackendMessage`/`ServiceMessage` and the dual dispatch.

Currently the registry is empty, so `Envelope` is inert on the wire. Native + wasm compile.

Refs #594

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #598, opened 2026-06-19
fix(ci): scope release build to the rings binary so musl links

r2d2-rs:
## Problem

The **Auto Release** workflow's `build linux version binary (x86_64, musl)` job fails at link time on `master` after #596:

```
error: linking with `cc` failed: exit status: 1
  /usr/bin/ld: /usr/lib/gcc/.../libstdc++.a(...): ...
  collect2: error: ld returned 1 exit status
```

The static-musl link of `rings` pulls SNARK's C++ deps (`pasta-msm` / `semolina`), and GNU `libstdc++.a` doesn't resolve into a `-static-pie` musl binary. macOS/wasm builds are fine (libc++ linked dynamically).

## Root cause

Feature unification, not link flags. The release step ran `cargo build --features node --no-default-features` at the **workspace root**, so it builds *every* member. #596 added the examples (including `examples/snark`) as workspace members; `examples/snark` enables `rings-node`'s `snark` feature, and with resolver-2 unification `rings-node` is then compiled **with** snark — so the `rings` binary links SNARK's C++ even though `--features node` alone doesn't enable it.

## Fix

Scope the release build to the package + binary:

```
cargo build -p rings-node --bin rings --features node --no-default-features --release --target <triple>
```

With `-p rings-node` the examples aren't built, nothing forces `rings-node/snark`, and `rings-snark` drops out of the dependency tree — so the musl binary has no C++ to statically link. Applied to both the linux/musl and macOS build steps.

## Verification

- `cargo tree -p rings-node --no-default-features --features node -i rings-snark` → `did not match any packages` (snark not in the tree).
- `cargo build -p rings-node --bin rings --no-default-features --features node` builds clean.
- QACI was already green (it builds on GNU linux, which links libstdc++ dynamically); only the static-musl release was affected.


r2d2-rs:
### Follow-up: PR CI now runs the same build steps as release (`532b240`)

The original break slipped through because **nothing on the PR path built the release artifacts** — only `Auto Release` (master/tag) did. Fixed the class, not just the instance:

- Extracted the macOS / static-musl-linux / wasm build jobs into a **reusable workflow** (`.github/workflows/release-build.yml`, `workflow_call` with a `version` input).
- `Auto Release` now calls it (keeps `gen_version` + the publishing `release` job; artifact flow unchanged).
- **QACI calls the same reusable workflow on every PR** (`version: ci`, throwaway artifacts, no publish).

Result: a release-only build failure (e.g. the static-musl C++ link) now fails **PR** CI, and the two pipelines share one build definition so they can't drift.

You can see it on this PR right now — the `Release build (parity) / build linux version binary (x86_64, musl)` check is running here, which is exactly the job that was red on master. Once it's green, the fix is proven on the PR itself.

r2d2-rs:
### Correction: PR path now only *compiles* the release targets (`3ab91f7`)

Good catch — you're right that a PR must not version or publish. My reusable-workflow commit pulled release machinery (version naming + artifact packaging) onto the PR, which was wrong even though it never published.

Split build from packaging in the reusable workflow with a `publish_artifacts` input (default `false`):

- The `cargo build` of the release targets **always** runs — that's the part a PR must verify.
- zip / `npm pack` / version naming / artifact upload run **only** when `publish_artifacts: true`.
- The wasm `npm pack` job is release-only (wasm compilation is already covered on PRs by QACI's `build_wasm`).

So now:
- **PR (QACI)** → calls with defaults → compiles macOS + static-musl `rings` binary only. **No version, no package, no upload, no publish.**
- **Auto Release** → calls with `publish_artifacts: true`; `gen_version` and the GitHub Release stay entirely in `auto-release.yml`.

The static-musl compile (the thing that broke) still runs on every PR; the rest of the release machinery doesn't touch the PR path.

r2d2-rs:
### Narrowed the PR secret surface + no double-build on master (`3e4322f`)

Both points addressed.

**`secrets: inherit` → `github.token`.** Dropped `secrets: inherit` from both callers. The reusable workflow now uses the automatic per-run `github.token` for `setup-protoc`, so the PR path — which checks out and builds untrusted PR code — no longer sees every caller-available repository/org secret. (The remaining `secrets.GITHUB_TOKEN` references are each workflow's *own* jobs: Auto Release's publish step and QACI's own protoc setup, which have direct access and don't cross the reusable-workflow boundary.)

**Cost note.** Gated the QACI `build_release_targets` parity job to `if: github.event_name == 'pull_request'`. On master pushes Auto Release already builds these targets, so this drops the duplicate release-artifact build there while keeping full PR coverage.

On the earlier verification caveat (a green QACI didn't prove the Auto Release build step): that's now resolved by the shared workflow — the static-musl build runs on this very PR (it was green in the parity run), so the fix is proven pre-merge without a release dispatch.

r2d2-rs:
### Browser job now builds on PRs too (`57451c6`)

Good catch — you're right that `npm pack` is not just packaging: the root `package.json` `prepare` runs `wasm-pack build … --features browser_default`, assembles `dist`, and runs `tsc`, which QACI's cargo-wasm jobs don't exercise. So a release-only browser break could still slip through.

Mirrored the native split on the browser job:
- Removed the job-level `if: publish_artifacts` gate, so **`npm install && npm pack` runs on every PR** (the real build path — wasm-pack + dist + tsc).
- Gated only the versioned rename (`mv … rings-<version>-…tgz`) and the artifact upload behind `publish_artifacts`.

Now all three release targets — macOS, static-musl linux, and browser — build on every PR; only packaging/upload stays release-only. The PR check shows `Build release targets / build browser version binary` running (no longer skipped).

r2d2-rs:
### Fixed the stale QACI comment (`<pending>`)

You're right — after the browser split the comment no longer matched the boundary. Updated it to say the PR builds **all** release targets (macOS, static-musl linux, **and** browser — the browser job runs `npm pack`, i.e. `wasm-pack`/`dist`/`tsc`), and that only the versioned rename + artifact upload are release-only. Also dropped the "macOS + static-musl only" target list.

That was the last open item; all release-target checks (musl, both macOS, browser) are green on this PR.

r2d2-rs:
### Fixed the reusable-workflow header too (`a-pending`)

You're right — I fixed the QACI comment but left the same stale boundary in `release-build.yml`'s header. Now corrected: every release target is *built* on both paths (`cargo build` for native, `npm pack` for browser — which runs `prepare`: `wasm-pack`/`dist`/`tsc`); only versioned artifact naming + upload are gated by `publish_artifacts`. Also updated the `publish_artifacts` input description, and swept all three workflow files so no other comment still implies "PRs never run npm pack / never package".

Issue #602, opened 2026-06-23
Harden ElGamal encryption against nonce reuse and malformed ciphertexts


RyanKung:
## Summary

The current `rings-core` ElGamal implementation in `crates/core/src/ecc/elgamal.rs` has several security and reliability problems that are not covered by the existing tests.

Inspected locally on `master` at commit `4bda65d4b595f8eb3cf21da1490471e533e54a69`.

## Findings

1. **One ephemeral scalar is reused for every plaintext block in a single encryption call**

   `encrypt` creates `random_sar` once before iterating over all plaintext chunks:

   - `crates/core/src/ecc/elgamal.rs:119-124`
   - `crates/core/src/ecc/elgamal.rs:130-139`

   Because the same scalar is reused for every block, all blocks in one message share the same `c1` and the same shared secret. This leaks block relationships (`c2_i - c2_j = m_i - m_j` in group terms) and makes repeated/plaintext structure observable inside a multi-block ciphertext. EC-ElGamal should use a fresh ephemeral scalar for each independently encrypted plaintext element, or the implementation should switch to a KEM/DEM construction that derives an AEAD key once and encrypts the byte payload as authenticated symmetric ciphertext.

2. **Malformed ciphertext can panic during decrypt**

   `decrypt` converts incoming `CurveEle<33>` values with `expect("bad curve point")`:

   - `crates/core/src/ecc/elgamal.rs:152-160`

   `CurveEle<33>` is an alias for `PublicKey<33>`, and its serde/base58 path validates encoding and length but not that the bytes are a valid secp256k1 point before `decrypt` receives them (`crates/core/src/ecc/types.rs:24-74`, `crates/core/src/ecc/types.rs:88-104`). A malformed encrypted message can therefore turn parse failure into a panic instead of a recoverable `Result::Err` such as `Error::DecryptionError` or `Error::UnexpectedEncryptedData`.

3. **Plaintext mapping is not lossless for valid Rust strings containing NUL bytes**

   `field_to_str` zeroes the first byte and then drops all leading zero bytes in each field chunk:

   - `crates/core/src/ecc/elgamal.rs:53-72`

   Since `str_to_field` chunks UTF-8 bytes in 31-byte pieces, valid strings with `\0` at the start of the message or at a chunk boundary will not round-trip. Existing tests only generate `Alphanumeric`, so they do not catch this.

4. **Debug `println!` calls remain in library crypto paths**

   `field_to_str` and `decrypt` print intermediate field/point data:

[...]

PR #603, opened 2026-06-24
refactor ECC ElGamal over curve groups

RyanKung:
## Summary

Refs #602.

This draft refactors EC ElGamal so the algorithm itself lives over an abstract cyclic group instead of the secp256k1 implementation. The secp256k1 message encoding and legacy public-key adapter now live in a separate `secp256k1_elgamal` module.

It also introduces shared ECC key abstractions for signing, verification, and encryption keys, adds curve-group adapters for secp256k1, secp256r1, BLS12-381 G1, and Ristretto255, and wires BLS/Ed25519 account verification through the session path.

## Issue 602 coverage

- Uses a fresh ElGamal ephemeral scalar for each encrypted group element.
- Returns errors instead of panicking on malformed secp256k1 ciphertext input.
- Preserves NUL bytes in the secp256k1 string-to-point adapter encoding.
- Removes debug output from the crypto path.
- Adds an ignored 4 KiB encrypt/decrypt performance probe.

## Validation

- `cargo test -p rings-core secp256r1 --lib`
- `cargo test -p rings-core secp256k1_elgamal::test::bench_encrypt_decrypt_4kb --release -- --ignored --nocapture`
- `git diff --check`

## Known follow-ups before ready review

- BLS signing secret material still uses the existing `SecretKey` wrapper; it should be moved to a dedicated BLS scalar/secret-key type.
- Random scalar sampling is currently part of `CyclicGroup`; a cleaner algebra boundary would move sampling into the ElGamal algorithm layer or a scalar sampler.
- BLS account parsing validates byte length but should also validate compressed G1 decoding at parse time.


RyanKung:
## Review

Overall a solid, well-tested refactor. The `secp256r1::verify` rewrite incidentally fixes a real signature-verification bypass: the old code fell through to `true` when `ct_pk.is_some()` was false (an undecodable P-256 key), so an invalid public key would accept any signature. The new `if !bool::from(ct_pk.is_some()) { return false }` closes it, with a regression test 👍. ElGamal round-trips, NUL-byte preservation, and the no-panic-on-malformed-input path are all correct and covered. No blocking correctness issues found.

A few actionable suggestions:

### 1. The secp256r1 / BLS12-381 G1 / Ristretto255 adapters in `group.rs` are dead on arrival
Production only ever instantiates `Group<Secp256k1>` (used by the ElGamal adapter). The `CurveGroup` impls for the other three curves, plus their `Point`/`Scalar` `From` conversions, are exercised solely by the in-module `group_laws` test — ~250 lines plus the `curve25519-dalek` direct dependency, with no caller. Suggest landing these adapters in the PR that actually uses them, or keeping only secp256k1 here.

### 2. Duplicated domain-separation bytes + unused `transcript_bytes` in `keys.rs`
`PublicKeyMaterial::transcript_bytes` builds `algorithm ++ 0x00 ++ raw_bytes` but is never called anywhere; the same construction is then hand-rolled again in `VerificationPublicKey::domain_separated_address` and `EncryptionPublicKey::did`. Both `did`/address sites could collapse onto `keccak256(&self.transcript_bytes())[12..]` (both types already `impl PublicKeyMaterial`), keeping the domain-separation rule in exactly one place.

### 3. `Session::account_pubkey()` is a silent public-API semantic change
It now returns `Err(UnknownAccount)` for `Secp256r1`/`Ed25519` (and the new `Bls12381`) accounts, where it previously returned `Ok(pk)`. The change is intentional and tested, and there's no in-repo production caller — but this is a `pub` API on `rings-core`, so it's worth a line in the changelog/migration notes rather than only the reworded doc comment.

### Already listed as follow-ups in the PR description (no action needed this round)
- BLS `from_account_parts` only length-validates the 48-byte key; a non-canonical/off-curve encoding is accepted as an `Account` and only fails later at `bls::verify` (returns `false`, no panic).
- `random_scalar` living on `CyclicGroup` and BLS secret material reusing `SecretKey` are both already on the follow-up list.

crates/core/src/ecc/elgamal.rs
```diff
@@ -1,327 +1,169 @@
...
-//! 2. Compute 𝑠−1, the inverse of 𝑠 in the group 𝐺
-//! 3. Compute 𝑚:=𝑐2⋅𝑠−1
+//! With the group written additively, encryption of a group element `m` under
+//! public key `h = xG` is:
 //!
-//! ref:
-//!    T. ElGamal. A Public Key Cryptosystem and a Signature Scheme Based on Discrete Logarithms. IEEE Trans. Info. Theory, IT 31:469–472, 1985.
-//!    ElGamal encryption <https://en.wikipedia.org/wiki/ElGamal_encryption>
-//!    <http://www.docsdrive.com/pdfs/ansinet/itj/2005/299-306.pdf>
-use libsecp256k1::curve::Affine;
```
RyanKung, line 25: you should not delete those document

crates/core/src/ecc/group.rs
```diff
@@ -0,0 +1,562 @@
+//! Elliptic-curve group abstractions and curve adapters.
+
```
RyanKung, line 26: need more details about the impl, you shoud doc, what is group, how we abstract doc.

crates/core/src/ecc/group.rs
```diff
@@ -0,0 +1,562 @@
...
+    fn mul(point: &Self::Point, scalar: &Self::Scalar) -> Self::Point {
+        *point * *scalar
+    }
+
+    fn eq(lhs: &Self::Point, rhs: &Self::Point) -> bool {
+        lhs == rhs
+    }
+}
+
```
RyanKung: for repeat impl, use derive macro to make it simplier

RyanKung: for repeat impl, use derive macro to make it simplier

crates/core/src/ecc/keys.rs
```diff
@@ -0,0 +1,726 @@
...
+
+/// A value that has a stable Rings DID.
+pub trait DidSubject {
+    /// Derive or return the Rings DID for this subject.
+    fn did(&self) -> Did;
+}
+
+/// Common public-key material behavior.
+pub trait PublicKeyMaterial: DidSubject {
+    /// Stable lower-case algorithm name.
```
RyanKung: I dont like this name, `Material`
what this Trait use for?

crates/core/src/ecc/keys.rs
```diff
@@ -0,0 +1,726 @@
...
+use crate::dht::Did;
+use crate::error::Error;
+use crate::error::Result;
+
+/// A value that has a stable Rings DID.
+pub trait DidSubject {
+    /// Derive or return the Rings DID for this subject.
+    fn did(&self) -> Did;
+}
+
```
RyanKung, line 48: it's actuall a did wrapper, and do we really need a wrapper?

crates/core/src/ecc/keys.rs
```diff
@@ -0,0 +1,726 @@
...
+            Self::Bip137(sk) => VerificationPublicKey::Bip137(sk.pubkey()),
+            Self::Secp256r1(sk) => VerificationPublicKey::Secp256r1(secp256r1_public_key(*sk)),
+            Self::Ed25519(sk) => VerificationPublicKey::Ed25519(sk.public_key()?),
+            Self::Bls12381(sk) => VerificationPublicKey::Bls12381(signers::bls::public_key(sk)?),
+        })
+    }
+}
+
+impl EncryptionPublicKey {
+    /// Encryption algorithm for this public key.
```
RyanKung: 'EncryptionPublicKey?' why it's releated to secp256k1?

crates/core/src/ecc/elgamal/impls/secp256k1.rs
RyanKung, line 1: this file should under /elgamal/impls path
and you should implement elgamal for all curves, nod only secp256k1

RyanKung:
## Re-review (v2)

The previous round is well addressed: `random_scalar` moved off the group axioms into `CryptographicGroup`/`CurveScalarSampler`, the three isomorphic curve adapters are folded into `impl_curve_group_adapter!`, the DID transcript construction is centralized in `public_key_transcript`/`domain_separated_address`, the `account_pubkey` change is documented in the README migration note, and the adapters are now exercised end-to-end via ElGamal. Comments below are organized along the four axes raised.

### 1. Functional abstraction — separate effect from pure (the biggest item)

Randomness is a hidden global effect. `random_scalar` reads the `GROUP_RNG` thread-local (`Hc128Rng::from_entropy`) with no injection point, and `ElGamal::encrypt` calls `G::random_scalar()` *inside* the `.map` closure. So `encrypt`'s type signature `(message, &pubkey) -> ciphertext` looks pure but is non-deterministic and impossible to seed — you can't property-test it against a fixed transcript or reproduce a ciphertext.

This is exactly the functional-core / imperative-shell split worth making explicit:

- Make sampling an explicit effect parameter — thread `rng: &mut impl RngCore` through `random_scalar` / `ElGamal::encrypt` / `ElGamalKeyPair::random`, keeping the `thread_local` only as the default at the outermost shell.
- Factor the per-block map into a *pure* kernel `encrypt_block(m, h, r) -> (c1, c2) = (rg, m + rh)` and an effectful layer that supplies the `r`s. Right now the pure transform `(m, r) ↦ (rg, m+rh)` and the effect `r ← sample` are entangled in one closure.

`Ed25519SecretKey::random` has the same hidden-effect shape and additionally builds a fresh `Hc128Rng::from_entropy()` per call.

### 2. Group theory / category theory — the isomorphic adapters

I checked all four adapters (secp256k1 hand-written, secp256r1/BLS12-381 G1/Ristretto255 via the macro). The wrapper `Group<C>`/`Point<C>`/`Scalar<C>` is a functor from the `CurveGroup` category to `GroupOps`, the `From`/`Into` pairs are the witnessing isomorphisms, and the per-curve op tables are correct; `group_laws` covers identity/inverse/associativity for all four. Two gaps:

- **The one non-trivial isomorphism is only tested transitively.** For secp256k1, `generator_mul` goes through `ecmult_gen` while `*` (`mul`) goes through `ecmult_const` — two distinct code paths that must agree. Their equality `generator_mul(s) == generator() * s` is only exercised indirectly through the ElGamal round-trip. Add a direct law test asserting it (for the macro curves it's definitionally `generator()*s`, so secp256k1 is the meaningful case).
- **`CurveGroup` documents operations but not their laws.** This trait is the adapter boundary an implementer must get right, yet `add`/`neg`/`mul`/`eq` are one-line "what" docs. The subtle one is `eq`: it must be *equality of group elements* (respecting projective representatives), which is why secp256k1 must normalize to affine bytes and why a naive `Jacobian` field-compare would silently be wrong. State the laws on the trait (associative+commutative `add`, neutral `identity`, `neg` yields inverses, `mul` is the scalar action, `eq` an equivalence consistent with the group). Note also that `Scalar: Clone` is opaque — the abstraction can't express scalar-action laws; that's an acceptable limit but worth a doc line.

### 3. Inefficient clones

The generic algorithm is written against the by-value `std::ops` on `GroupOps::Element`, so it clones on every operation: `ElGamal::encrypt` does `public_key.as_element().clone()` per block, and `decrypt` does `c2.clone() + -(c1.clone() * secret_key.as_scalar().clone())` — three clones per element, including re-cloning the secret scalar each iteration. It's free today only by accident: every current point/scalar type happens to be `Copy`. The inner `CurveGroup` is already borrow-based (`add(&,&)`, `mul(&,&)`), so the asymmetry is purely in the wrapper. Either document that `GroupOps::Element` is expected to be cheap-to-clone (`Copy`-ish), or expose reference-based ops so the algorithm borrows instead of cloning — otherwise a future non-`Copy` point type silently turns each block into several deep copies.

### 4. Documentation robustness

Module-level docs are good. The thin spot is the most error-prone code: the secp256k1 message encoding (`str_to_field` / `decode_field_bytes` / `lift_x` in `elgamal/impls/secp256k1.rs`). The `[0xFF, 0x52 marker, len, ...data]` field layout, the invariant that `lift_x` clobbers byte[0] (so decode must ignore it) which is what makes the marker/length scheme reversible, and the NUL-preservation rationale are all undocumented. `lift_x` also has an undocumented `panic!` path (`Some(0)`) — the "no valid x found in 254 tries" failure mode of `encrypt` should be stated as an invariant. Please document the format and the byte[0] interaction inline; it's the part most likely to break under a future refactor.

Nothing here is a correctness blocker — the crypto round-trips and the prior secp256r1 verify-bypass fix still hold. These are about pushing the effect/algebra boundary and the docs to the bar the abstraction is aiming for.

RyanKung:
## Re-review (v3)

All four points from the previous round are addressed, and the changes are backed by tests rather than just docs. Verified:

**1. Effect / pure separation.** `ElGamal::encrypt_block(m, h, r) = (rg, m + rh)` is now a pure kernel with no sampling; `encrypt_with_rng(.., rng: &mut impl RngCore)` threads randomness explicitly; `random_scalar_with_rng` is the primitive and `random_scalar()` is a documented thread-local convenience shell. The secp256k1 adapter and `ElGamalSecretKey`/`ElGamalKeyPair`/`Ed25519SecretKey` all gained `*_with_rng` variants. Reproducibility is now a test (`encrypt_with_rng_is_reproducible_for_same_seed`, `encrypt_block_is_pure_group_operation`, `ed25519_random_with_rng_is_reproducible_for_same_seed`) — exactly the determinism that was missing.

**2. Group/category laws + the gen/const isomorphism.** `CurveGroup` and `GroupOps` now state the laws (identity, associativity+commutativity, inverse, scalar action, and `eq` as an equivalence over group elements rather than raw representation, compatible with the ops). The non-trivial secp256k1 isomorphism `generator_mul (ecmult_gen) == mul (ecmult_const)` now has a direct assertion in `group_laws` (`generator_mul(a) == mul_ref(generator(), a) == generator()*a`), instead of only being exercised transitively. The opaque-`Scalar` limitation is documented.

**3. Clones.** `GroupOps` grew `add_ref`/`neg_ref`/`mul_ref`, overridden by `Group<C>` to call the native borrowed curve ops. `decrypt` and `encrypt_block` are now fully borrow-based — the per-block public-key clone and per-element `c1`/`c2`/scalar clones are gone, so a future non-`Copy` point type no longer silently deep-copies.

**4. Docs.** The secp256k1 encoding is now documented end to end: the `[0xFF, 0x52 marker, len, data, padding]` field layout, the `lift_x` byte[0]-clobber invariant that makes the mapping reversible, the NUL-preservation rationale, and the `panic!` as a stated invariant failure.

Nothing blocking. Two optional cleanups:

- `CurveScalarSampler::random_scalar()` (the marker-level default) appears unused — `Group<C>` samples via `CryptographicGroup::random_scalar`, and nothing calls `C::random_scalar()` directly. Could drop it to keep one sampling entry point per tower.
- There are now two independent default samplers (`GROUP_RNG` in `group`, `KEY_RNG` in `keys`), both `Hc128Rng::from_entropy`. Fine as-is given the module split, just noting the duplicated "default thread-local sampler" concept.

This now hits the effect/algebra/clone/doc bar the abstraction was aiming for.

PR #604, opened 2026-06-24
Default correct chord and bump 0.10.0

### LINKED ISSUE #388: Correctness of DHT
### FACT

1. Chord must be initialized with a ring containing a minimum of `r +1` nodes, where `r` is the length of each node’s list of successors. In fact, to be proven correct, a Chord network must maintain a “stable base” of `r + 1` nodes that remain members of the network throughout its lifetime.

2. The Chord Paper defined the maintenance and use of `ﬁnger tables`, which improve lookup speed by providing pointers that cross the ring like chords of a circle. **Because ﬁnger tables are an optimization and they are built from successors and predecessors, correctness does not depend on them.**

### IMPROVEMENTS

1. Join Operator: Sync successor list from new successor. Add new remote action:

```
RemoteAction::FetchSuccessorList(Did)
```

After join operator, call `FindSuccessor(did)` and `FetchSuccessorList(Did)` immediately。

2. Stab Operator: For current implementation, we notify predecessor for successor updating. But on paper HMCC, It defined a new operation that fetch successor's predecessor and successor_list periodically. And instead of notify predecessor, It notify successors.

The loop is guaranteed to terminate before succList is empty, based on the assumption that successor lists are long enough so that each list contains at least one live node.

Screenshot 2023-03-31 at 10 13 33 PM

3. Rectify operator, In origin Chord paper the operator is so called `check_predecessor`.

Screenshot 2023-03-31 at 10 18 33 PM

```

```

ref: How to Make Chord Correct  https://arxiv.org/pdf/1502.06461.pdf

RyanKung:
## Summary

Fixes #388.

- make the HMCC/Zave CorrectChord join and stabilize paths the default instead of gating them behind `experimental`
- keep join-time convergence messages best-effort so transient send failures do not suppress the application-level `Connected` event
- fix `RemoteAction::Notify` delivery so the predecessor payload is sent to the remote target
- align `PeerRing::stabilize` with the `CorrectStabilize` operation, including empty successor info, improved-predecessor query behavior, and successor notify behavior
- ensure stabilization-learned successor/predecessor candidates become live transport connections before direct topology/notify maintenance depends on them
- expose configurable DHT finger-table size while keeping the production default at 160 slots and clamping out-of-range sizes for 160-bit `Did`
- keep the full async FIFO/LIFO DHT schedule test, but run it with a 16-slot configured finger table and actionable convergence diagnostics
- remove random dummy message delay while controlled delivery is enabled, so schedule tests measure deterministic protocol progress instead of wall-clock jitter
- add TLA+-style `CorrectStabilize` comments plus Rust spec mirror tests for the default path
- add stronger successor-list validation for order independence, duplicate/self filtering, truncation to K nearest successors, and ring wrap-around ordering
- add a dummy-transport regression test proving `join_dht` keeps the local join when follow-up convergence sends fail
- add a direct handler regression test for `RemoteAction(target, Notify(pred))` delivery direction and payload semantics
- note that `CorrectStabilize` now emits `NotifyPredecessor` for the current successor each stabilize round when the head is not self, matching the spec while increasing steady-state notify traffic versus the old default path
- bump the Rust workspace and npm package version to `0.10.0`

## Checks

- `cargo +nightly fmt --all`
- `cargo +nightly fmt --all --check`
- `cargo check -p rings-core --features dummy`
- `cargo check -p rings-transport --features dummy`
- `cargo clippy -p rings-core --features dummy --tests -- -D warnings`
- `cargo test -p rings-core --features dummy successor`
- `cargo test -p rings-core --features dummy dht_convergence`
- `cargo test -p rings-core --features dummy dht_convergence -- --test-threads=1`
- `cargo test -p rings-core --features dummy correct_chord`
- `cargo test -p rings-core --features dummy test_finger_table_size_bounds -- --test-threads=1`
- `cargo test -p rings-core --features dummy test_handle_dht_notify_remote_action_sends_predecessor_to_target -- --test-threads=1`
- `cargo test -p rings-core --features dummy test_join_dht_keeps_local_join_when_convergence_send_fails -- --test-threads=1`
- `cargo test -p rings-core --features dummy test_stabilization -- --test-threads=1`
- `cargo test -p rings-core --features dummy test_handle_join -- --test-threads=1`
- `cargo test -p rings-core --features dummy dht_stateright -- --test-threads=1`
- `cargo test -p rings-core --features dummy dht_trace_replay`
- `cargo test -p rings-core --features dummy dht_schedule -- --test-threads=1`
- `cargo test -p rings-core --features dummy test_message_handler -- --test-threads=1`
- `cargo test -p rings-core --features dummy message::handlers::connection::tests -- --test-threads=1`
- `cargo test -p rings-core --features dummy message::handlers::stabilization::test -- --test-threads=1`
- `cargo test -p rings-core --features dummy tests::default::test_connection -- --test-threads=1`
- `cargo test -p rings-core --features dummy -- --test-threads=1`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4254,7 +4254,7 @@ dependencies = [
 
 [[package]]
 name = "rings-browser-example"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "bytes",
  "js-sys",
@@ -4266,7 +4266,7 @@ dependencies = [
 
 [[package]]
 name = "rings-core"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "ark-bls12-381",
  "ark-ec",
@@ -4336,7 +4336,7 @@ dependencies = [
 
 [[package]]
 name = "rings-derive"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -4346,7 +4346,7 @@ dependencies = [
 
 [[package]]
 name = "rings-native-example"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "async-trait",
  "base64 0.13.1",
@@ -4360,7 +4360,7 @@ dependencies = [
 
 [[package]]
 name = "rings-node"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "anyhow",
  "arrayref",
@@ -4421,7 +4421,7 @@ dependencies = [
 
 [[package]]
 name = "rings-relay-example"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "rings-core",
  "rings-node",
@@ -4430,7 +4430,7 @@ dependencies = [
 
 [[package]]
 name = "rings-rpc"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "async-trait",
  "base64 0.13.1",
@@ -4447,7 +4447,7 @@ dependencies = [
 
 [[package]]
 name = "rings-snark"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "bellpepper-core",
  "byteorder",
@@ -4476,7 +4476,7 @@ dependencies = [
 
 [[package]]
 name = "rings-snark-example"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "async-trait",
  "rings-core",
@@ -4488,7 +4488,7 @@ dependencies = [
 
 [[package]]
 name = "rings-transport"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "async-trait",
  "bincode",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ resolver = "2"
 members = ["crates/*", "examples/browser", "examples/native", "examples/relay", "examples/snark"]
 
 [workspace.package]
-version = "0.9.0"
+version = "0.10.0"
 edition = "2021"
 license = "GPL-3.0"
 authors = ["RND <dev@ringsnetwork.io>"]
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -11,9 +11,8 @@ authors.workspace = true
 repository.workspace = true
 
 [features]
-# Feature "experimental" is used to mark an implementation as experimental, which means that:
-# It may not have been thoroughly tested.
-# The corresponding API may be deprecated or adjusted in the future.
+# Deprecated compatibility alias. The formerly experimental correct-chord path
+# now runs by default.
 experimental = ["std"]
 default = ["std"]
 redundant = ["default"]
--- a/crates/core/src/dht/chord.rs
+++ b/crates/core/src/dht/chord.rs
@@ -95,7 +95,10 @@ pub enum RemoteAction {
     FindVNode(Did),
     /// Need `did_a` to find VirtualNode for operating.
     FindVNodeForOperate(VNodeOperation),
-    /// Let `did_a` [notify](Chord::notify) `did_b`.
+    /// Send a predecessor notification to `did_a`.
+    ///
+    /// `did_a` is the remote recipient from [`PeerRingAction::RemoteAction`].
+    /// This field is the predecessor DID announced in `NotifyPredecessorSend`.
     Notify(Did),
     /// Let `did_a` sync data with it's successor.
     SyncVNodeWithSuccessor(Vec<VirtualNode>),
@@ -203,6 +206,9 @@ impl PeerRing {
     }
 
     /// Same as new with config, but with a given storage and finger table size.
+    ///
+    /// `Did` is 160-bit. Sizes above [`DEFAULT_FINGER_TABLE_SIZE`] are clamped
+    /// by [`FingerTable::new`]; zero is allowed to disable finger maintenance.
     pub fn new_with_storage_and_finger_table_size(
         did: Did,
         succ_max: u8,
@@ -616,7 +622,12 @@ impl CorrectChord<PeerRingAction> for PeerRing {
     }
 
     /// Stabilize Operation:
-    /// Perform stabilization for the successor list.
+    ///
+    /// Mirrors the TLA+-style `CorrectStabilize` operator in
+    /// `tests/default/dht_convergence.rs`.
+    /// The old head is captured before updating successors for the improved-successor
+    /// query check; the remote successor list contributes `but_last`; and notify
+    /// is emitted for the post-update head when that head is not self.
     fn stabilize(&self, info: TopoInfo) -> Result<PeerRingAction> {
         let mut ret = vec![];
         let successors = self.successors();
--- a/crates/core/src/dht/chord.rs
+++ b/crates/core/src/dht/chord.rs
@@ -11,6 +11,7 @@ use serde::Deserialize;
 use serde::Serialize;
 
 use super::did::BiasId;
+use super::finger::DEFAULT_FINGER_TABLE_SIZE;
 use super::successor::SuccessorSeq;
 use super::types::Chord;
 use super::types::ChordStorage;
@@ -193,11 +194,25 @@ impl From<Vec<PeerRingAction>> for PeerRingAction {
 impl PeerRing {
     /// Same as new with config, but with a given storage.
     pub fn new_with_storage(did: Did, succ_max: u8, storage: VNodeStorage) -> Self {
+        Self::new_with_storage_and_finger_table_size(
+            did,
+            succ_max,
+            storage,
+            DEFAULT_FINGER_TABLE_SIZE,
+        )
+    }
+
+    /// Same as new with config, but with a given storage and finger table size.
+    pub fn new_with_storage_and_finger_table_size(
+        did: Did,
+        succ_max: u8,
+        storage: VNodeStorage,
+        finger_table_size: usize,
+    ) -> Self {
         Self {
             successor_seq: SuccessorSeq::new(did, succ_max),
             predecessor: Arc::new(Mutex::new(None)),
-            // for Eth address, it's 160
-            finger: Arc::new(Mutex::new(FingerTable::new(did, 160))),
+            finger: Arc::new(Mutex::new(FingerTable::new(did, finger_table_size))),
             storage,
             cache: Box::new(MemStorage::new()),
             did,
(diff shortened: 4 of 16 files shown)


RyanKung:
## Review

The `stabilize`/`Notify` correctness work is solid and well-mirrored by the new spec tests. The `RemoteAction::Notify` fix is a genuine bug fix and self-contained: `chord.rs:628` is the only producer and `handlers/mod.rs:138` the only consumer, so re-defining `RemoteAction(target, Notify(pred))` as "send `pred` to `target`" is safe — the old code sent `NotifyPredecessorSend{ self.did }` to the payload DID (i.e. to itself) instead of to the successor. The `succ_len - 1` panic on empty `info.successors` is fixed via `saturating_sub` (with a regression test), and the improved-predecessor check now correctly compares the topo predecessor against the *pre-update* old head, matching `CorrectStabilize.Improved`. `experimental` is cleanly reduced to a no-op alias with no remaining `cfg` gates.

One thing worth a closer look, plus two minor notes.

### 1. (medium) `join_dht` drops the old default path's best-effort error handling, which now gates `Connected`

The previous non-experimental branch intentionally swallowed `handle_dht_events` failures:

```rust
if let Err(e) = self.handle_dht_events(&dht_ev).await {
    tracing::warn!("Failed to handle dht events while joining {peer}: {e:?}");
}
Ok(())
```

with the comment *"A send here can legitimately fail if the peer churned away, so log and continue instead of panicking."* The new unified path propagates instead (`...join_then_sync(conn).await?; self.handle_dht_events(&dht_ev).await`).

This matters at the hot caller `swarm/callback.rs:204`: `on_data_channel_open` does `self.message_handler.join_dht(did).await?;` **before** emitting `ConnectionStateChange{ Connected }` (lines 209–214). So a transient convergence-message send failure during join — exactly the churn case the deleted comment cited — now aborts `on_data_channel_open` and the peer is never surfaced as `Connected` to the application callback. Since `experimental` was *not* in `default` before, this is a behavior change for every default build, not just a relabel.

Suggest keeping the join-path `handle_dht_events` best-effort (log & continue) while adopting `join_then_sync`, so connection establishment isn't gated on best-effort convergence sends succeeding.

### 2. (low) Spec mirror `correct_stabilize_query` uses `current.first()`, production uses `successors.min()`

Production reads the old head via `SuccessorSeq::min()` (index 0 of the distance-sorted list). The spec helper takes `current.first()` of the raw input slice, so the two agree only when `current` is already sorted by `dist(me, ·)`. All four new tests pass a 0- or 1-element `current`, so it's never exercised, but a future multi-element/unsorted case would diverge from production for a spurious reason. Computing the nearest by `dist` (or sorting) in the helper keeps the mirror exact.

### 3. (info) Notify is now emitted every stabilize round

Previously the notify was gated inside `if let Some(_) = info.predecessor`; now it fires whenever the post-update head ≠ self. This is correct and matches the spec's unconditional `notify' = IF Len(next)=0 THEN <<>> ELSE <<Head(next)>>`, but it does increase steady-state `NotifyPredecessor` traffic compared to the old default — worth a line in the PR description.

RyanKung:
## Re-review — test coverage & documentation

Both prior findings are resolved: `join_dht` restores the best-effort `handle_dht_events` (with a churn/`Connected` comment), and the spec mirror now derives the old head via `min_by_key(dist)` with a dedicated unsorted-input test plus a `successor::extend` sort test. Focusing now on coverage and docs as requested.

### Test coverage

**Pure `stabilize()` — thorough.** The eight `correct_stabilize_*` spec-mirror cases now cover every branch: improved / farther / self / absent predecessor, unsorted current successors, self-dedup + K-truncation, `but_last` (ignore-last-topo), and the empty no-op. Each asserts both the resulting successor list and the exact action vector against independent spec helpers. This is the right shape.

**Gap 1 — the #388 handler delivery has no direct test.** The actual bug site (`handlers/mod.rs:144`, `RemoteAction(did, Notify(pred)) => send NotifyPredecessorSend{pred} to did`) is only covered transitively: `test_stabilization_once` asserts `node2`'s predecessor becomes `node1` after `node1.stabilize()`, which would fail if the notify were misdelivered. That's real but indirect — it also depends on the query/round-trip path and async delivery timing. Since the bug was precisely a swapped *(target, payload)*, a direct handler-level unit test (in the `test_message_handler.rs` style: feed `MultiActions([RemoteAction(target, Notify(pred))])` to `handle_dht_events` and assert the sent message's destination = `target` and `NotifyPredecessorSend.did = pred`) would pin the semantics independently and protect against re-inversion.

**Gap 2 — the restored best-effort branch is untested.** The whole point of keeping `if let Err(e) = self.handle_dht_events(..) { warn } Ok(())` is that a failing convergence send must not abort the join nor suppress the `Connected` event. There's no test exercising that error path. With the `dummy` transport you can fault-inject a send failure and assert `join_dht` still returns `Ok` (and, ideally at the `on_data_channel_open` level, that `Connected` is still emitted). Otherwise this resilience can silently rot back to the propagating behavior.

### Documentation

**Doc 1 (the important one) — `RemoteAction::Notify(Did)` doc is stale and encodes the exact confusion this PR fixes.** `crates/core/src/dht/chord.rs:97`:

```rust
/// Let `did_a` [notify](Chord::notify) `did_b`.
Notify(Did),
```

By the enum's own convention (`did_a` = the `PeerRingAction::RemoteAction(Target, _)` target, `did_b` = this field), the handler actually does the opposite of what this reads: `did_a`/Target is the **recipient** and `did_b` is the **predecessor payload** announced to it ("send `did_b` to `did_a`"). The phrasing "let did_a notify did_b" inverts actor and recipient — which is the precise misreading behind the original misdelivery. Please update the variant doc to state the delivery direction explicitly, so the canonical definition matches the (now-correct) handler comment.

**Doc 2 — production `CorrectChord::stabilize` doesn't reference its spec.** `chord.rs:596` carries only "Stabilize Operation: Perform stabilization for the successor list." The correctness of this function is defined by the `CorrectStabilize` TLA+ operator, which lives in `tests/default/dht_convergence.rs` — `stabilization.rs` already links there, but the function that *implements* the operator does not. Add the spec reference plus a line on the non-obvious choices the body makes (old head captured *before* the update; notify keyed on the *post-update* head ≠ self; `but_last` drops the last topo successor). The spec helper functions themselves are well documented — this is just about pointing the production operator at its specification.

Net: correctness coverage is strong for the pure operator; the two gaps are the handler-delivery unit test (the literal bug site) and the best-effort error path, and the main doc issue is the misleading `Notify` variant comment.

RyanKung:
## Re-review

This revision adds a fair amount: configurable finger-table size, proactive `connect_dht_peer` establishment in the handlers, and a much stronger controlled-schedule harness. Reviewing the new surface plus the status of prior findings.

### Resolved since last round
- **Best-effort error path now tested.** `test_join_dht_keeps_local_join_when_convergence_send_fails` drives `set_max_message_size(0)` so follow-up sends fail, then asserts `join_dht` still returns `Ok` and the local successor list keeps `node2`. That closes the exact gap I raised.
- **`SyncSuccessor` no longer joins peers it isn't connected to** (`get_connection(*peer).is_some()` guard) — removes the abort-the-loop hazard.
- **Schedule determinism hardened**: `ControlledDeliveryGuard` (RAII enable/disable, survives panics), no `random_delay` while controlled, plus `MAX_DELIVERIES_PER_ROUND` / `MAX_STAGNANT_ROUNDS` progress guards and structured `RoundTrace` diagnostics. This is the right shape for reproducible convergence testing.

### New production logic — looks correct
The `connect_dht_peer` deferral (Notify/QueryForSuccessorList: if `!is_connected`, connect and `return Ok(())` instead of sending) is safe for convergence: the unconditional per-round notify re-emits the action once the connection exists, and the `Stabilization` branch pre-connects predecessor + successors so the subsequent actions usually send immediately. `connect_dht_peer` treats `Err(AlreadyConnected)` as success, which correctly handles the connect race under `MultiActions`' parallel `join_all`. No correctness issue found here.

### Documentation — the two items from last round are still open

1. **`RemoteAction::Notify(Did)` doc is unchanged and still inverted** (`chord.rs:98`): `/// Let did_a [notify](Chord::notify) did_b.` Per the enum's own convention `did_a` is the target/recipient and `did_b` the predecessor payload, so the handler does the opposite of what this reads ("send `did_b` to `did_a`"). This is the canonical definition of the variant whose misdelivery this very PR fixes — leaving it phrased the inverted way is the most likely way the bug comes back. Please correct it.

2. **`CorrectChord::stabilize` still carries only "Stabilize Operation / Perform stabilization for the successor list"** (`chord.rs:618`), with no link to the `CorrectStabilize` operator that defines its correctness (the operator and its prose live in `tests/default/dht_convergence.rs`; `stabilization.rs` already links there). Add the spec reference and a line on the non-obvious choices (old head captured pre-update; notify keyed on post-update head ≠ self; `but_last` drops the last topo successor).

### New, smaller doc/robustness notes
- **`connect_dht_peer` (`mod.rs:70`) has no doc comment**, yet it encodes real policy: skip self, skip already-connected, treat `AlreadyConnected` as success. Given this is now the single funnel for DHT-driven connects, a short contract doc (idempotent, never errors on an existing connection) is worth it. (It also returns `Result<bool>` that every caller discards — `Result<()>` unless the flag is intended for future use.)
- **The `Stabilization` branch pre-connects predecessor + all successors (`connection.rs:52`) with no comment.** It's non-obvious new behavior (establish overlay links before applying `PeerRing::stabilize` so the resulting Notify/Query don't defer); one line of rationale would help the next reader.
- **`fix_fingers` / `DEFAULT_FINGER_TABLE_SIZE`**: the size feeds `Did::from(2^index)`, which is only well-defined for a table ≤ 160 over a 160-bit `Did` (index 160 wraps to `Did(0)`). The new `dht_finger_table_size` builder accepts any `usize`. Worth either documenting the ≤160 expectation on the setter or clamping; the `== 0` guard is good, the upper end is unguarded. Low priority since defaults/tests are in range.

### Coverage
Strong now. Pure `stabilize()` has full branch coverage; the best-effort path is tested; the schedule harness asserts full configured DHT convergence with stagnation detection. The one still-missing piece is a *direct* unit test of the `Notify` handler arm asserting destination = `did` and payload = `predecessor` (today it's only transitive via `test_stabilization_once`'s predecessor assertion) — lower priority now, but it's the literal line that was buggy, so a 5-line direct assertion would lock it.

RyanKung:
## Re-review

All outstanding items from the previous rounds are resolved, verified against the files:

- **`RemoteAction::Notify` doc corrected** — now "Send a predecessor notification to `did_a`" with `did_a` = recipient and the field = announced predecessor. No longer inverted.
- **`CorrectChord::stabilize` doc** now links the `CorrectStabilize` operator and explains the three non-obvious choices (pre-update old head, `but_last`, post-update head ≠ self).
- **`connect_dht_peer`** has a contract doc (idempotent, `AlreadyConnected`-as-success for the `MultiActions` race) and now returns `Result<()>` (the bool was discarded everywhere).
- **`Stabilization` branch** has a rationale comment for pre-connecting topo candidates.
- **Finger-table size** is clamped to `DEFAULT_FINGER_TABLE_SIZE` in `FingerTable::new` (with the `2^index` wrap explained on the setter and builder), and `test_finger_table_size_bounds` pins both the clamp and the `0` case.
- **Direct Notify-delivery unit test** added (`test_handle_dht_notify_remote_action_sends_predecessor_to_target`): feeds `RemoteAction(node2, Notify(node3))` and asserts the message lands at `node2` carrying `NotifyPredecessorSend{ did: node3 }` — the exact (target, payload) that was the original bug.

No new issues in the latest changes (doc-only additions, the clamp, and the signature change are all safe). Coverage and documentation are both in good shape now: the pure operator has full branch coverage and a spec link, the handler delivery and best-effort join paths each have a direct test, and the controlled schedule asserts full convergence with stagnation/budget guards. Nothing further from me on correctness, coverage, or docs.

PR #622, opened 2026-06-25
Fix strict lint debt and wasm/IDB robustness

### LINKED ISSUE #615: Track Rustacean strict lint debt outside PR #609
## Context

PR #609 exposed existing Rustacean strict-lint debt outside the Entry/storage changes. This issue tracks that unrelated debt separately so #609 can stay focused on the storage model and sync correctness work.

The Rustacean strict lint set is:

```bash
-D warnings \
-D clippy::unwrap_used \
-D clippy::expect_used \
-D clippy::panic \
-D clippy::todo \
-D clippy::unimplemented \
-D clippy::indexing_slicing
```

## Current production baseline

Captured locally on 2026-06-26 from `refactor/entry-storage-model-607` after PR #609 head `8638429`.

Runnable commands:

```bash
cargo clippy -p rings-derive --lib --no-deps -- \
  -D warnings \
  -D clippy::unwrap_used \
  -D clippy::expect_used \
  -D clippy::panic \
  -D clippy::todo \
  -D clippy::unimplemented \
  -D clippy::indexing_slicing

cargo clippy -p rings-transport --lib --no-deps -- \
  -D warnings \
  -D clippy::unwrap_used \
  -D clippy::expect_used \
  -D clippy::panic \
  -D clippy::todo \
  -D clippy::unimplemented \
  -D clippy::indexing_slicing

cargo clippy -p rings-core --lib --features dummy --no-deps -- \
  -D warnings \
  -D clippy::unwrap_used \
  -D clippy::expect_used \
  -D clippy::panic \
  -D clippy::todo \
  -D clippy::unimplemented \
  -D clippy::indexing_slicing
```

Counts by crate:

- `rings-derive`: 1 production violation
- `rings-transport`: 7 production violations
- `rings-core --features dummy --lib`: 44 production violations

## Violations

```text
rings-derive
- crates/derive/src/lib.rs:16 clippy::unimplemented

rings-transport
- crates/transport/src/core/pool.rs:85 clippy::expect_used
- crates/transport/src/core/pool.rs:92 clippy::indexing_slicing
- crates/transport/src/ice_server.rs:79 clippy::unwrap_used
- crates/transport/src/notifier.rs:28 clippy::unwrap_used
- crates/transport/src/notifier.rs:38 clippy::unimplemented
- crates/transport/src/notifier.rs:64 clippy::unimplemented
- crates/transport/src/notifier.rs:109 clippy::unwrap_used

rings-core
- crates/core/src/dht/did.rs:215 clippy::unwrap_used
- crates/core/src/dht/finger.rs:57 clippy::unwrap_used
- crates/core/src/dht/finger.rs:67 clippy::indexing_slicing
- crates/core/src/dht/finger.rs:75 clippy::indexing_slicing
- crates/core/src/dht/finger.rs:89 clippy::indexing_slicing
- crates/core/src/dht/finger.rs:109 clippy::unwrap_used
- crates/core/src/dht/finger.rs:125 clippy::indexing_slicing
- crates/core/src/dht/finger.rs:141 clippy::indexing_slicing
- crates/core/src/dht/finger.rs:147 clippy::indexing_slicing
- crates/core/src/dht/finger.rs:161 clippy::indexing_slicing
- crates/core/src/dht/successor.rs:117 clippy::indexing_slicing
- crates/core/src/ecc/elgamal/impls/secp256k1.rs:131 clippy::indexing_slicing
- crates/core/src/ecc/elgamal/impls/secp256k1.rs:153 clippy::indexing_slicing
- crates/core/src/ecc/elgamal/impls/secp256k1.rs:157 clippy::indexing_slicing
- crates/core/src/ecc/elgamal/impls/secp256k1.rs:184 clippy::panic
- crates/core/src/ecc/group.rs:586 clippy::expect_used
- crates/core/src/ecc/keys.rs:424 clippy::expect_used
- crates/core/src/ecc/keys.rs:427 clippy::expect_used
- crates/core/src/ecc/keys.rs:427 clippy::indexing_slicing
- crates/core/src/ecc/signers/bip137.rs:81 clippy::indexing_slicing
- crates/core/src/ecc/signers/bip137.rs:82 clippy::indexing_slicing
- crates/core/src/ecc/signers/bls.rs:125 clippy::unwrap_used
- crates/core/src/ecc/signers/bls.rs:134 clippy::unwrap_used
- crates/core/src/ecc/signers/ed25519.rs:40 clippy::unwrap_used
- crates/core/src/ecc/signers/secp256r1.rs:65 clippy::expect_used
- crates/core/src/ecc/signers/secp256r1.rs:73 clippy::panic
- crates/core/src/ecc/types.rs:61 clippy::unwrap_used
- crates/core/src/ecc/types.rs:63 clippy::unwrap_used
- crates/core/src/ecc/mod.rs:107 clippy::expect_used
- crates/core/src/ecc/mod.rs:158 clippy::unwrap_used
- crates/core/src/message/protocols/relay.rs:72 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:92 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:92 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:109 clippy::unwrap_used
- crates/core/src/message/protocols/relay.rs:115 clippy::unwrap_used
- crates/core/src/message/protocols/relay.rs:136 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:137 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:138 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:141 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:141 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:142 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:142 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:143 clippy::indexing_slicing
- crates/core/src/message/protocols/relay.rs:146 clippy::unwrap_used
```

## Additional observation

`cargo clippy --workspace --all-targets --all-features` does not currently reach this lint baseline cleanly. With the same strict lint set, `rings-transport --all-features` first hits compile/type errors around `crates/transport/src/connections/dummy/mod.rs`, `crates/transport/src/connections/native_webrtc/mod.rs`, and duplicate notifier implementations. That feature-combination problem should either be fixed or made explicitly unrepresentable before strict all-feature CI can be required.

## Acceptance criteria

- Replace production `unimplemented!`, `panic!`, `unwrap`, `expect`, and unchecked indexing/slicing above with typed errors, checked accessors, total transformations, or proven constructors.
- Preserve the domain propositions in names, not comments only: e.g. checked finger-slot access, non-empty relay path, total successor lookup, typed notifier state transitions, and checked cryptographic encodings.
- Add focused tests for each changed domain boundary.
- The three crate-level production commands above pass.
- After the production baseline passes, extend the cleanup to `--all-targets` test/support code or explicitly scope test assertion panics away from production lint policy.
- Decide whether `--all-features` is a valid feature combination. If yes, make it compile; if no, encode mutually exclusive feature constraints and document the supported strict-lint command.

RyanKung:
## Summary
- Enable staged workspace strict lint policy for `rings-derive`, `rings-transport`, and `rings-core`.
- Replace production `unwrap`/`expect`/`panic`/`unimplemented`/unchecked indexing paths with typed errors, checked access, or non-panicking fallbacks.
- Reject invalid native+wasm transport feature combinations with a crate-root `#[cfg(...)] compile_error!` guard.
- Keep the wasm stabilizer loop declarative by routing `poll!` through a panic-free `js_utils::spawn_interval` adapter.
- Harden IDB LRU pruning by making `last_visit_time` per-key monotonic across same-millisecond or backward-clock revisits.
- Clean up finger-table access/write APIs and share ICE-server parse degradation through `parse_ice_servers_or_warn`.

## Root Cause
Issue #615 tracked strict Rust lint debt left outside PR #609. The affected crates had production paths that could still panic under `clippy::unwrap_used`, `clippy::expect_used`, `clippy::panic`, `clippy::unimplemented`, and `clippy::indexing_slicing`. The cleanup also exposed an unsupported `--all-features` transport configuration, which is now rejected explicitly at crate compile time instead of failing later through internal type errors.

## Scope Notes
- The wasm stabilizer rescheduling now keeps timer effects at the wasm adapter boundary: `poll!` delegates to `js_utils::spawn_interval`, which uses `window_sleep` rather than hand-rolled `Closure`/`forget` recursion in DHT code.
- The IDB visit timestamp change protects the `last_visit_time` prune index per key, so a just-touched entry is not made artificially old by equal-ms timestamps or wall-clock rollback.
- Finger-table cleanup removed `get_ref`, `Index`, and `EMPTY_FINGER_SLOT`; bounded writes now go through `write_slot`.
- Transport ICE parsing now has one shared warn-and-empty fallback helper across dummy, native, and web-sys transports.

## Validation
- `cargo clippy -p rings-transport --lib --features dummy --no-deps -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `cargo clippy -p rings-transport --lib --features native-webrtc --no-deps -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `cargo clippy -p rings-transport --lib --no-default-features --features web-sys-webrtc --target wasm32-unknown-unknown --no-deps -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `cargo clippy -p rings-core --lib --features dummy --no-deps -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `cargo clippy -p rings-core --lib --no-default-features --features wasm --target wasm32-unknown-unknown --no-deps -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `cargo test -p rings-core --lib --features dummy`
- `cargo test -p rings-transport --lib --features dummy`
- `cargo test -p rings-core --lib --no-default-features --features wasm --target wasm32-unknown-unknown --no-run`
- `git diff --check`

Closes #615


RyanKung:
## Review — strict-lint debt cleanup (#615)

High-quality, careful pass. The lint policy is wired the right way — `[workspace.lints]` (rust `warnings=deny` + the clippy strict set) with per-crate `[lints] workspace = true` for `core`/`derive`/`transport`, and production strictness kept while test ergonomics are scoped out via `#![cfg_attr(test, allow(...))]`. That matches the config-driven approach from #615 and means plain `cargo clippy` now enforces it (no CI-only flags).

The risk in a PR like this is mechanical `unwrap`→fallback conversions silently changing behavior. I went through the conversions; the great majority are either semantically identical or strict improvements (panic → typed error, several with new tests: `SuccessorIndexOutOfBounds`, `RoundRobinPoolEmpty`, `Secp256k1PointLiftFailed`, the empty-path relay test). I verified the two byte-layout-sensitive refactors are exactly equal to the originals:
- `From<VerifyingKey> for PublicKey<33>`: old reverse/push/reverse and new `[0u8;33]` + `zip(skip(1))` both produce `[0, k0..k31]`.
- `bip137 varint_buf_num`: old `[255] + n_le[0..4] + (n>>32)_le[0..4]` equals new `[255] + n.to_le_bytes()` = `[255, n_le[0..8]]`.
- `RoundRobinPool::select`: `fetch_update` (returns prev) → `fetch_add(1) % len` yields the same index sequence; existing ordering test still holds.
- `lift_x`: recursion-with-`panic!` → loop returning `Result`, same candidate set (byte0 = bias, 254..=1), now `Err(Secp256k1PointLiftFailed)` instead of panic, propagated through `str_to_affine`/`TryFrom<Plaintext>`.

### Findings (all minor; none block)

1. **`MessageRelay::origin_sender()` now silently falls back to `self.destination` on an empty path** (was panic). The PR correctly switched every production caller I can see (`connection.rs:87,108`, `stabilization.rs:51`, the `report` builder) to the checked `try_origin_sender()?`, so the hot paths are safe. Residual: `origin_sender()` stays public with that silent fallback, and this branch is based on `master` — please confirm no *other* caller (e.g. in the node crate) relies on the old panic semantics, since substituting `destination` for the origin could misroute rather than fail loudly. Worth a doc note on the fallback.

2. **`let _ = field.set_b32(&data)` in `str_to_field`** drops an invariant the old `assert!(field.set_b32(...))` enforced: that the encoded 32 bytes are below the secp256k1 field modulus. It is currently safe only because the fixed prefix `[0xFF, 0x52, …]` (marker `0x52 < 0xFF`) keeps the value under `p`. Ignoring the bool silently relies on that; keep a `debug_assert!` or a comment pinning it so a future change to the marker/encoding can't silently produce an overflowed field element.

3. **Silent ICE-config degradation:** `WebrtcTransport::new` / `WebSysWebrtcTransport::new` / `DummyTransport::new` now log and continue with an empty ICE list on invalid config instead of panicking. Since `new` returns `Self`, error propagation isn't available without an API change; acceptable, but an empty ICE list degrades NAT traversal opaquely — consider validating ICE config at load (`try_new`) so bad config surfaces loudly somewhere.

Minor/positive: `derive`'s `unimplemented!` → `syn::Error::to_compile_error()` (proper macro diagnostic instead of a panic); the `transport/build.rs` guard cleanly rejects `web-sys-webrtc` + native feature combos (the #615 `--all-features` observation); poisoned-lock recovery (`unwrap_or_else(|p| p.into_inner())`) is reasonable for the dummy/notifier state; the finger-table `get_mut` no-ops are safe because indices are bounded by `size`. Note for #623: the non-webrtc `Notifier::set_timeout_ms` now spawns a thread per timeout — fine for that rarely-used fallback cfg, but worth remembering on the low-resource track.

### Caveat
Reviewed from the diff; I could not run the strict lint set or check out the branch (repo SSH access denied here). The PR documents running the exact per-crate strict commands, and CI on the branch is the enforcement floor.

**Verdict:** correct, semantically-preserving lint cleanup with good test coverage and the right enforcement mechanism. LGTM pending the origin_sender() caller audit (F1) and CI green.

RyanKung:
## Re-review — `ea7ec568` ("Fix wasm stabilizer scheduling")

Round-1 items addressed:

- **F2 (set_b32 invariant)** — resolved: `str_to_field` now carries the comment pinning why ignoring the `set_b32` bool is safe (`FIELD_ENCODING_MARKER` sits below the secp256k1 modulus prefix, so the encoded candidate always fits the field). A `debug_assert!(field.set_b32(&data))` would additionally catch a future marker/encoding change at runtime, but the comment satisfies the ask.
- **F1 (origin_sender fallback)** — resolved: `origin_sender()` now documents the empty-path fallback to `destination` and points strict callers to `try_origin_sender()`. Combined with the handlers already using the checked form, the fallback is now explicit rather than surprising.

### The wasm stabilizer rewrite — correct, and a real fix

The previous `loop { window_sleep().await; spawn_local(stabilize) }` form depended on something continuously driving the `wait()` future. The new form registers a self-perpetuating `setTimeout` callback with the JS event loop instead, which is the right wasm idiom — it self-drives without an external poller. Mechanics check out:

- The closure takes `js_sys::Function` and is re-scheduled via `set_timeout_with_callback_and_timeout_and_arguments(callback, millis, [callback])`, so JS invokes `closure(closure)` each tick and the inner `func` re-arms the next timeout — the standard "pass self as arg to re-schedule" pattern.
- `Closure::wrap(Box<dyn FnMut(..)>)` is the correct wrapper for a repeatedly-invoked callback, and `func.forget()` keeps it alive (required, since it's handed to JS).
- Failure paths (`global()` is `None`, or `set_timeout` errors) log and stop instead of panicking.

Notes (non-blocking):
- `func.forget()` permanently leaks one closure. Fine for a lifetime-of-node stabilization daemon (called once), but worth remembering on the low-resource track (#623) if `wait()` could ever be invoked repeatedly.
- If a mid-loop re-schedule fails, stabilization silently stops for good (degrades rather than crashes — acceptable, but it's a silent stop).
- Same as before: `spawn_local(stabilize)` is fire-and-forget, so a stabilize slower than `interval` can overlap — unchanged from the prior behavior, not a regression.

### Caveat
Reviewed from the diff; I couldn't run the strict lint set or check out the branch (repo SSH access denied here). CI on the branch is the enforcement floor.

**Verdict:** F1/F2 resolved, the wasm scheduling rewrite is correct and an improvement. F3 (silent ICE-config degradation) remains as the only earlier note, and it's acceptable given the `new() -> Self` signature. LGTM pending CI green.

RyanKung:
### Suggestion: replace the `Closure` + `forget()` + recursive-`setTimeout` with `spawn_local` + an async loop

The current wasm `Stabilizer::wait` hand-rolls a self-rescheduling `setTimeout` callback and `forget()`s the `Closure`. That works, but `forget()` leaks the closure with **no way to cancel** stabilization, and it's ~50 lines of JS-interop boilerplate. The same behavior can be expressed by spawning an async loop:

```rust
impl Stabilizer {
    /// Run stabilization in a loop.
    pub async fn wait(self: Arc<Self>, interval: Duration) {
        let millis = i32::try_from(interval.as_millis()).unwrap_or(i32::MAX);
        spawn_local(async move {
            loop {
                if let Err(error) = window_sleep(millis).await {
                    tracing::error!("stabilization interval failed: {error:?}");
                    return;
                }
                self.stabilize()
                    .await
                    .unwrap_or_else(|e| tracing::error!("failed to stabilize {e:?}"));
            }
        });
    }
}
```

**Why this works in the browser (and is non-blocking):** `spawn_local` hands the task to the main-thread wasm-bindgen-futures executor, which re-polls via microtasks. Each iteration `window_sleep(millis).await` is a `JsFuture` over a `setTimeout`-backed Promise, so it yields control to the JS event loop and is re-polled when the timer fires — i.e. it uses the *same* `setTimeout` + event-loop machinery as the current recursive-callback version, just expressed as an awaited future. The one rule is that each loop turn must await a real JS-backed yield point (`window_sleep` qualifies); a loop awaiting only already-ready futures would starve the event loop, but that's not the case here.

**Why it's also more correct than the pre-`ea7ec568` async loop:** the earlier `loop` lived inside `wait()` and only ran while someone kept polling `wait()`'s future. Here the loop is owned by the executor via `spawn_local`, so `wait()` returns immediately (fire-and-forget) and stabilization self-perpetuates regardless of the caller — the property the `setTimeout`+`forget` version was reaching for, without the closure plumbing.

**Optional — make it cancellable** (the real downside of `forget()` is that stabilization can never be stopped). Wrap the loop in `futures::future::Abortable` and keep the `AbortHandle` on the `Stabilizer`/swarm so shutdown can stop it cleanly:

```rust
let (task, handle) = futures::future::abortable(async move { loop { /* … */ } });
spawn_local(async move { let _ = task.await; });
// store `handle`; call handle.abort() on teardown
```

Note: `futures` is already a dependency (`crates/core/Cargo.toml`). `window_sleep` is the safe timer choice since it's already in use; `futures_timer::Delay` (also already a dep) could unify the native/wasm paths if its wasm path is confirmed. I couldn't run this on the branch (no fetch access here), so please validate, but the mechanism is the standard wasm `spawn_local` + awaited-`setTimeout` pattern already used elsewhere in the crate (see `extension/transport/wt.rs`).

RyanKung:
## Rustacean-standard conformance review

Reviewed against the skill's assertions. State the fact, name the witness.

### Satisfied
- **A9 (panic is a failed proof) — the core of the PR, satisfied.** Production panic paths become fallible returns with named variants: `lift_x` → `Result<Affine>` + `Error::Secp256k1PointLiftFailed`; `SuccessorReader::get` → `Error::SuccessorIndexOutOfBounds { index, len }`; `RoundRobinPool::select` → `Error::RoundRobinPoolEmpty`; relay `next_hop`/origin → `Error::CannotInferNextHop`; `secp256r1::sign` panic-closure → `Result`; `derive`'s `unimplemented!` → `syn::Error::to_compile_error()`. Indexing replaced by slice patterns / `get`.
- **A9 test exception used correctly.** `#![cfg_attr(test, allow(clippy::unwrap_used, …))]` in `core`/`transport` `lib.rs` keeps assertion macros legal in tests while production stays strict — exactly the assertion-9 carve-out.
- **A8 (errors are algebraic).** New failures are `thiserror` variants local to the domain enum; no `anyhow`/`Box<dyn Error>`/stringly substitution.
- **A5 (quality is witnessed).** Each new variant has a proposition-named test: `get_out_of_bounds_returns_typed_error`, `empty_rr_pool_returns_typed_error`, `parsing_rejects_missing_host`, `empty_path_origin_sender_is_checked`. The enforcement floor is wired into config — `[workspace.lints]` (rust `warnings=deny` + the six clippy denies) with per-crate `[lints] workspace = true` — so plain `cargo clippy` is the witness, not a CI-only flag.
- **A11 (clone/copy).** No `.clone()` smells introduced; finger access uses `.copied().flatten()` on `Did` (Copy, scalar-like).

### Minor gaps (assertion → witness)
1. **A3 (effects at named boundaries).** The wasm stabilizer (`ea7ec568`) places `Closure` + `set_timeout` + `global()` timer plumbing directly in `dht/stabilization.rs`, and `notifier.rs::set_timeout_ms` (non-webrtc) calls `std::thread::spawn` inside a core type. Time/scheduling/threads are effects; A3 wants them at adapter boundaries (the `window_sleep` adapter already exists in `utils::js_utils`). This is the same reason the `spawn_local` + `window_sleep` form suggested earlier is preferable here — it keeps the timer effect in the adapter and leaves the core with a pure loop.
2. **A2 (repetition is evidence — three times).** The ICE fallback `match IceServer::vec_from_str(..) { Ok(..) => .., Err(error) => { warn; Vec::new() } }` is duplicated across `dummy`, `native_webrtc`, and `web_sys_webrtc` transports — three copies, one proposition. And the 3-arm `Window | WorkerGlobal | ServiceWorkerGlobal` `set_timeout` dispatch is duplicated between `set_stabilize_timeout` and `window_sleep`. Abstract each to one source.
3. **A5/A10 (test the model, name the incompleteness).** `Error::Secp256k1PointLiftFailed` has no test exercising the Err arm, and the comment asserts it is effectively unreachable ("at least one of the 254 candidates should lift"). Either force the Err in a test or label it explicitly as an unreachable-but-typed safeguard (the typed return is correct; the model claim should be the named witness).
4. **A9 nuance (unchecked casts).** The enforced `[workspace.lints]` set is the rubric's six lints but omits cast lints, so bounded `as` casts remain (`k as u32` in `finger.join`, `x.len() as u8` in `str_to_field`). They are safe by bound (k < size ≤ 160; chunk ≤ 29) and pre-existing, but A9 forbids unchecked numeric casts — prefer `u32::try_from`/a documented bound if the set is later tightened.
5. **A12 (logic deserves a name).** `matches!(window, [left, right] if left == right)` in relay path validation is a naked comparison; a `has_adjacent_duplicate()` predicate would read as the proposition. Local and trivial, so allowed — noting for completeness.

### Review-protocol note
The skill's protocol runs `cargo clippy --all-targets --all-features`. That command is, by this PR's own design, no longer representable: `transport/build.rs` makes `web-sys-webrtc` + native a hard build error. So the rubric's `--all-features` step does not apply here; the PR's per-target strict commands plus the `[workspace.lints]` config are the correct adaptation. (I could not run the strict set or check out the branch here — no fetch access; CI is the enforcement floor.)

**Verdict:** strongly conforms to the rustacean standard — it *is* the standard's panic-removal + algebraic-error program, with tests witnessing each new variant and enforcement moved into source config. The gaps are minor: effect placement in the stabilizer/notifier (A3), the three-way ICE-parse and global-dispatch repetition (A2), one untested/unreachable Err (A5/A10), and residual bounded `as` casts (A9).

RyanKung:
## Re-review — `7f9ec8f8` ("Use async wasm stabilizer loop")

The wasm stabilizer is now the `spawn_local` + async-loop form (−53/+10):

```rust
pub async fn wait(self: Arc<Self>, interval: Duration) {
    let millis = i32::try_from(interval.as_millis()).unwrap_or(i32::MAX);
    spawn_local(async move {
        loop {
            if let Err(error) = window_sleep(millis).await {
                tracing::error!("failed to wait for stabilization interval {:?}", error);
                return;
            }
            self.stabilize().await.unwrap_or_else(|e| tracing::error!("failed to stabilize {:?}", e));
        }
    });
}
```

Correct and a net improvement:
- **Browser-safe / non-blocking:** `spawn_local` runs on the main-thread wasm-bindgen-futures executor; each turn `window_sleep(millis).await` is a `setTimeout`-backed `JsFuture` that yields to the JS event loop, so it self-drives via the same timer + microtask machinery as the old recursive callback, without freezing the tab.
- **Fire-and-forget preserved:** the task is owned by the executor, so `wait()` returns immediately and stabilization self-perpetuates regardless of who polls `wait()` — the property the `setTimeout`+`forget` version was reaching for, now without the `Closure`/`forget()`/`unchecked_ref` plumbing.
- **Resolves prior items:** removes the uncancellable leaked `Closure` (`forget()` gone), and closes the **rustacean A3** gap I raised — the timer effect now lives entirely in the `window_sleep` adapter (`utils::js_utils`) and `stabilization.rs` holds only the loop. It also removes the `set_stabilize_timeout` vs `window_sleep` global-dispatch duplication (**A2**).
- **Minor positive behavior change:** stabilize is now `await`ed inline before the next sleep, so runs no longer overlap (period = interval + stabilize time, no pile-up) — preferable to the previous fixed-period spawn that could overlap.

Optional (unchanged from before): if clean shutdown is ever wanted, wrap the loop in `futures::future::Abortable` and hold the `AbortHandle`; today the spawned task runs for the node lifetime.

Still-open minor rustacean items from the prior pass (none blocking): three-way `IceServer::vec_from_str` warn-and-`Vec::new()` repetition across the dummy/native/web_sys transports (**A2**); the `Error::Secp256k1PointLiftFailed` Err arm is untested and documented as effectively unreachable (**A5/A10** — name it as a typed-but-unreachable safeguard or force it in a test); residual bounded `as` casts not caught by the enforced lint set (**A9**).

**Verdict:** the stabilizer change is clean and resolves the `forget()`/A3 concerns; the PR remains a strong, standard-conformant lint cleanup. (Reviewed from the diff; no branch fetch access here — CI is the strict-lint enforcement floor.)

crates/core/src/dht/finger.rs
```diff
@@ -12,6 +12,7 @@ use crate::dht::Did;
 
 /// Default number of Chord finger slots for a 160-bit `Did`.
 pub const DEFAULT_FINGER_TABLE_SIZE: usize = 160;
+const EMPTY_FINGER_SLOT: Option<Did> = None;
```
RyanKung: this is only used once, why not just `None`

crates/core/src/dht/stabilization.rs
```diff
@@ -264,22 +264,24 @@ mod stabilizer {
     use wasm_bindgen_futures::spawn_local;
 
     use super::*;
-    use crate::poll;
+    use crate::utils::js_utils::window_sleep;
```
RyanKung, line 266: you should implement this in poll mod, but not expose details in concrete implementation

RyanKung:
## Correction to my earlier reviews — I was too lenient

My prior comments called this "clean" and "strongly conforms". I withdraw that wording — acceptance isn't mine to declare, and I reviewed the lint-pass surface instead of the abstraction quality underneath. Several real smells were left, including ones the rustacean rubric explicitly targets:

### 1. Layering regression: the `poll!` interval abstraction was discarded, not fixed (A3)
`poll!` (`crates/core/src/macros/mod.rs:78`, `#[macro_export]`) is the codebase's "run a closure every `ttl`" interval abstraction. This PR removed its only caller and **inlined the scheduling mechanics** (`spawn_local` + `loop` + `window_sleep` + error handling) directly into `Stabilizer::wait` — a dht business method. Recurring scheduling is an effect/adapter concern (A3: effects at named boundaries), not business logic.

The root cause is that `poll!` is full of `.unwrap()`s, so rather than make the abstraction panic-free, the PR bypassed it. That also means `poll!` is now **orphaned dead code that still contains panic paths** (macro bodies escape the lint until expanded — so a strict-lint PR has, ironically, left an unexported-from-use, panic-laden macro behind). Either:
- make `poll!` (or a new `spawn_interval(millis, closure)` adapter in `utils::js_utils`) panic-free and keep `wait()` declarative — `poll!(stabilize, ttl)`; or
- delete `poll!` and own the inlining as the chosen design.

(I share the blame here — my `spawn_local + loop` suggestion pushed the inlining instead of restoring the abstraction.)

### 2. One-use const papering over an awkward API (A1/A12)
`const EMPTY_FINGER_SLOT: Option<Did> = None;` exists only so `get_ref` can return `self.finger.get(index).unwrap_or(&EMPTY_FINGER_SLOT)` — a named constant used exactly once to hand back `&None`. The deeper issue: `get_ref(&self) -> &Option<Did>` is a pointless borrow — `Did` is `Copy`, so `get(&self) -> Option<Did>` already returns it cheaply. The const and `get_ref` should both go; collapse to `get`. The lint fix preserved a needless distinction instead of removing it.

### 3. Copy-pasted checked-write idiom (A2)
`if let Some(slot) = self.finger.get_mut(i) { *slot = … }` is repeated across `set`, `join`, and `fix_finger` (and silently no-ops on out-of-bounds in each). Three+ copies of one proposition → one named method, e.g. `fn write_slot(&mut self, index: usize, value: Option<Did>)`, so the bounded-write rule has a single source instead of being scattered.

### 4. Lint-driven readability contortions (A9 nuance)
Several `indexing_slicing` fixes replaced a clear slice copy with a manual element-by-element loop: `From<VerifyingKey>` does `for (t, s) in data.iter_mut().skip(1).zip(key.to_bytes()) { *t = s }` where `data[1..].copy_from_slice(&key.to_bytes())` read plainly; `str_to_field` similarly turned three assignments + `copy_from_slice` into two zip-mutate loops. These pass the lint but read worse. A9 wants checked access *with a proof/bound* — a small `prefix_zero::<33>(bytes)` helper or a documented fixed-size `copy_from_slice` is cleaner than open-coding the copy.

### 5. (restated) Repetition across transports (A2)
The `IceServer::vec_from_str(..) => warn + Vec::new()` fallback is copy-pasted in dummy/native/web_sys `new()`. One `parse_ice_servers_or_warn` helper.

These are why I should not have signed it off as "clean." The panic-removal and typed-error work is still correct and valuable; but the abstraction/layering quality needs another pass before it meets the bar.

crates/transport/build.rs
```diff
@@ -0,0 +1,22 @@
+fn main() {
```
RyanKung: transport is a lib, I dont think build.rs is necessary

RyanKung:
## Re-review — `d67a9a1d` ("Address strict lint review quality issues")

Each item from my correction is resolved; witnesses below.

- **`poll!` abstraction restored and made panic-free (was the layering regression).** `poll!` no longer expands to 80+ lines of `Closure`/`forget`/`unwrap`/3-arm dispatch; it now expands to `$crate::utils::js_utils::spawn_interval($ttl, move || $future)`. The scheduling effect lives in the new `spawn_interval` **adapter** in `utils::js_utils` (`spawn_local(async { loop { window_sleep(millis).await?; task().await } })`, panic-free), and `Stabilizer::wait` is declarative again via `poll!(async-block, millis)`. This is better than what I suggested — the abstraction is kept *and* the panic paths are gone *and* the effect is at the adapter boundary. The orphaned-panic-laden-macro concern is gone too.
- **`EMPTY_FINGER_SLOT` and the `&Option<Did>` API removed.** The one-use const is deleted, `get_ref` is deleted, and `impl Index for FingerTable` (its only consumer) is deleted; `get(&self) -> Option<Did>` remains. The needless distinction is gone.
- **Checked-write idiom has one source.** `set`/`join`/`fix_finger` now call a single `FingerTable::write_slot(index, Option<Did>)` instead of three copies of `if let Some(slot) = get_mut`.
- **ICE-parse repetition unified.** `parse_ice_servers_or_warn(config, transport)` in `ice_server.rs`; `dummy`/`native_webrtc`/`web_sys_webrtc` `new()` all call it.
- **Bounded cast made checked.** `finger.join` replaced `pow(k as u32)` with `u32::try_from(k)` + `else { continue }`.
- **Bonus:** the `web-sys-webrtc` + native feature conflict moved from a `build.rs` (`eprintln!` + `process::exit(1)`) to a crate-root `#[cfg(...)] compile_error!` — a compile-time guard is more idiomatic than a build script, and `build.rs` is deleted.

One micro-nit (non-blocking): `write_slot` returns `bool` (slot existed?) but all three call sites use it as a statement, so the return is currently dead. Either consume it where an out-of-bounds write should be noticed, or make it `()`.

I reviewed from the diff (no branch fetch/strict-lint run here). The abstraction/layering objections I raised are addressed; acceptance is your call.

RyanKung:
## Re-review — `4c0a2c16` + `f5ff94f6`

**`4c0a2c16` ("Remove unused finger slot write result")** — resolves the micro-nit: `write_slot` now returns `()` instead of an unused `bool`. Done.

**`f5ff94f6` ("Make IDB visit timestamps monotonic")** — a real correctness improvement, with a couple of notes.

`last_visit_time` is the IndexedDB index `prune()` evicts by (`store.index("last_visit_time")`), i.e. the LRU key. The change
```rust
fn next_visit_time(previous: i64) -> i64 {
    chrono::Utc::now().timestamp_millis().max(previous.saturating_add(1))
}
```
correctly ensures a re-visit's recorded time is strictly greater than that entry's prior time even when `Utc::now()` returns the same millisecond or steps backward — so a just-touched entry can no longer be made to look old and be evicted prematurely. It's panic-free (`saturating_add`, `max`).

Notes (none blocking; acceptance is your call):
- **The guarantee is per-key, not global.** `next_visit_time` is monotonic against *this entry's* own previous value; two different entries visited in the same millisecond can still receive equal timestamps (their prune order is then arbitrary). That's the correct property for "don't evict a just-visited entry" and matches the title, but worth stating in a one-line `Post:`/`Invariant:` comment so the scope of "monotonic" is explicit (rustacean A4).
- **No test.** `next_visit_time` is now a pure free function, so it's unit-testable without IndexedDB — a regression test (`now <= previous => result == previous + 1`; `now > previous => result == now`) would witness the property (A5). None was added.
- **Scope.** This is a functional behavior change unrelated to the strict-lint-debt purpose of #622 (#615). It's small and self-contained, but it'd be cleaner as its own commit/PR rather than folded into the lint cleanup (keep-changes-minimal).

I reviewed from the diff (no branch fetch/strict-lint run here). The earlier abstraction findings remain resolved; the finger nit is closed; the IDB change is correct, with the per-key scope, missing test, and scope-creep notes above.

RyanKung:
## Stricter pass — layering & abstraction

The panic-removal/error-typing is solid and the earlier abstraction objections (poll!, EMPTY_FINGER_SLOT, write_slot, ICE helper, checked casts) are resolved. Looking specifically at layering and abstraction across the whole PR, these remain — facts and witnesses, acceptance is yours.

### Required: the IDB monotonic change must be reworked
`f5ff94f6` ("Make IDB visit timestamps monotonic") is a **functional behavior change** that has no place in a strict-lint-debt PR, and it ships without a contract or a test. Please:
- **Move it out of this PR** (its own commit/PR), or explicitly justify folding a behavior change into #615's lint cleanup — `keep-changes-minimal-to-goal`.
- **Name the invariant.** `next_visit_time` guarantees *per-key* monotonicity only (a re-visit beats that key's prior time); two keys touched in the same millisecond can still tie. State it: `// Post: result > previous; the prune index (last_visit_time) is strictly increasing per key under equal-ms or backward wall-clock.`
- **Add a test.** `next_visit_time` is now a pure free function: `now > previous => result == now`; `now <= previous => result == previous + 1`. No IndexedDB needed (A5).

### Layering / abstraction findings

1. **The `Global` three-way dispatch is open-coded at every call site instead of being a method on the enum (A2 + A3).** `js_utils::window_sleep` matches `Global::{Window, WorkerGlobal, ServiceWorkerGlobal}` and calls the same `set_timeout_with_callback_and_timeout_and_arguments_0(func, millis)` in all three arms; `notifier::set_timeout_at` does the *same* three-arm match for its `set_timeout`; the old `poll!` did it too. `schedule_sleep` only factored out the inner Promise body — the dispatch itself is still duplicated within `window_sleep` and again across modules. The web-sys variants don't share a Rust trait, so the right home is a method on the enum:
   ```rust
   impl Global {
       fn set_timeout_0(&self, cb: &js_sys::Function, millis: i32) -> Result<i32, JsValue> { /* one match */ }
   }
   ```
   Then `window_sleep` and `notifier` call `global.set_timeout_0(..)` — the platform dispatch lives in one place (the adapter), not scattered through callers.

2. **`Notifier::set_timeout_ms` (non-webrtc) spawns a raw OS thread per call (A3).** `std::thread::spawn(move || { sleep(millis); this.wake() })` puts a heavyweight, one-thread-per-timeout effect inside a low-level sync primitive. Threads belong at an outer/adapter layer, and thread-per-timeout is a resource smell (also relevant to #623). It was `unimplemented!()` before, so this is a newly introduced layering choice — prefer a shared timer/scheduler (or a single timer thread / runtime timer) over spawning per call, or scope this stub more explicitly as a degraded fallback.

3. **Poison-recovery lock idiom duplicated (A2).** `lock().unwrap_or_else(|poisoned| poisoned.into_inner())` is repeated in `Notifier::state`, `DummyConnection::remote_rand_id`, and `DummyConnection::connection_state`. Extract one `fn lock_recover<T>(m: &Mutex<T>) -> MutexGuard<'_, T>` so the poison policy has a single source.

4. **`str_to_field` open-codes the field-candidate encoding with two zip-mutate loops to dodge `indexing_slicing` (A12/readability).** The structured intent — "byte0 = 0xFF lift bias, byte1 = marker, byte2 = len, bytes3.. = chunk" — is now spread across two `iter_mut().zip(..)` loops. A small `fn encode_field_candidate(chunk: &[u8]) -> [u8; 32]` with a documented layout would carry the proposition far more clearly than the loops, and keep the bound proof local.

### Resolved (for the record)
`poll!` restored as a panic-free macro delegating to the `spawn_interval` adapter (effect at boundary); `EMPTY_FINGER_SLOT`/`get_ref`/`Index` removed; checked writes via `write_slot` (now `()`); `parse_ice_servers_or_warn` shared; `k as u32` → `u32::try_from`; feature conflict via crate-root `compile_error!`.

Reviewed from the diff; no branch fetch/strict-lint run here.

RyanKung:
## Correction: keep the IDB change here — but the PR description must match the actual scope

I withdraw the "move the IDB monotonic change out" requirement. Keeping it in this PR is fine. The real problem is that the **title/body no longer describe what the PR does** — and a PR whose description omits or misstates its scope is the issue, not the change's location.

**Required — update the PR title and body so they cover the real scope:**

1. **Fix the stale line.** The body says *"Add a `rings-transport` build-time guard for invalid native+wasm transport feature combinations."* That guard was removed — there is no longer a `build.rs`; the conflict is now a crate-root `#[cfg(...)] compile_error!`. Update this bullet (and the Root Cause mention of the `--all-features` failure) to describe the compile-time guard.

2. **Add the changes the body currently omits.** The PR is no longer only "strict lint debt"; it also contains, as consequences of the cleanup:
   - wasm stabilizer rescheduling rewritten — `poll!` now delegates to a panic-free `js_utils::spawn_interval` adapter instead of a hand-rolled `Closure`/`forget`/`setTimeout` recursion;
   - IDB LRU robustness — `last_visit_time` (the `prune` eviction index) made per-key monotonic via `next_visit_time`, so same-millisecond / backward-clock revisits can't evict a just-touched entry;
   - finger-table API cleanup — `get_ref`/`Index`/`EMPTY_FINGER_SLOT` removed, checked writes via `write_slot`;
   - shared `parse_ice_servers_or_warn` across the three transports.

   The title ("Fix strict Rust lint debt") should be broadened accordingly, or the body should make explicit that the lint cleanup entailed these refactors/fixes.

3. The `next_visit_time` change still needs its **`Post:`/`Invariant:` comment** (per-key monotonic) and a **unit test** (it's a pure function) — independent of where it lives.

The layering/abstraction findings in my previous comment (the open-coded `Global` three-way dispatch → method on the enum; `Notifier::set_timeout_ms` spawning a thread per timeout; the duplicated poison-recovery lock idiom; the `str_to_field` zip-loop encoding) still stand.

RyanKung:
## Re-review — `07d61cd3` ("Address latest strict lint review feedback")

Body/title and every prior finding are addressed; one new gap.

**Description now matches scope.** Title broadened to "…and wasm/IDB robustness"; the Summary lists all six areas, a Scope Notes section documents the wasm-stabilizer / IDB / finger / ICE changes, and the stale "build-time guard" line is corrected to the crate-root `compile_error!`. The body requirement is met.

**Strict findings resolved (witnesses):**
- **`Global` dispatch → method.** `impl Global { fn set_timeout_0(&self, cb, millis) -> Result<i32, JsValue> }` added (in both the core and transport `js_utils` copies); `window_sleep` collapses to a single `Some(global) => … global.set_timeout_0(..)` arm, and `notifier::set_timeout_at` likewise. The three-way dispatch lives in one place per enum now.
- **Notifier timer (this is the strong one).** Thread-per-timeout is gone; `set_timeout_ms` routes through a new `native_timeout_scheduler` — a single lazily-spawned (`OnceLock`) timer thread draining a `BinaryHeap<ScheduledWake>` ordered earliest-deadline-first (correct min-heap inversion + `AtomicU64` FIFO tiebreak), fed by an `mpsc` channel, with `checked_add`/`saturating_duration_since` and an immediate-wake fallback if the thread can't start or the send fails. That's the right shape, not just a patch.
- **`lock_recover`.** Extracted to `transport/src/sync_utils.rs` and used by both `Notifier` and `DummyConnection`.
- **`encode_field_candidate(chunk) -> [u8;32]`.** `str_to_field` now calls a named, bounded helper (`payload_len = chunk.len().min(FIELD_CHUNK_SIZE)`, checked `get_mut`/`get`) instead of open-coded loops; the "marker < modulus prefix" invariant comment is retained.
- **IDB.** `next_visit_time_after(previous, now)` is a pure `pub(crate)` fn with a `// Post:` per-key-monotonic invariant (correctly scoped "for previous < i64::MAX"), and three `wasm_bindgen_test`s cover advance / stall+rewind / `i64::MAX` saturation.

### New finding (A5/A10): the timer scheduler is untested
`native_timeout_scheduler` is a non-trivial concurrent component (shared timer thread + deadline-ordered heap + mpsc + fallback) and it is compiled and live under the `dummy` build, yet has no test. Per the rustacean "stateful/concurrent logic needs a witnessed property" bar, add tests for at least: earliest-deadline-first wake ordering for out-of-order `schedule_wake` calls, and the fallback (immediate wake) when scheduling can't proceed. The ordering logic and fallback are exactly the parts a future edit could silently break.

### Minor (pre-existing, note only)
The whole `js_utils` `Global` enum is duplicated between `rings-core` and `rings-transport` (the transport copy even says "copied from … rings-core"); `set_timeout_0` is now added to both copies. Deduping would need a shared wasm-utils crate — out of scope here, but the cross-crate copy persists.

Reviewed from the diff (no branch fetch/strict-lint run). Acceptance is your call; from my side the only thing I'd still ask for is a test on the new timer scheduler.

RyanKung:
## Re-review — `54ba2074` ("Test native notifier timer scheduler")

Resolves my last open finding. Two deterministic tests, both witnessing the properties I flagged:

- `scheduled_wake_heap_orders_earliest_deadline_first` — drives a `BinaryHeap<ScheduledWake>` via a `#[cfg(test)] ScheduledWake::at(deadline, sequence)` constructor and asserts pop order is earliest-deadline-first. Tests the `Ord` contract directly, with explicit `Instant`s — no real timer or sleep, so it's not timing-flaky.
- `send_or_wake_falls_back_when_scheduler_is_missing_or_closed` — `schedule_wake` now delegates to a testable `send_or_wake(notifier, request, sender)`; the test covers `sender = None` and a dropped-receiver channel, asserting both fall back to immediate `wake()`.

Good choices: the tests isolate the pure decisions (heap ordering, send-vs-fallback) from the effectful thread/clock, and the `send_or_wake` extraction is a clean testability seam rather than test-only scaffolding bolted onto the loop. (The `run_timer_thread` deadline-wait loop itself remains untested, but that's thin glue over the `Ord`/fallback logic now covered, and testing it deterministically would need a time mock — reasonable to leave.)

I have no remaining findings. Across the rounds: panic removal + typed errors, the `poll!`→`spawn_interval` adapter, finger-table API cleanup, shared `parse_ice_servers_or_warn`/`lock_recover`, `Global::set_timeout_0`, the single-thread min-heap notifier scheduler, the IDB per-key-monotonic prune index, and an accurate title/body — all addressed with tests. Acceptance is your call; from my side there's nothing open. Reviewed from the diff (no branch fetch/strict-lint run here).

RyanKung:
LGTM

Issue #611, opened 2026-06-25
Complete DHT storage hand-off with a copy → ack → delete protocol


RyanKung:
## Summary

DHT storage hand-off on successor change is currently **copy-only**. `sync_entries_with_successor` ([`chord.rs`](crates/core/src/dht/chord.rs)) copies the affected entries to the new successor as `Vec<PlacedEntry>` and **never deletes the local copy** — the code comments explicitly call this an interim measure "until an explicit ack/delete protocol exists" (see `ChordStorageSync` doc in `types.rs` and the preservation note in `sync_entries_with_successor`).

That interim is safe (message loss can't destroy the only copy) but incomplete: the original owner accumulates entries it is no longer responsible for and never reclaims that space. This issue tracks completing the hand-off into a proper **copy → ack → delete** protocol, and it must be resolved within PR #609 before merge.

## Current state

- `sync_entries_with_successor(new_successor)` selects entries whose placement key is beyond `new_successor`, wraps each as `PlacedEntry { key, entry }`, and returns `RemoteAction::SyncEntriesWithSuccessor(Vec<PlacedEntry>)`. It does **not** remove them locally.
- Triggered from the stabilization flow in `message/handlers/stabilization.rs` (`HandleMsg<NotifyPredecessorReport>`).
- Receiver: `HandleMsg<SyncEntriesWithSuccessor>` (`message/handlers/storage.rs`) persists each `placed.entry.into_storage_entry()` at `placed.key`. It sends **no acknowledgement**.

## Required protocol: copy → ack → delete

1. Sender copies the entries to `new_successor` (already done).
2. Receiver persists them and replies with an **ack** listing the placement keys it durably stored.
3. Sender deletes a local entry **only after** it receives an ack for that specific placement key.

### Invariants (must hold)

- **No data loss:** the local copy is removed strictly after a confirmed durable store on the new owner. If the ack is lost or the receiver fails, the local copy is retained.
- **Idempotency / self-healing:** a lost ack must not strand data — the next stabilization round re-runs the hand-off, and re-storing an already-present entry is a no-op (the receiver already overwrites at `placed.key`).
- **Partial progress:** if only a subset of keys is acked, only that subset is deleted locally; the rest are retried.
- **Key-faithful:** delete keys by the `PlacedEntry.key` that was sent, not by `entry.did` (the two differ for redundant replicas — this is exactly why `PlacedEntry` was introduced).

## Suggested shape (non-binding)

[...]

PR #628, opened 2026-06-26
Make DHT storage and topology algebraic

### LINKED ISSUE #627: Make the distributed system thoroughly algebraic: replicated state as CRDT (join-semilattice), routing as a monotone fixpoint, pure transitions over the effect coproduct
## Thesis

Our distributed system is algebraic only at the edges. The identifier space is becoming a cyclic group (#626) and the message effects are a functor coproduct with an interpreter (#606, `message/effects.rs`: `PayloadRelayFunctor`/`MessageSendFunctor`/`ConnectionFunctor` → `CoreEffect` → `CoreEffectInterpreter`). But the **core** — the replicated DHT state and the routing/convergence state machine — is imperative. Correctness there is an *emergent property of hand-written code plus an external model checker*, not a *law of the data types*. This issue is to make the distributed system algebraic end-to-end.

## Diagnosis (with evidence)

**1. The replicated state is not a CRDT.** There is no `JoinSemilattice`/`merge`/lattice/`lub` anywhere in `crates/core/src` (the only `join`s are `FingerTable::join` and node `Chord::join`, unrelated). `Entry::operate` dispatches `Overwrite` / `Extend` / `Touch` / `JoinSubring`, and these are not join-semilattice operations:
- `Overwrite` is destructive last-writer-wins by replacement — not commutative, not idempotent, no version.
- `Extend` appends — order-dependent, and re-applying the same element duplicates (not idempotent).
- `Touch` reorders by recency — not a join.
- `JoinSubring` is set-union — the *only* one that is actually a semilattice, but it isn't typed as one.

Because the merge is not a semilattice, replica convergence (strong eventual consistency) is **not** a property of the type; it is propped up by imperative machinery — the copy→ack→delete hand-off (#611), anti-entropy repair (#612), and the versioned conditional-delete still owed (#614) — and validated by a *separate* `dht_stateright` model. The data type itself proves nothing.

**2. Routing/convergence is imperative; the algebra exists only in the test spec.** Chord transitions are concrete methods on `PeerRing` returning `PeerRingAction`/`RemoteAction` enums (`dht/chord.rs`), behind the `Chord`/`ChordStorage*`/`CorrectChord` traits (`dht/types.rs`). The one place convergence is stated algebraically — `Converged(n)` as the fixpoint of the `spec::*` operators in `tests/default/dht_convergence.rs` — lives in the **test** module, not the production type structure. So the protocol's "correct state is the least fixpoint of a monotone operator" is asserted in a TLA-style comment and checked by sampling, but it is not encoded.

**3. State transitions are not a pure, typed relation.** There is no `step: State × Event → (State, Actions)` core; transitions mutate `Arc<Mutex<…>>` (`predecessor`, `finger`) in place and emit action enums. The pure/effect split exists for *effects* (#606) but not for *state*.

Net: the distributed system has an effect algebra and (soon) an identifier group, but no algebra for **what it actually is** — a convergent replicated state under a monotone routing relation.

## Target: a thoroughly algebraic distributed system

Four layers, each an algebraic object with laws, composing:

### A. Replicated state = bounded join-semilattice (CRDT) — the central change
Define `trait JoinSemilattice { fn join(self, other: Self) -> Self; }` with the laws idempotent / commutative / associative (⇒ a partial order with least upper bounds). Model each `EntryKind` as a CRDT and make merge/sync/repair/operate **all** be `join`:
- Subring membership → G-Set (grow-only set); join = union.
- Data topic buffer → OR-Set or LWW-Element-Set; `Extend` = OR-Set add (dedup by unique tag), `Touch` = LWW by timestamp.
- `Overwrite` → LWW-Register (carries the version/timestamp #614 already needs).
- `RelayMessage` → set with TTL; delete = tombstone (2P-Set), so "no lost update" is a lattice fact.

Payoff: **Strong Eventual Consistency becomes a theorem.** Any replica that has observed the same set of joins — in any order, with any duplication — is in the same state, because state is the lub of a semilattice. This *subsumes* #611/#612/#614: ack/repair/anti-entropy become "deliver more joins" (idempotent, order-free), and conditional-delete becomes a monotone tombstone rather than a race-prone special case. Convergence stops being something the imperative code must carefully preserve and becomes something the type cannot violate.

### B. Routing/convergence = monotone operator with a least fixpoint
Model per-node protocol state as a value and stabilization/notify/fix_finger as a **pure monotone operator** `Step: State × Event → State` over the network snapshot; the correct ring is its least fixpoint. Promote the `dht_convergence` `spec::*` operators from the test module to the production transition definition, with monotonicity as the stated law. Convergence proofs then reduce to "the operator is monotone and the lattice has finite height."

### C. State transition = pure core, monadic
Introduce `fn step(&State, Event) -> (State, Vec<Action>)` as the pure Chord core (state monad shape), with `PeerRing` (`Arc<Mutex<…>>`) reduced to the interpreter/shell. `Action` is the existing effect coproduct (#606). The system becomes: pure state transition ∘ effect algebra ∘ interpreter — replayable from `(state, event)`.

### D. Identifier space = cyclic group
The `algebra.rs` group/field/module work (#626) is the foundation: routing offsets, finger targets (generator powers), and `bias` (group translation) are expressed as the group action over `Z / 2^160`.

### Verification (rustacean, distributed-systems section)
- Law tests for `JoinSemilattice` (idempotent/commutative/associative on representative samples) and for transition monotonicity — generic, witnessed once per carrier.
- SEC under schedule topology: extend `dht_stateright` to check that any interleaving/duplication/partition-then-merge of joins reaches one lub (the topology-aware test the skill asks for, now backed by the lattice law rather than ad-hoc state).
- `N`/`N+1` induction for ring convergence: the `dht_convergence` `Even(2..=8)` ladder is already the inductive step — name the reduction and the preserved fixpoint property.

## Phasing
1. `JoinSemilattice` trait + laws + generic law tests; reclassify the four `EntryKind`s as concrete CRDTs; route `operate`/sync/repair/ack through `join`. (Subsumes #611/#612/#614.)
2. Pure `step` core + monotone routing operator; lift `dht_convergence::spec` into production; `PeerRing` becomes the interpreter.
3. Unify with `algebra.rs` (#626) for the identifier group and with the effect coproduct (#606) for actions.

## Acceptance criteria
- [ ] A `JoinSemilattice` (CRDT) trait with idempotent/commutative/associative laws; each `EntryKind` is a named CRDT (G-Set / OR-Set / LWW / 2P-Set) implementing it; `operate`/sync/repair/ack are expressed as `join`.
- [ ] Strong Eventual Consistency is witnessed by a generic law test (any permutation/duplication of a join multiset → identical state) and by a `dht_stateright` topology model, not by imperative copy/ack/delete reasoning.
- [ ] Chord transitions exposed as a pure `step: State × Event → (State, Actions)`; the convergence operator is the production transition with a stated monotonicity law; `dht_convergence::spec` is no longer a separate test-only oracle.
- [ ] Identifier arithmetic and the effect layer are the `algebra.rs` group (#626) and the effect coproduct (#606) respectively — one algebra each, no parallel encodings.
- [ ] #611/#612/#614 are closed *as consequences* of the semilattice model (or explicitly explained where the lattice does not cover them).

## Related
#626 (algebra tower / identifier group), #606 (effect functor coproduct), #609 (Entry storage model), #611 (copy→ack→delete), #612 (durability/anti-entropy), #614 (versioned conditional delete), #620 (deterministic tests).

RyanKung:
Closes #627.

## Summary
- make replicated DHT entries explicit CRDT carriers with `JoinSemilattice` laws, including relay-message tombstones as a production `EntryOperation`
- expand the production pure Chord topology transition to cover join, remove, successor update, notify, stabilize, finger fix, and indexed finger-apply reports
- keep `PeerRing` as the interpreter shell over pure topology state and lower DHT actions through the existing effect coproduct, including `FindSuccessorForFix { did, index }`
- replace the storage convergence proof in `dht_stateright` with a finite partition/merge join-semilattice SEC model, while retaining copy/ack/delete only as cleanup safety
- update storage sync/repair documentation so #611/#612/#614 are consequences of join delivery plus bounded conditional cleanup, not separate convergence mechanisms

## Coverage for #627
- [x] `JoinSemilattice` laws and named DHT entry CRDTs; operate, sync persistence, repair, and ack persistence route through entry join
- [x] SEC witnessed by generic law tests and by a `dht_stateright` partition-then-merge topology model
- [x] Chord transitions exposed as a pure `step(State, Event) -> (State, Actions)` production core; convergence tests use `dht::topology` rather than a parallel oracle
- [x] Identifier arithmetic stays on `algebra.rs`/`Did`; DHT side effects lower through the existing effect coproduct without a parallel action encoding
- [x] #611/#612/#614 documented as lattice consequences or cleanup boundaries

## Validation
- `cargo check -p rings-core --tests`
- `cargo test -p rings-core dht::entry --lib`
- `cargo test -p rings-core dht_stateright --lib`
- `cargo test -p rings-core dht::topology --lib`

### DIFF
--- a/crates/core/src/algebra.rs
+++ b/crates/core/src/algebra.rs
@@ -28,6 +28,8 @@
 //! - [`Field`] is a commutative ring whose non-zero elements have inverses.
 //! - [`Module`] is a right scalar action of a commutative ring on an abelian
 //!   group.
+//! - [`JoinSemilattice`] is the CRDT merge structure used by replicated DHT
+//!   state.
 //!
 //! ## Rings DHT
 //!
@@ -65,6 +67,24 @@ use std::ops::Add;
 use std::ops::Mul;
 use std::ops::Neg;
 use std::ops::Sub;
+
+/// Join-semilattice for state-based CRDT merge.
+///
+/// `join` returns the least upper bound of two states from the same carrier.
+/// Implementors must make this operation inflationary with respect to the
+/// carrier's partial order; because Rust does not expose that order here, the
+/// obligation is witnessed through the idempotence, commutativity, and
+/// associativity laws below.
+///
+/// Law: `a.join(a) == a`.
+///
+/// Law: `a.join(b) == b.join(a)`.
+///
+/// Law: `a.join(b).join(c) == a.join(b.join(c))`.
+pub trait JoinSemilattice: Sized {
+    /// Return the least upper bound of `self` and `other`.
+    fn join(self, other: Self) -> Self;
+}
 
 /// Additive identity for an additive carrier.
 ///
@@ -192,6 +212,56 @@ where T: AbelianGroup + Clone + Eq + Debug {
             }
         }
     }
+}
+
+/// Assert join-semilattice laws for a representative finite sample.
+///
+/// This helper checks the state-based CRDT merge laws that imply strong
+/// eventual consistency for replicas that observe the same set of deltas in any
+/// order, with any duplication.
+#[cfg(test)]
+pub fn assert_join_semilattice_laws<T>(values: &[T])
+where T: JoinSemilattice + Clone + Eq + Debug {
+    for a in values {
+        assert_eq!(a.clone().join(a.clone()), *a);
+
+        for b in values {
+            assert_eq!(a.clone().join(b.clone()), b.clone().join(a.clone()));
+
+            for c in values {
+                let lhs = a.clone().join(b.clone()).join(c.clone());
+                let rhs = a.clone().join(b.clone().join(c.clone()));
+                assert_eq!(lhs, rhs);
+            }
+        }
+    }
+}
+
+/// Assert strong eventual consistency for one base state and a finite delta set.
+///
+/// The witness applies the same deltas in forward, reverse, and duplicated
+/// schedules. A lawful join-semilattice must materialize the same least upper
+/// bound for each schedule.
+#[cfg(test)]
+pub fn assert_strong_eventual_consistency<T>(base: T, deltas: &[T])
+where T: JoinSemilattice + Clone + Eq + Debug {
+    let forward = deltas
+        .iter()
+        .cloned()
+        .fold(base.clone(), JoinSemilattice::join);
+    let reverse = deltas
+        .iter()
+        .rev()
+        .cloned()
+        .fold(base.clone(), JoinSemilattice::join);
+    let duplicated = deltas
+        .iter()
+        .cloned()
+        .chain(deltas.iter().cloned())
+        .fold(base, JoinSemilattice::join);
+
+    assert_eq!(forward, reverse);
+    assert_eq!(forward, duplicated);
 }
 
 /// Assert the commutative-ring laws for a representative finite sample.
--- a/crates/core/src/algebra.rs
+++ b/crates/core/src/algebra.rs
@@ -72,9 +72,14 @@ use std::ops::Sub;
 ///
 /// `join` returns the least upper bound of two states from the same carrier.
 /// Implementors must make this operation inflationary with respect to the
-/// carrier's partial order; because Rust does not expose that order here, the
-/// obligation is witnessed through the idempotence, commutativity, and
-/// associativity laws below.
+/// carrier's induced partial order `a <= b iff a.join(b) == b`; because Rust
+/// does not expose that order here, the obligation is witnessed through the
+/// idempotence, commutativity, and associativity laws below.
+///
+/// The operation takes both arguments by value to model a pure state transition
+/// into a canonical least upper bound. Implementations that need an in-place
+/// merge can provide that adapter separately and keep this law-facing
+/// signature as the common algebraic surface.
 ///
 /// Law: `a.join(a) == a`.
 ///
@@ -152,6 +157,8 @@ pub trait CommutativeRing: AbelianGroup + Mul<Self, Output = Self> + One {}
 /// A field is a commutative ring whose non-zero values form a multiplicative
 /// group. `try_inverse` is fallible only because zero has no multiplicative
 /// inverse; returning `None` for a non-zero value violates the trait law.
+///
+/// Law: [`Zero::zero`] is distinct from [`One::one`].
 ///
 /// Law: non-zero values have a multiplicative inverse.
 pub trait Field: CommutativeRing {
@@ -300,12 +307,14 @@ where T: CommutativeRing + Clone + Eq + Debug {
 
 /// Assert the field inverse laws for a representative finite sample.
 ///
-/// This helper first checks the commutative-ring laws. It then checks that zero
-/// has no inverse and every sampled non-zero value has a two-sided inverse.
+/// This helper first checks the commutative-ring laws and the field
+/// non-degeneracy law `zero() != one()`. It then checks that zero has no inverse
+/// and every sampled non-zero value has a two-sided inverse.
 #[cfg(test)]
 pub fn assert_field_laws<T>(values: &[T])
 where T: Field + Clone + Eq + Debug {
     assert_commutative_ring_laws(values);
+    assert_ne!(T::zero(), T::one());
 
     for a in values {
         if a.is_zero() {
--- a/crates/core/src/dht/chord.rs
+++ b/crates/core/src/dht/chord.rs
@@ -353,13 +353,13 @@ impl PeerRing {
     /// value and `incoming` when a previous value exists; otherwise it is
     /// `incoming` normalized for storage.
     pub(crate) async fn join_storage_entry(&self, key: Did, incoming: Entry) -> Result<Entry> {
-        let incoming = incoming.into_storage_entry();
+        let incoming = incoming.try_into_storage_entry()?;
         let stored = if let Some(local) = self.storage.get(&key.to_string()).await? {
             local.join(incoming)?
         } else {
             incoming
         }
-        .into_storage_entry();
+        .try_into_storage_entry()?;
         self.storage.put(&key.to_string(), &stored).await?;
         Ok(stored)
     }
@@ -540,7 +540,7 @@ impl<const REDUNDANT: u16> ChordStorage<PeerRingAction, REDUNDANT> for PeerRing 
                         Some(this) => this,
                         None => op.clone().gen_default_entry()?,
                     };
-                    let entry = this.operate(op.clone())?;
+                    let entry = this.operate(op.clone(), self.did)?;
                     self.join_storage_entry(entry_key, entry).await?;
                     Ok(PeerRingAction::None)
                 }
--- a/crates/core/src/dht/chord.rs
+++ b/crates/core/src/dht/chord.rs
@@ -18,6 +18,7 @@ use super::entry::PlacementMiss;
 use super::finger::DEFAULT_FINGER_TABLE_SIZE;
 use super::successor::SuccessorSeq;
 use super::topology;
+use super::topology::FindSuccessorStep;
 use super::topology::TopologyAction;
 use super::topology::TopologyEvent;
 use super::topology::TopologyState;
@@ -118,7 +119,12 @@ pub enum RemoteAction {
     FindSuccessorForConnect(Did),
 
     /// Need `did_a` to find `did_b` then send back with `for finger table fixing` flag.
-    FindSuccessorForFix(Did),
+    FindSuccessorForFix {
+        /// DID whose successor should populate the finger slot.
+        did: Did,
+        /// Finger slot that should be updated by the report.
+        index: usize,
+    },
 
     /// Fetch successor_list from successor
     QueryForSuccessorList,
@@ -257,22 +263,12 @@ impl PeerRing {
     /// Also remove it from successor sequence.
     /// If successor_seq become empty, try setting the closest node to it.
     pub fn remove(&self, did: Did) -> Result<()> {
-        let mut finger = self.lock_finger()?;
-        let successor = self.successors();
-        let mut predecessor = self.lock_predecessor()?;
-        if let Some(pid) = *predecessor {
-            if pid == did {
-                *predecessor = None;
-            }
-        }
-        finger.remove(did);
-        successor.remove(did)?;
-        if successor.is_empty()? {
-            if let Some(x) = finger.first() {
-                successor.update(x)?;
-            }
-        }
-        Ok(())
+        let next = topology::step(
+            &self.topology_state()?,
+            TopologyEvent::Remove { peer: did },
+            self.successors().capacity(),
+        );
+        self.interpret_topology_state(&next.state)
     }
 
     /// Calculate bias of the Did on the ring.
@@ -281,20 +277,74 @@ impl PeerRing {
     }
 
     fn topology_state(&self) -> Result<TopologyState> {
+        let finger = self.lock_finger()?;
         Ok(TopologyState::new(
             self.did,
             self.successors().list()?,
             *self.lock_predecessor()?,
+            finger.list().clone(),
+            finger.fix_finger_index(),
         ))
     }
 
-    fn replace_successors(&self, next: &[Did]) -> Result<()> {
+    fn interpret_topology_state(&self, next: &TopologyState) -> Result<()> {
         let successors = self.successors();
         for did in successors.list()? {
             successors.remove(did)?;
         }
-        successors.extend(next)?;
+        successors.extend(&next.successors)?;
+        *self.lock_predecessor()? = next.predecessor;
+        self.lock_finger()?
+            .replace_state(&next.fingers, next.fix_finger_index);
         Ok(())
+    }
+
+    fn topology_action(&self, action: TopologyAction) -> PeerRingAction {
+        match action {
+            TopologyAction::FindSuccessorForConnect { next, did } => {
+                PeerRingAction::RemoteAction(next, RemoteAction::FindSuccessorForConnect(did))
+            }
+            TopologyAction::FindSuccessorForFix { next, did, index } => {
+                PeerRingAction::RemoteAction(next, RemoteAction::FindSuccessorForFix { did, index })
+            }
+            TopologyAction::QuerySuccessorList(did) => {
+                PeerRingAction::RemoteAction(did, RemoteAction::QueryForSuccessorList)
+            }
+            TopologyAction::Notify(did) => {
+                PeerRingAction::RemoteAction(did, RemoteAction::Notify(self.did))
+            }
+        }
+    }
+
+    fn topology_leaf_actions(&self, actions: Vec<TopologyAction>) -> PeerRingAction {
+        let mut actions = actions
+            .into_iter()
+            .map(|action| self.topology_action(action))
+            .collect::<Vec<_>>();
+        match actions.len() {
+            0 => PeerRingAction::None,
+            1 => actions.pop().unwrap_or(PeerRingAction::None),
+            _ => PeerRingAction::MultiActions(actions),
+        }
+    }
+
+    fn topology_multi_actions(&self, actions: Vec<TopologyAction>) -> PeerRingAction {
+        PeerRingAction::MultiActions(
+            actions
+                .into_iter()
+                .map(|action| self.topology_action(action))
+                .collect(),
+        )
+    }
+
+    /// Apply a reported finger successor through the pure topology transition.
+    pub(crate) fn apply_fixed_finger(&self, index: usize, successor: Did) -> Result<()> {
+        let next = topology::step(
+            &self.topology_state()?,
+            TopologyEvent::ApplyFinger { index, successor },
+            self.successors().capacity(),
+        );
+        self.interpret_topology_state(&next.state)
     }
 
     /// Join an incoming replicated entry delta into local storage.
@@ -324,47 +374,38 @@ impl Chord<PeerRingAction> for PeerRing {
     /// The caller will send it to the node identified by `did`, and let the node find
     /// the successor of current node and make current node connect to that successor.
     fn join(&self, did: Did) -> Result<PeerRingAction> {
-        if did == self.did {
-            return Ok(PeerRingAction::None);
-        }
-
-        let mut finger = self.lock_finger()?;
-
-        finger.join(did);
-        // Always try update
-        self.successors().update(did)?;
-        Ok(PeerRingAction::RemoteAction(
-            did,
-            RemoteAction::FindSuccessorForConnect(self.did),
-        ))
+        let next = topology::step(
+            &self.topology_state()?,
+            TopologyEvent::Join { peer: did },
+            self.successors().capacity(),
+        );
+        self.interpret_topology_state(&next.state)?;
+        Ok(self.topology_leaf_actions(next.actions))
     }
 
     /// Find the successor of a Did.
     /// May return a remote action for the successor is recorded in another node.
     fn find_successor(&self, did: Did) -> Result<PeerRingAction> {
-        let successor = self.successors();
-        let finger = self.lock_finger()?;
-
-        let succ = {
-            if successor.is_empty()? || self.bias(did) <= self.bias(successor.min()?) {
-                // If the did is closer to self than successor, return successor as the
-                // successor of that did.
-                Ok(PeerRingAction::Some(successor.min()?))
-            } else {
-                // Otherwise, find the closest preceding node and ask it to find the successor.
-                let closest_predecessor = finger.closest_predecessor(did);
-                Ok(PeerRingAction::RemoteAction(
-                    closest_predecessor,
-                    RemoteAction::FindSuccessor(did),
-                ))
+        let state = self.topology_state()?;
+        let succ = match topology::find_successor(&state, did) {
+            FindSuccessorStep::Local(successor) => {
+                // If the DID is closer to self than the successor head, return
+                // that head as the successor. With an empty successor list, the
+                // pure topology model returns `self.did`, matching
+                // `SuccessorSeq::min`.
+                Ok(PeerRingAction::Some(successor))
             }
+            FindSuccessorStep::Remote { next, did } => Ok(PeerRingAction::RemoteAction(
+                next,
+                RemoteAction::FindSuccessor(did),
+            )),
         };
 
         tracing::debug!(
             "find_successor: self: {}, did: {}, successor: {:?}, result: {:?}",
             self.did,
             did,
-            successor,
+            state.successors,
             succ
         );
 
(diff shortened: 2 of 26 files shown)


RyanKung:
## Strict review — algebraic DHT (#627)

This substantially achieves #627: the three `EntryKind`s are genuine state-based CRDTs (`GSet` union, `DataTopicBuffer` LWW-element-set with a reset `register`, `RelayMessageSet` 2P-set), `operate`/sync/repair persist through `join` (LUB), and routing is a pure `topology::step` consumed by production `PeerRing` at 8 sites (not a test-only parallel model). The bounded cap is a deterministic top-N over the total `(dot, value)` order, which is associative — good, the hard part is right. Strict findings below, ranked; none are "won't work," but several are latent SEC hazards and the laws hold only on an un-enforced sub-domain.

### 1. Two different cap strategies compose; SEC rests on an unstated "inputs ≤ N" invariant
`into_storage_entry` caps by **front-trim of the current `Vec` order** (`cap_recent_data` + skip from the front), while `join` → `materialize_elements` caps by **top-N of the `(dot, version)` order**. `entry_operate` and `join_storage_entry` wrap incoming deltas with `into_storage_entry` around the `join`. These two caps select *different* subsets once an input exceeds `ENTRY_DATA_MAX_LEN`. It is safe today only because stored/synced entries are already ≤ N, so `into_storage_entry` is a no-op. But the lattice's bounded merge is top-N-by-version; `into_storage_entry`'s independent front-cap is a second, divergent bound. If a >N entry ever reaches it, front-trim drops elements the lattice would keep and two replicas diverge — an SEC break. There should be **one** bounded-merge: route `into_storage_entry` through the same top-N-by-version materialize, not `cap_recent_data`.

### 2. `into_storage_entry` can desynchronize `data` and `dots`
In `into_storage_entry`, dots are capped in lockstep only `if self.crdt.dots.len() == self.data.len()`; otherwise `dots` is kept unchanged while `data` is trimmed. After that, `data.len() != dots.len()`, and `elements()` indexes `crdt.dots.get(index)` and **synthesizes a dot at `register`** for any missing index — silently replacing the real per-element LWW versions with the register version. That corrupts conflict resolution for the affected elements. Cap `data` and `dots` together (or, per #1, only ever cap through `materialize`, which keeps them aligned by construction).

### 3. The semilattice laws hold only on a normalized sub-domain that the types don't enforce
`DataTopicBuffer`/`RelayMessageSet` `join` is idempotent/associative only when every element has `version >= register` (and `adds ∩ removes = ∅`). The `pub(super) new(...)` constructors accept arbitrary maps; nothing guarantees normalization, so `join(a, a) != a` is constructible. The law tests pass because their samples are normalized — i.e. the laws are witnessed on the normalized domain only. Make the constructors normalizing (filter to `version >= register` at construction) so the invariant is a type guarantee rather than a discipline maintained by `materialize`.

### 4. `Did::default()` used as a sentinel "no actor"
`overwrite`/`extend`/`touch` call `ensure_*_stamp_after(Did::default(), …)`. The zero DID is a *valid identifier*, not a null; using it as "actor unknown, rely on an upstream stamp" is a magic-value sentinel. It is masked because `entry_operate` pre-stamps with `self.did`, but any `operate` on an unstamped entry collapses the LWW actor tiebreak to a shared zero. Make the actor explicit/required at the type level rather than defaulting to a real-looking DID.

### 5. The LWW overwrite `register` lacks a deterministic tiebreak
Element dots break ties by `(version, index, value)`, but the overwrite `register` is just an `EntryVersion = (epoch_ms, actor)`. Two overwrites with the same `epoch_ms` and same `actor` (e.g. two writes entering the same node in the same millisecond) compare **equal**, so `register.max` cannot deterministically choose between their two different payloads — a real (if narrow) non-convergence at the register. The register needs the same content/sequence tiebreak the element dots have. (Related: `EntryVersion.actor` is the first *processing node*, not the writer — consistent for SEC, but document that "actor" means the entry node, not authorship.)

### 6. SEC is witnessed by 3 fixed schedules on one carrier, not by the model
`assert_strong_eventual_consistency` checks forward/reverse/duplicated on `DataTopicBuffer` only. #627's stronger bar was topology-aware SEC — arbitrary interleavings and partition-then-merge — backed by extending `dht_stateright`. The data-type witness is good but does not cover arbitrary schedules or the distributed model; `dht_stateright` was not extended to the CRDT state. Worth closing that gap before claiming the SEC half of #627 in full.

### Minor
- `EntryDot.value: Did` is a content digest named `value` (it is the digest, not the value) — rename `digest`. `EntryDot::for_encoded` does `u32::try_from(index).unwrap_or(u32::MAX)`, a silent saturating dot-identity collision at ≥ 2^32 elements (unreachable, but it is a silent cast in an identity field).
- `topology.rs` doc states "stabilize and rectify are monotone refinements; their least fixpoint is the converged state" — good; but monotonicity itself is not witnessed by a test (only convergence-to-spec is). A `step` monotonicity property test would back the stated law.

The CRDT direction and the production-wired pure `topology::step` are the right model and a real achievement. The above is what I'd require before merge: collapse to a single bounded-merge (#1/#2), make normalization a type invariant (#3), remove the `Did::default()` actor sentinel (#4), give the overwrite register a total tiebreak (#5), and extend SEC to the model (#6). Reviewed from the diff; strict lint not run here.

RyanKung:
## Re-review — commits `f661945` (CRDT storage invariants) + `a1938e6` (model coverage)

All six findings from the previous strict pass are addressed, several with new witnessing tests. Verified the fixes are correct, not just present:

**#1 / #2 — dual cap + data/dots desync → unified.** `cap_recent_data` is deleted and `into_storage_entry` is replaced by `try_into_storage_entry`, which routes through the single lattice `materialize` (top-N by `(version)` total order). Data and dots can no longer drift apart because they are produced from one materialization. Witness: `storage_normalization_uses_lattice_top_n_order`.

**#3 — normalization not enforced → single fallible chokepoint.** `try_into_storage_entry` is now the one normalization path, invoked at every storage boundary (`chord` put, `storage_repair`, `handlers::storage` persist/sync, and the entry-equality check). Its fallibility is *algebraic*: the dot-index conversion that previously narrowed via `as u32` now returns `EntryDotIndexOutOfBounds { index }`, and tombstoning a non-tombstonable kind returns `EntryNotTombstonable`. Information loss is a typed error rather than a silent truncation. (Note: this is a boundary discipline, not a type-level invariant on the `pub(super)` constructors — acceptable since they are internal, but worth keeping in mind.)

**#4 — `Did::default()` actor sentinel → removed.** No longer present anywhere in the diff; the actor is threaded through `EntryVersion::issued_by(actor, operation)`.

**#5 — register tiebreak undefined for same `(epoch_ms, actor)` → total order.** `EntryVersion` gains `operation: Did`, "a deterministic content digest of the stamped operation," making `(epoch_ms, actor, operation)` a total order on versions. Two same-timestamp/same-actor overwrites with distinct payloads now converge deterministically. Witness: `overwrite_register_tiebreaker_converges_for_same_timestamp_actor`. The choice is digest-ordered rather than recency-meaningful at equal ms — correct for SEC, and that is all the property needs.

**#6 — SEC only asserted on local joins → topology-aware model.** `dht_stateright` now carries `StorageJoinState`: a partition-then-merge model where replicas exchange join deltas under arbitrary delivery/partition schedules, with `reachable_storage_join_states()` BFS and the property `∀node. replica[node] ≤ A ⊔ B ⊔ C` plus quiescent-LUB closure. The decomposition is sound: this model proves the delivery/partition/merge machinery converges to one LUB for *any* semilattice (`StorageJoinValue`), and the per-carrier law tests (`assert_join_semilattice_laws`, `assert_strong_eventual_consistency`) witness that Entry's `join` is a semilattice — together they give SEC for Entry over the gossip topology without exploding the state space on the full carrier.

Strict-lint sweep of the production deltas (`entry.rs`, `entry/crdt.rs`, `topology.rs`, `message/effects.rs`): no `unwrap/expect/panic/todo/unimplemented/indexing` (the `expect` strings are `#[test]` assertions).

One area I did not audit line-by-line in this pass: the large `chord.rs` rewrite in `a1938e6` (260 lines) and the `message/effects.rs` effect plumbing (+104). The `topology::step` wiring and the find-successor report effect are architecturally consistent with the established functor/step model, and `dht_find_successor_for_fix_sends_direct_indexed_report` covers the new report path, but the rewrite's breadth means CI (and the convergence/stateright suites) is carrying the correctness load there rather than this review.

RyanKung:
## Strict code-quality pass (entry.rs / entry/crdt.rs / dht_stateright.rs)

Issues only.

### Model coverage — the "complete model coverage" commit does not model the carrier

1. **The SEC stateright model checks a trivial G-Set, not the Entry CRDT.** `StorageJoinValue(u8)` with `join = Self(self.0 | other.0)` is a 3-element grow-only set. The partition/merge checker therefore proves only that *bitwise-OR* converges — which is trivially true. None of the parts that make `Entry::join` non-trivial are exercised under any interleaving: the LWW `register` reset floor, the bounded top-N cap (`ENTRY_DATA_MAX_LEN`), tombstone resurrection. Those are exactly where associativity/idempotence can break, and they are never model-checked. The `entry.rs` law tests only cover a few fixed samples, not schedules. So the convergence argument has a hole precisely at the carrier the PR is about.

2. **`preserves_lub_bound` is a vacuous invariant.** `dht_stateright.rs:690` asserts `value.is_subset_of(StorageJoinValue::ALL)` for every replica. Replicas start as A/B/C and only ever `|=`, so every reachable value is `⊆ 0b111` by construction — the predicate is `true` for all reachable states and asserts nothing. Only `is_quiescent_lub` carries content.

3. **The partition shape is hardcoded to one split.** `topology_permits` uses `from < 2 && to < 2` while `Partitioned` (`dht_stateright.rs:~650`), i.e. only the `{A,B | C}` partition is ever explored. Other partition shapes (and the magic `2`) are not modeled, so "arbitrary partition then merge" is overstated.

### Version semantics are incoherent

4. **`EntryVersion.epoch_ms` is documented as wall-clock but `after()` fabricates it.** The field doc says "Milliseconds since Unix epoch when the operation was issued" (`crdt.rs:38`), but `after()` (`crdt.rs:62`) returns `floor.epoch_ms.saturating_add(1)` when the real timestamp does not dominate the local floor. The field is simultaneously a wall clock and a Lamport counter. Either this is a hybrid logical clock — then name/document it as one and stop calling it "epoch ms when issued" — or the fabrication silently corrupts the timestamp's meaning.

5. **Two overlapping "current version" methods diverge silently.** `max_observed_version` (`entry.rs:375`) folds register + dots + **tombstones**; `observed_operation_version` (`entry.rs:384`) folds register + dots, **excludes tombstones**, and falls back through a `*version != EntryVersion::default()` sentinel to a freshly issued version. The tombstone divergence is undocumented (why does the overwrite floor see tombstones but the incoming-version computation not?), and the `EntryVersion::default()`-as-absent sentinel is the same default-sentinel anti-pattern that finding #4 of the previous round removed — reintroduced one layer down.

6. **`stamped()`'s documented purpose is overridden by `operate()`.** `stamped` (`entry.rs:230`) is documented to preserve existing CRDT witnesses "so forwarded operations keep the origin's dot/version instead of being reissued by every routing hop." But in the only place it is called (`chord.rs:524`, immediately before `operate` at `543`), `operate` → `overwrite/extend/touch` → `ensure_*_stamp_after` → `stamp_overwrite`/`stamp_delta` unconditionally re-derives `register` and **all** element dots from one floor-bumped version (`entry.rs:333-340, 397-405`) — there is no `has_witness` short-circuit on this path. Multi-element forwarded deltas thus have their per-element versions collapsed to a single version, and the preservation logic in `stamped` is dead on the operate path.

### Abstraction / duplication

7. **Four stamp methods, two redundant pairs.** `ensure_overwrite_stamp`/`ensure_delta_stamp` and `ensure_overwrite_stamp_after`/`ensure_delta_stamp_after` (`entry.rs:357-405`) are pairwise identical except for a single `stamp_overwrite` vs `stamp_delta` call, and `stamp_overwrite`/`stamp_delta` differ only in whether `register` is set. This is one parametrized operation expanded into four near-copies.

8. **`overwrite` is a merge wearing a replace's name.** `overwrite` (`entry.rs:608`) calls `self.join(...)`; replacement is an emergent side effect of the bumped `register` filtering old dots inside `materialize_elements`. The replace intent lives three indirections away (`after` → floor bump → `dot.version >= register` filter) with no comment at the `overwrite` site explaining why a join replaces.

9. **Bespoke hash framing instead of the crate's encoder.** `operation_digest` + the module-level free fn `append_digest_part` (`entry.rs:49, 342`) hand-roll a length-prefixed canonical encoding (`len:value`) purely to feed `gen_did`. This re-implements canonical serialization the crate already has (`Encoder`/serde); `append_digest_part` is a one-use file-scope helper.

### Smaller

10. **`crdt.rs` module doc mislabels invariants as "Laws."** The "Laws" block (`crdt.rs:8-15`) lists constructor postconditions ("`new` preserves only values whose dot is at or after the floor"); the actual semilattice laws (idempotent / commutative / associative) — the ones the carriers must satisfy — are not stated where the carriers are defined.

11. **`EntryDot.digest` is redundant with the map key.** In `DataTopicBuffer.values: BTreeMap<Encoded, EntryDot>`, `dot.digest == gen_did(value)` of the key, so it is stored and serialized but always recomputable from the payload it is keyed by.

12. **Inconsistent kind predicates.** `is_data_entry`/`is_subring_entry` exist and are used by `overwrite`/`extend`/`touch`, but `tombstone` (`entry.rs:653`) does a raw `self.kind != EntryKind::RelayMessage`; there is no `is_relay_entry`.

13. **Minor style.** `DataTopicBuffer::new`/`RelayMessageSet::new` (`crdt.rs:164, 192`) take a value then rebind `let mut x = x;` instead of a `mut` parameter; `materialize_elements` (`entry.rs:473`) clones every `Encoded` into the sort key (`(*dot, value.clone())`); `EntryElement` + `elements()` (`entry.rs:43, 417`) allocate a one-use Vec only to be folded straight into a `BTreeMap` by `topic_buffer`.

RyanKung:
## Re-review — commits `348e6c1`, `bd5fdbb`, `f4a7647`, `f085d67`

Re-read the new commits (entry.rs +213, crdt.rs +101, dht_stateright.rs +530), not just a closure-check of the prior round.

### Verified resolved (with witnesses)

- **Version semantics** (prev #4/#5/#6): `epoch_ms` → `logical_time_ms`, documented as a hybrid logical clock; `register` is now `Option<EntryVersion>` so "no reset floor" is distinct from "floor at t=0"; the second `observed_operation_version` method and its `EntryVersion::default()` sentinel are gone; `ensure_stamp_after` short-circuits on `has_write_witness()`, so forwarded dots are preserved rather than re-derived. The stamping pathways no longer contradict.
- **Stamp duplication** (prev #7): four methods collapsed into one `ensure_stamp_after(actor, floor, EntryStampKind)`.
- **Hash framing** (prev #9): `operation_digest` now serializes a typed `OperationDigest` via `bincode` instead of the hand-rolled `len:value` framing (determinism is fine — `Vec<u8>::encode` is pure base58 `encode_check`).
- **Doc/abstraction** (prev #8/#10/#11/#12): `overwrite` has a Preservation note; crdt module doc splits "Semilattice laws" from "Constructor postconditions"; `EntryDot.digest` (redundant with the map key) removed; `is_relay_entry` added and used by `tombstone`.
- **SEC model carrier** (prev #1): `StorageJoinValue` now wraps a real `Entry` and `join` runs the real `Entry::join` over three non-trivial carriers (bounded top-N, overwrite reset, relay tombstone). Crucially, `storage_entry_join_satisfies_semilattice_laws` proves on the **real entries** that `join(a,b).entry == canonical(a.bits|b.bits).entry` plus idempotence/commutativity/associativity — this is what makes the `bits` abstraction sound, and it's the right way to do it.

### Open findings

**1. The `Always` safety invariant `is_below_global_lub` is vacuous (prev #2, renamed not fixed).**
`is_subset_of` is `self.join(other) == other`, and `StorageJoinValue`'s `Eq` compares only `(carrier, bits)`. Every reachable value has `bits ⊆ global_lub.bits == 0b111`, so `is_below_global_lub` is `true` in every reachable state and can never fail. It exercises nothing about the real entry — a buggy `Entry::join` producing an out-of-lattice value would not be caught here (the value's `bits` are tracked independently of the entry). Convergence itself is genuinely established (law test + quiescent-LUB), so the model isn't worthless, but this particular invariant is decoration giving false confidence. Either drop it or make `is_subset_of` compare the actual entry.

**2. `assert_join_semilattice_laws(&values)` (dht_stateright.rs:1311) checks the proxy, not the carrier.**
`values` are `StorageJoinValue`s whose `Eq`/`Ord` ignore `.entry`, so this call asserts the bitmask G-Set laws — trivially true. The genuine carrier-law check is the manual `assert_eq!` on `.entry` directly below it. As written the call reads as if it validates the CRDT carrier while it validates the bitmask. Drop it, or run the helper over `.entry` values.

**3. The topology model's soundness is load-bearing on a separate test, but undocumented.**
With bits-only `Eq/Hash/Ord`, `storage_join_topology_model_converges_after_partition_merge` verifies only (a) bitmask convergence and (b) that real `Entry::join` never errors (via `panic!` in `storage_join_entry`). It lifts to *entry* convergence solely because `storage_entry_join_satisfies_semilattice_laws` separately proves `join(a,b).entry == canonical(a.bits|b.bits)`. That correspondence is the crux of the whole argument and should be stated where the topology model is documented; otherwise a reader over-trusts the topology stage in isolation.

**4. Minor — `RelayMessageSet::new` still rebinds `let mut adds = adds;`** (crdt.rs:201). The same nit was fixed in `DataTopicBuffer::new` (now `mut values:` parameter) but left here; take `mut adds: DataTopicBuffer`.

**5. Minor — `operation_digest` round-trips bincode bytes through base58 only to satisfy `gen_did(&str)`.** `bincode::serialize(..)` → `.encode()` (base58) → `gen_did` (sha1 of the string). The base58 step is pure overhead; hashing the bincode bytes directly would drop a allocation+encode per stamp. (Not a correctness issue — base58 is deterministic.)

RyanKung:
## Re-review — commit `e66cb84` "Tighten storage model review fixes"

Read the new commit fresh and re-verified the surrounding algorithm, not just the deltas.

### Round-3 findings — all resolved, with witnesses

1. **Vacuous `is_below_global_lub` (round-3 #1)** — `is_subset_of` now compares the **real entry**: `storage_join_entry(self.entry, other.entry) == other.entry` (with a carrier guard), instead of the bits-only `join == other`. The safety invariant now exercises the actual CRDT lattice — a non-monotone or lossy-cap `Entry::join` would make `join(replica, lub) != lub` and fail the invariant.
2. **`assert_join_semilattice_laws(&values)` on the bits proxy (round-3 #2)** — removed; the real carrier-law check is now solely the manual `assert_eq!` over `.entry`, and the unused import is gone.
3. **Undocumented load-bearing coupling (round-3 #3)** — added the "Quotient:" doc block stating that `StorageJoinValue` hashes/compares on `(carrier, bits)` to keep BFS finite, and that `storage_entry_join_satisfies_semilattice_laws` is the refinement witness (`Entry::join == canonical(bits_a ∪ bits_b)`). The dependency is now explicit.
4. **`RelayMessageSet::new` rebind (round-3 #4)** — now `mut adds: DataTopicBuffer` parameter.
5. **`operation_digest` base58 round-trip (round-3 #5)** — now `Did::try_from(HashStr::from_bytes(&bytes))`, hashing the bincode bytes directly. Bonus: the new `HashStr::from_bytes` is reused by `HashStr: From<T: Into<String>>`, which deletes the hand-rolled hex loop and its `write!(..).unwrap()` — one fewer panic path in production.

Verified the fixes don't break anything: `hex = "0.4.3"` is a dependency; `TryFrom<HashStr> for Did` exists; the digest stays in the sha1/H160 domain.

### Verification I ran on the unchanged core (no defects found)

- **Bounded top-N cap associativity** — `top_n(top_n(A∪B) ∪ C) == top_n(A∪B∪C)` holds because any element of `top_n(A∪B∪C)` that lies in `A∪B` is necessarily in `top_n(A∪B)` (fewer competitors in the subset), so the intermediate cap never drops a survivor. Dots are totally ordered (`(version{logical_time, actor, operation}, index)` with a value tiebreak in `materialize`), and within the model all dots are distinct, so the order has no ambiguous ties. The cap is genuinely exercised: each `DataBoundedTopN` singleton is already `ENTRY_DATA_MAX_LEN` long, so every pairwise join overflows.
- **Digest determinism** — `bincode::serialize(OperationDigest)` then sha1 is platform-stable; the `operation` tiebreak agrees across nodes.

### Minor (non-blocking, your call)

- **The topology model's *safety* invariant is now correct but largely redundant with the law test.** `is_below_global_lub` checks `join(replica, lub) == lub`, which is a consequence of idempotence/associativity/commutativity that `storage_entry_join_satisfies_semilattice_laws` already proves exhaustively over the same 8-value domain. The topology stage's *unique* contribution is the **liveness** result (`drive_to_quiescent_lub` reaches the LUB under every partition/merge schedule); the safety half is defense-in-depth. Fine to keep — just noting the safety property isn't carrying independent weight.
- **Heads-up, not a defect:** `operation_digest` now produces a different value than before this commit (sha1 of bincode vs sha1 of base58(bincode)). Irrelevant on an unreleased branch since `operation` is recomputed at stamp time and all nodes run one version — only flagging in case any fixture hard-codes a digest.

I did not find defects in this round. Not declaring it mergeable — that's your call.

RyanKung:
## Review — algebraic modeling / formal logic / trait design / Rust idiom / naming

Latest commit `41553f3` removed the redundant `is_below_global_lub`/`is_subset_of` safety check (carrier safety is the law test's job, the topology model's is liveness) and pinned the digest canonicalization with a test — both good. The review below is the requested deeper pass over the algebra (`algebra.rs`, `ecc/group.rs`, `dht/did.rs`, `dht/entry/crdt.rs`).

### 1. Algebraic modeling — correct and self-consistent

- `Did` = ℤ/2¹⁶⁰ as `AbelianGroup` and *deliberately not* `CommutativeRing`. Verified the impl matches the claim: `Did` implements `Zero`, `Neg`, `Add`, `Sub`, `AbelianGroup` only — no `One`/`Mul`/`CommutativeRing`. The doc's "Chord has no identifier multiplication" is honored in code.
- `Point<C>: AbelianGroup + Module<Scalar<C>> + CyclicModule`, `Scalar<C>: Field`. Modeling `Module<Scalar>` with the scalar as a **type parameter** (not an associated type) is the mathematically faithful choice — an abelian group can be a module over more than one ring, and an associated type would forbid that. Good.
- The bound stratification is coherent: `point * scalar` (`Mul<Scalar<C>>`) only needs `CurveGroup`, while scalar-field arithmetic needs `CurveScalarField`; a curve can expose the action without exposing full field arithmetic.

### 2. Formal logic — one real completeness gap

- **`Field` omits the non-degeneracy axiom `0 ≠ 1`.** The only stated law is "non-zero values have a multiplicative inverse," and `assert_field_laws` checks `zero → None` and `nonzero → two-sided inverse`. The **trivial ring** (where `0 == 1`, no non-zero elements) satisfies all of that *vacuously*, so it would type-check and pass the law witness as a `Field`. A complete field axiomatization requires `zero() != one()`. It's true for every real curve scalar field here, but it is neither stated nor witnessed. Fix: add the axiom to the `Field` doc and `assert_ne!(T::zero(), T::one())` to `assert_field_laws`.
- Correctly handled as *derivable* (not gaps), and worth keeping that way: right-distributivity from commutativity, and the absorbing law `a·0 = 0` from distributivity. The doc already notes the first; fine.
- `JoinSemilattice`: the three equational laws (idempotent/commutative/associative) are complete for a join-semilattice. Minor: the doc says the partial order is induced but never writes the inducing relation `a ⊑ b ⟺ a ∨ b = b`; stating it would make "inflationary"/"least upper bound" precise rather than informal.

### 3. Trait implementations — correct/consistent, two design notes

- Consistency verified end-to-end: every carrier implements exactly the traits its docs claim; `Zero: Add` and `One: Mul` supertraits make the identity laws well-typed (the earlier "law references an out-of-scope operation" issue is gone).
- **`JoinSemilattice::join(self, other) -> Self` is by-value**, which forces clone-heavy call sites throughout (`self.clone().join(other.clone())` in `entry.rs`, and per-state `.clone()` joins in the stateright model). It is algebraically clean, but given this project's stated preference for fewer brute clones and more borrowing, consider either an in-place `fn merge(&mut self, other: Self)` (LUB into `self`) with `join` as a thin by-value wrapper, or an explicit note that the clone cost is accepted for the functional signature. This is a trade-off, not a defect.
- **Name clash `algebra::Field` vs the curve base-field type `Field`** (libsecp256k1) forces `use crate::algebra::Field as AlgebraField` and the impl reads `impl ... AlgebraField for Scalar<C>`. The alias works but is a readability seam on the most important trait in the tower. Worth a deliberate rename decision (e.g. keep `algebra::Field` and alias the curve type instead, or qualify at use sites).
- `Module<S>` (type-param, general) coexisting with `CyclicModule::Scalar` (associated, canonical) is a slight conceptual tension, but the `CyclicModule` doc justifies it (the Rust supertrait-cycle prevents `CyclicModule: Module<Self::Scalar>`). Acceptable, and well-documented.

### 4. New Rust syntax (gen / yield) — not appropriate here

`gen` blocks are still unstable (`gen_blocks`); adopting them would force a nightly **build** toolchain, which conflicts with this crate's stable clippy/build gate (nightly is used only for `fmt`). So generators are premature for production code. Where laziness actually helps, stable `impl Iterator` is the right tool and is already used (`GSet::iter`). The Vec-returning spots don't benefit: `materialize_elements` needs `.len()` for the top-N cap, so it is inherently collect-then-cap and cannot be a pure lazy generator; `affine` is bounded-small (could return `impl Iterator`, but marginal). Recommendation: no gen/yield; keep `impl Iterator` where it pays.

### 5. Naming — strong overall, small nits

- Role/type-based naming is consistently good: `EntryVersion`, `EntryDot`, `DataTopicBuffer`, `RelayMessageSet`, `logical_time_ms`, `legacy_floor`, `has_write_witness`, `global_lub`, `bottom_like`, `materialize_elements`, `try_into_storage_entry`.
- Nits: `AlgebraField` alias (see §3); `StorageJoinValue.bits` is terse for "which initial-write contributions are present" (documented, so OK); `assert_module_action_laws` is precise (good rename from the old `assert_module_laws`).

The headline actionable item is §2: encode `0 ≠ 1` so the `Field` trait/test actually excludes the trivial ring. The rest are design/idiom notes for your call.

Issue #620, opened 2026-06-25
Fix sleep-dependent (non-deterministic) tests; finger-table/trim are secondary


RyanKung:
## Context

The test suite takes ~30 minutes, but slowness is a **symptom**. The root problem is that many tests synchronize with the system under test via fixed `sleep(N)` calls. **A `sleep`-then-assert test is incorrect, not merely slow** — and the fix is to make those tests correct. The size/trim items below are secondary optimizations.

The slow tests run over the **dummy (in-process) transport** (`cargo test -p rings-core --features dummy`), so there is no real network latency to wait on. The thing being waited for is in-process message propagation / stabilization state, which is directly observable — there is no reason to guess at it with a clock.

## Finding A (core): sleep-dependent tests are incorrect

`sleep(N); assert!(...)` does not assert the property it claims to. It asserts a **timing assumption** — "the async effect will have completed within N" — which is:

- **Racy / flaky:** under CI load, GC, a slower machine, or scheduler jitter, the effect may not have happened within `N`, so the test fails for reasons unrelated to the code. Conversely it can pass while observing a stale/intermediate state.
- **Non-causal:** the assertion is not tied to the event it depends on; nothing in the test establishes that the awaited transition actually occurred before the check. A green run is evidence of "N was big enough this time", not of correctness.
- **Slow by necessity:** because it must over-provision `N` to suppress the flakiness above, every run pays the worst-case wait even though the effect completes in milliseconds over the dummy transport.

So shortening the sleeps is the wrong fix — it trades slowness for flakiness while leaving the test just as unsound. The correct fix is to **synchronize on the actual condition**:

- poll-until-quiescent / poll-until-state, with a bounded timeout used only as a failure deadline (the suite already has `wait_for_msgs` / `assert_no_more_msg` — extend that pattern to the sleep sites);
- or drive the stabilization timer with `tokio::time::pause()` + `advance()` so timer-based logic is stepped deterministically;
- or the dummy controlled message queue, so message delivery order is explicit and the assertion runs exactly after the awaited delivery.

The timeout that remains must be a *deadline for declaring failure*, never the mechanism that paces a passing run.

### Where the unsound waits are

`crates/core/src/tests/default` + `crates/node/src/measure.rs` (lower bound; excludes the 30 s quiescence ceiling and connection/wasm tests):

| pattern | count |
|---|---|
| `sleep(from_millis(10000))` | 3 |
| `sleep(from_millis(5000))` | 1 |
[...]

PR #629, opened 2026-06-26
Add streaming E2E ElGamal encryption

### LINKED ISSUE #608: End-to-end ElGamal encryption with streaming support
## Summary

We have a working ElGamal primitive (`crates/core/src/ecc/elgamal`, generic over
cyclic groups after #603), but it is **not wired into the message path** — no
handler currently encrypts payloads with it. This issue proposes building a real
**end-to-end (E2E) encryption layer** for node-to-node messages on top of that
primitive, designed from the start for **streaming / chunked** payloads rather
than whole-message-in-memory encryption.

## Current state

- `ElGamal<G>` operates on *group elements*. Encryption is the pure kernel
  `(c1, c2) = (rg, m + rh)` over additive group notation; decryption is
  `m = c2 - x·c1`. Randomness is injectable (`encrypt_with_rng`) and a default
  shell (`encrypt`) wraps the thread-local sampler.
- The only message adapter is `impls/secp256k1.rs`, which maps a **`&str`** into
  secp256k1 points via a reversible x-coordinate lift (29 plaintext bytes per
  block → a `(CurveEle<33>, CurveEle<33>)` pair).
- Nothing in `crates/core/src/message` or `crates/node/src` calls it; there is no
  E2E encryption on the wire today.

## Problems to solve

1. **Plaintext type.** The adapter only accepts UTF-8 `&str`. E2E needs arbitrary
   bytes (`&[u8]`).
2. **No streaming.** The whole message is materialized as a `Vec<Point>` and
   encrypted/decrypted in one shot. This does not compose with the negotiated
   `max_message_size` chunking added in #601.
3. **Ciphertext expansion + cost.** ~29 plaintext bytes become two 33-byte points
   (~2.3× expansion), and **every block costs a fresh `lift_x` search plus two
   scalar multiplications**. Pure per-block ElGamal is far too expensive for
   large/streamed payloads.
4. **No integrity / authentication.** Textbook ElGamal is malleable and provides
   no integrity. E2E must bind ciphertext to an authenticated channel.

## Proposed direction: hybrid (KEM/DEM) streaming encryption

Rather than ElGamal-encrypting every payload block, use ElGamal as a **KEM** to
wrap a fresh symmetric key, then stream the payload with an **AEAD**:

1. Sender derives shared secret via ElGamal ephemeral (`s = rh`, send `c1 = rg`).
2. KDF over `s` → symmetric key for an AEAD (e.g. ChaCha20-Poly1305 / AES-GCM).
3. Payload is encrypted in **AEAD frames aligned to the chunk layer** (#601), each
   frame authenticated, with a per-frame nonce/counter and an end-of-stream
   marker to prevent truncation.
4. Receiver recovers `s = x·c1`, derives the key, decrypts frame-by-frame.

This gives constant per-message asymmetric cost (one ElGamal op), streaming-
friendly framing, and authenticated integrity — while keeping the existing pure
ElGamal kernel as the KEM building block.

## Open questions

- **Group choice:** keep secp256k1 (reuse node identity keys) or prefer
  Ristretto255 for cleaner/constant-time scalar ops? The primitive is already
  generic over both.
- **Where does the layer live?** A new `e2e`/`crypto` module in core vs. a message
  handler. How does it interact with the existing chunk reassembly state machine?
- **Key agreement & identity:** reuse the node's secp256k1 identity public key as
  the recipient key, or introduce dedicated encryption keys (forward secrecy via
  ephemeral-static vs. ephemeral-ephemeral)?
- **AEAD choice and dependency footprint** (native + wasm targets).
- **Should pure per-block ElGamal remain a supported mode** for small/structured
  payloads, or is hybrid the only path?

## Acceptance criteria (draft)

- [ ] Byte-oriented (`&[u8]`) E2E API, not `&str`-only.
- [ ] Streaming/chunked encrypt + decrypt that composes with `max_message_size`.
- [ ] Authenticated integrity (AEAD or equivalent), truncation-resistant.
- [ ] Works on native and wasm targets; clippy/fmt/rustdoc gates pass.
- [ ] Property tests (round-trip over random binary payloads and chunk sizes) and
      a documented threat model.

RyanKung:
## Summary

- Add direct secp256k1 ElGamal body encryption for E2E message payloads.
- Add signed E2E handshake request/response messages for public-key discovery and DID ownership verification.
- Add node Processor APIs to initiate E2E handshakes and send encrypted messages with verified recipient public keys.
- Add handler and Processor tests covering handshake response generation, invalid requester keys, and two-node handshake exchange.

## Design Notes

- ElGamal encrypts the message body directly; this does not use ElGamal as a KEM for wrapping a symmetric key.
- The encrypted message carries the sender public key, and handlers verify that the key resolves to the signed sender DID.
- The current API keeps key caching at the application layer.

Closes #608.

## Validation

```bash
cargo +nightly fmt --all -- --check
cargo test -p rings-core --lib --features dummy message::e2e
cargo test -p rings-core --lib --features dummy ecc::elgamal
cargo test -p rings-core --lib --features dummy message::handlers::e2e
cargo test -p rings-node --lib --features node test_processor_e2e_handshake_exchanges_verified_public_keys -- --nocapture
cargo test -p rings-node --lib --features node test_processor_handshake_msg -- --nocapture
cargo clippy -p rings-core --lib --features dummy --no-deps -- -D warnings
cargo clippy -p rings-node --lib --no-deps -- -D warnings
cargo clippy -p rings-core --lib --no-default-features --features wasm --target wasm32-unknown-unknown --no-deps -- -D warnings
cargo clippy -p rings-core --lib --tests --features dummy --no-deps -- -D warnings
cargo clippy -p rings-node --lib --tests --features node --no-deps -- -D warnings
RUSTDOCFLAGS='-D warnings' cargo doc -p rings-core --lib --features dummy --no-deps
```

### DIFF
--- a/crates/core/src/ecc/elgamal/impls/secp256k1.rs
+++ b/crates/core/src/ecc/elgamal/impls/secp256k1.rs
@@ -232,8 +232,8 @@ fn decode_field_bytes(mut bytes: [u8; 32]) -> Vec<u8> {
 /// The initial candidate uses its own parity. If it is not on the curve, the
 /// search decrements byte `0` from `254` to `1` and retries. This keeps bytes
 /// `1..` intact, so decoding can recover the original marker, chunk length, and
-/// plaintext bytes. The panic at `Some(0)` is an invariant failure: for the
-/// adapter's 254 alternate high-byte candidates, at least one should lift.
+/// plaintext bytes. If no high-byte candidate lifts, the adapter returns a
+/// typed error.
 fn lift_x(x: &Field) -> Result<Affine> {
     let mut ec = Affine::default();
     let mut x = *x;
--- a/crates/core/src/ecc/elgamal/impls/secp256k1.rs
+++ b/crates/core/src/ecc/elgamal/impls/secp256k1.rs
@@ -50,8 +50,17 @@ const FIELD_ENCODING_OVERHEAD: usize = 3;
 const FIELD_CHUNK_SIZE: usize = 32 - FIELD_ENCODING_OVERHEAD;
 const FIELD_CHUNK_SIZE_U8: u8 = 29;
 
+/// Plaintext bytes carried by one secp256k1 point in this adapter.
+pub const PLAINTEXT_BLOCK_SIZE: usize = FIELD_CHUNK_SIZE;
+
+/// One serialized ElGamal ciphertext block over secp256k1.
+pub type CiphertextBlock = (CurveEle<33>, CurveEle<33>);
+
 /// Plaintext input before it is mapped into secp256k1 group elements.
 pub struct Plaintext<'a>(&'a str);
+
+/// Binary plaintext input before it is mapped into secp256k1 group elements.
+pub struct PlaintextBytes<'a>(&'a [u8]);
 
 /// secp256k1 group elements that encode one plaintext message.
 pub struct MessagePoints(Vec<Point<Secp256k1>>);
@@ -59,6 +68,13 @@ pub struct MessagePoints(Vec<Point<Secp256k1>>);
 impl<'a> Plaintext<'a> {
     /// Plaintext string before group encoding.
     pub fn as_str(&self) -> &'a str {
+        self.0
+    }
+}
+
+impl<'a> PlaintextBytes<'a> {
+    /// Plaintext bytes before group encoding.
+    pub fn as_bytes(&self) -> &'a [u8] {
         self.0
     }
 }
@@ -72,6 +88,12 @@ impl MessagePoints {
 
 impl<'a> From<&'a str> for Plaintext<'a> {
     fn from(message: &'a str) -> Self {
+        Self(message)
+    }
+}
+
+impl<'a> From<&'a [u8]> for PlaintextBytes<'a> {
+    fn from(message: &'a [u8]) -> Self {
         Self(message)
     }
 }
@@ -95,7 +117,15 @@ impl<'a> TryFrom<Plaintext<'a>> for MessagePoints {
     type Error = Error;
 
     fn try_from(message: Plaintext<'a>) -> Result<Self> {
-        Ok(str_to_affine(message.as_str())?
+        MessagePoints::try_from(PlaintextBytes::from(message.as_str().as_bytes()))
+    }
+}
+
+impl<'a> TryFrom<PlaintextBytes<'a>> for MessagePoints {
+    type Error = Error;
+
+    fn try_from(message: PlaintextBytes<'a>) -> Result<Self> {
+        Ok(bytes_to_affine(message.as_bytes())?
             .into_iter()
             .map(Point::<Secp256k1>::from)
             .collect::<Vec<_>>()
@@ -107,11 +137,19 @@ impl TryFrom<MessagePoints> for String {
     type Error = Error;
 
     fn try_from(points: MessagePoints) -> Result<Self> {
+        String::from_utf8(Vec::<u8>::try_from(points)?).map_err(Error::Utf8Encoding)
+    }
+}
+
+impl TryFrom<MessagePoints> for Vec<u8> {
+    type Error = Error;
+
+    fn try_from(points: MessagePoints) -> Result<Self> {
         let affines = points
             .into_iter()
             .map(Affine::from)
             .collect::<Vec<Affine>>();
-        affine_to_str(&affines)
+        Ok(affine_to_bytes(&affines))
     }
 }
 
@@ -122,8 +160,13 @@ impl TryFrom<MessagePoints> for String {
 /// search bias for `lift_x`; the marker and length bytes make the mapping
 /// reversible and preserve leading or embedded NUL bytes.
 pub fn str_to_field(s: &str) -> Vec<Field> {
-    s.as_bytes()
-        .chunks(FIELD_CHUNK_SIZE)
+    bytes_to_field(s.as_bytes())
+}
+
+/// Convert arbitrary bytes into field elements using the adapter encoding.
+pub fn bytes_to_field(bytes: &[u8]) -> Vec<Field> {
+    bytes
+        .chunks(PLAINTEXT_BLOCK_SIZE)
         .map(|x| {
             let data = encode_field_candidate(x);
             let mut field = Field::default();
@@ -155,13 +198,17 @@ fn encode_field_candidate(chunk: &[u8]) -> [u8; 32] {
 
 /// Decode field elements produced by [`str_to_field`].
 pub fn field_to_str(f: &[Field]) -> Result<String> {
-    String::from_utf8(f.iter().fold(vec![], |mut acc, x| {
+    String::from_utf8(field_to_bytes(f)).map_err(Error::Utf8Encoding)
+}
+
+/// Decode field elements produced by [`bytes_to_field`].
+pub fn field_to_bytes(f: &[Field]) -> Vec<u8> {
+    f.iter().fold(vec![], |mut acc, x| {
         let mut field = *x;
         field.normalize();
         acc.extend(decode_field_bytes(field.b32()));
         acc
-    }))
-    .map_err(Error::Utf8Encoding)
+    })
 }
 
 fn decode_field_bytes(mut bytes: [u8; 32]) -> Vec<u8> {
@@ -223,7 +270,12 @@ fn lift_x(x: &Field) -> Result<Affine> {
 
 /// Convert a string into secp256k1 points using the adapter encoding.
 pub fn str_to_affine(s: &str) -> Result<Vec<Affine>> {
-    str_to_field(s)
+    bytes_to_affine(s.as_bytes())
+}
+
+/// Convert arbitrary bytes into secp256k1 points using the adapter encoding.
+pub fn bytes_to_affine(bytes: &[u8]) -> Result<Vec<Affine>> {
+    bytes_to_field(bytes)
         .into_iter()
         .map(|a| lift_x(&a))
         .collect::<Result<Vec<Affine>>>()
@@ -231,11 +283,16 @@ pub fn str_to_affine(s: &str) -> Result<Vec<Affine>> {
 
 /// Decode secp256k1 points produced by `str_to_affine`.
 pub fn affine_to_str(a: &[Affine]) -> Result<String> {
-    field_to_str(a.iter().map(|x| x.x).collect::<Vec<Field>>().as_slice())
+    String::from_utf8(affine_to_bytes(a)).map_err(Error::Utf8Encoding)
+}
+
+/// Decode secp256k1 points produced by [`bytes_to_affine`].
+pub fn affine_to_bytes(a: &[Affine]) -> Vec<u8> {
+    field_to_bytes(a.iter().map(|x| x.x).collect::<Vec<Field>>().as_slice())
 }
 
 /// Encrypt a string with the current secp256k1 compatibility adapter.
-pub fn encrypt(s: &str, k: PublicKey<33>) -> Result<Vec<(CurveEle<33>, CurveEle<33>)>> {
+pub fn encrypt(s: &str, k: PublicKey<33>) -> Result<Vec<CiphertextBlock>> {
     let public_key = ElGamalPublicKey::<Point<Secp256k1>>::from_element(k.try_into()?);
     let points = MessagePoints::try_from(Plaintext::from(s))?;
     ElGamal::<Point<Secp256k1>>::encrypt(points, &public_key)
@@ -249,9 +306,18 @@ pub fn encrypt_with_rng(
     s: &str,
     k: PublicKey<33>,
     rng: &mut impl RngCore,
-) -> Result<Vec<(CurveEle<33>, CurveEle<33>)>> {
+) -> Result<Vec<CiphertextBlock>> {
+    encrypt_bytes_with_rng(s.as_bytes(), k, rng)
+}
+
+/// Encrypt arbitrary bytes with caller-supplied randomness for ElGamal ephemerals.
+pub fn encrypt_bytes_with_rng(
+    bytes: &[u8],
+    k: PublicKey<33>,
+    rng: &mut impl RngCore,
+) -> Result<Vec<CiphertextBlock>> {
     let public_key = ElGamalPublicKey::<Point<Secp256k1>>::from_element(k.try_into()?);
-    let points = MessagePoints::try_from(Plaintext::from(s))?;
+    let points = MessagePoints::try_from(PlaintextBytes::from(bytes))?;
     ElGamal::<Point<Secp256k1>>::encrypt_with_rng(points, &public_key, rng)
         .into_iter()
         .map(|(c1, c2)| Ok((c1.try_into()?, c2.try_into()?)))
@@ -259,7 +325,12 @@ pub fn encrypt_with_rng(
 }
 
 /// Decrypt ciphertext produced by the current secp256k1 compatibility adapter.
-pub fn decrypt(m: &[(CurveEle<33>, CurveEle<33>)], k: SecretKey) -> Result<String> {
+pub fn decrypt(m: &[CiphertextBlock], k: SecretKey) -> Result<String> {
+    String::from_utf8(decrypt_bytes(m, k)?).map_err(Error::Utf8Encoding)
+}
+
+/// Decrypt arbitrary bytes produced by [`encrypt_bytes_with_rng`].
+pub fn decrypt_bytes(m: &[CiphertextBlock], k: SecretKey) -> Result<Vec<u8>> {
     let secret_key =
         ElGamalSecretKey::<Point<Secp256k1>>::from_scalar(GroupScalar::<Secp256k1>::from(k));
     let ciphertext = m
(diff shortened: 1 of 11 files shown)


RyanKung:
## Review — #629 streaming E2E ElGamal (closes #608)

Reviewed the crypto design and the protocol/layering, not just lints. Findings below, most-severe first.

### 1. It is not actually streaming, and the streaming state machine is dead weight

The production path is single-shot: `Processor::send_e2e_message` → `e2e::encrypt_bytes` (eager, whole plaintext) → **one** `swarm.send_message(Message::E2eEncryptedMessage)`. All frames are assembled into one `E2eEncryptedMessage` and signed as a single `MessagePayload`. Nothing is ever sent or received incrementally.

Given that, the entire `E2eEncryptor`/`E2eDecryptor`/`sequence`/`is_final`/`finish()` apparatus guards threats the single outer signature already eliminates: within one signed message, frame reordering, truncation, and "frame after final" cannot occur without breaking the signature. So `E2eFrameSequenceMismatch`, `E2eFrameAfterFinal`, and `E2eMissingFinalFrame` are effectively unreachable in production — they can only fire on a message the sender itself mis-built. This is a streaming protocol bolted onto a non-streaming transport. Either:
- make it real streaming (each `E2eFrame` an independently sent+signed unit, so the sequence/final machinery does real work across messages), or
- drop the per-frame state machine and just chunk into `frames: Vec` inside the one signed message.

As written it does not meet #608's streaming intent and carries a state machine that no real schedule exercises.

### 2. The receive/decrypt path is entirely unwired

`decrypt_message`, `E2eDecryptor`, and `secp256k1::decrypt_bytes` have **zero callers** outside the module's own tests (verified by grep). All three `HandleMsg` impls only verify and return:
- `HandleMsg<E2eEncryptedMessage>` calls `verify_sender` and discards the message — it never decrypts or surfaces plaintext;
- `HandleMsg<E2eHandshakeResponse>` calls `verify_responder` and **throws the verified responder key away** — nothing caches or emits it.

So a node can send but has no path to receive: the "application-layer caching" note hides that there is no decryption wiring at all, and the handler-level verification is redundant with what the app must redo after pulling the raw `Message` from the callback. At minimum the response handler should surface the verified key, and there should be an end-to-end decrypt test through the node, not only the in-module round-trip.

### 3. Primitive choice: textbook per-block ElGamal instead of a hybrid KEM+AEAD

`encrypt_bytes_with_rng` maps plaintext to curve points and ElGamal-encrypts each 29-byte block directly (design note confirms: "not a KEM"). Concrete costs of this choice for an E2E feature:
- **No authenticated encryption.** The cipher provides confidentiality only; integrity rests entirely on the outer signature. EC-ElGamal is homomorphic/malleable, so any context that ever handles a ciphertext outside the signed envelope has no integrity.
- **~2.3× ciphertext expansion**: each 29 plaintext bytes → two 33-byte points (66 bytes), per block, forever on the wire.
- **No forward secrecy** (see #4).

The standard construction (ECIES: ephemeral ECDH → KDF → AEAD such as ChaCha20-Poly1305) gives authenticated encryption, ~1× size, and per-message ephemeral keys. If direct ElGamal is a deliberate aesthetic choice, the trade-offs (size, malleability scope, no AEAD) should be documented as accepted, not just "does not use a KEM."

### 4. The identity/signing key is reused as the ElGamal encryption key

`send_e2e_message` uses `swarm.account_pubkey()` as both sender identity and encryption material, and decryption uses the account `SecretKey` — the same keypair that signs sessions/DIDs (`session.account_pubkey()`, one account key, no separate encryption key exists). This is cross-protocol key reuse (ECDSA signing + ElGamal decryption on one secp256k1 key), with two consequences: compromise of the identity key retroactively decrypts every past E2E message (no forward secrecy, no rotation), and signing+encryption sharing a key is a known footgun. Derive a dedicated encryption subkey rather than reusing the DID key.

### 5. Dead / half-abandoned code

- `ciphertext_block_count` (`e2e.rs:528`) is `pub` and has zero callers.
- `next_sequence`'s overflow branch (`e2e.rs:620`) is unreachable: 2⁶⁴ frames in one in-memory message is impossible.
- `MessagePoints` is now used by the `&str` path (`encrypt`) but bypassed on the byte path (`encrypt_bytes_with_rng` maps `bytes_to_affine().map(Point::from)` directly) and in `decrypt_bytes` (manual `Affine::from` collect). The abstraction is applied on one path and skipped on the parallel one — pick one.

### 6. Tests reintroduce the sleep-dependent anti-pattern (#620)

`wait_for_inbound_message` polls `inbound` 20×100ms and `wait_processors_connected` does `sleep(1s)`. This is exactly the sleep-timed, non-deterministic test shape #620 says is incorrect — flaky under load, slow, and not a real synchronization. Drive it off an event/notify, not a poll-with-sleep.

### 7. Minor

- `ensure_valid_public_key` (`e2e.rs:615`) uses a fully-qualified inline type `let _: crate::ecc::group::Point<crate::ecc::group::Secp256k1>` instead of a `use`.
- The three handler bodies repeat the `if let Some(effect) = e2e_local_or_forward_effects(..) { return self.run_effects([effect]).await; }` prologue verbatim.
- `drop(inbound)` (processor test) instead of a `{ }` scope.
- Renaming `shared_secret` → `mask` in `elgamal/mod.rs` is debatable: `s = rh = x·(rg)` genuinely is the Diffie–Hellman shared secret, so "shared secret" was the more precise term.

RyanKung:
## Re-review — force-pushed `593ca94`

Read the new revision fresh (e2e.rs rewritten, processor.rs +390, handlers/e2e.rs, secp256k1 adapter). My previous 7 findings are addressed; new findings below.

### Resolved (with witnesses)

1. **Now actually streaming (prev #1).** `send_e2e_message_with_frame_len` emits each `E2eStreamFrame` as its own signed `Message::E2eStreamFrame` in the chunk loop, so the `stream_id`/`sequence`/`is_final` machinery now does real work across independently-signed messages. The dead single-message `E2eEncryptedMessage` wrapper is gone.
2. **Receive/decrypt path is now wired and tested (prev #2).** New Processor APIs `verify_e2e_handshake_request/response`, `e2e_stream_decryptor`, `decrypt_e2e_stream_frame`, plus an end-to-end test (`..._streams_and_decrypts_with_receiver_identity_key`) that sends, collects frames, decrypts, and checks plaintext + wrong-key rejection.
3. **Primitive & key-reuse trade-offs documented (prev #3/#4).** The module doc now states the direct-ElGamal (non-AEAD) choice and its homomorphic-future rationale and malleability caveat, and that the DID/account key is intentionally the ElGamal key. `e2e_stream_decryptor` enforces `pubkey(identity)==self.did()` and documents why the session key can't decrypt. Accepted trade-offs, now explicit.
4. **Dead code removed (prev #5).** `ciphertext_block_count` gone. The `MessagePoints` asymmetry is fixed cleanly: a `PlaintextBytes` newtype + `TryFrom<PlaintextBytes> for MessagePoints` + `TryFrom<MessagePoints> for Vec<u8>`, with the str/bytes and encrypt/decrypt paths all routed through `MessagePoints` symmetrically.
5. **Tests are event-driven (prev #6).** `tokio::sync::Notify` + an `on_event` connection-state hook + `tokio::time::timeout` deadlines replace the fixed `sleep(1s)` / 100ms poll loop.
6. **Style (prev #7).** Forward prologue factored into `run_e2e_local_or_forward`; `ensure_valid_public_key` uses imported `Point<Secp256k1>`; `drop(inbound)` replaced by a scope.

### New findings

**1. The frame-chunking state machine is implemented twice; `encrypt_stream*` is now production-dead.**
`send_e2e_message_with_frame_len` (processor.rs) reimplements the exact chunk/`peekable`/`is_final`/empty-payload logic that `encrypt_stream_with_rng` (e2e.rs) already contains. Because the processor drives `E2eStreamEncryptor` directly to send each frame as it is produced, `encrypt_stream` / `encrypt_stream_with_rng` are now called only from tests — the one production site that should own the chunking doesn't use them. Consolidate: either have the processor call a shared frame-driver (e.g. `encrypt_stream*` returning frames, or a driver taking a per-frame async sink) or delete the now-redundant free functions. Right now there are two sources of truth for "how a byte stream becomes frames."

**2. The decryptor assumes in-order, gapless delivery and hard-errors otherwise — but frames are now separate messages that can be relayed.**
`E2eStreamDecryptor::decrypt_next` rejects any frame whose `sequence != next_sequence` with no reorder buffer. Sending each frame as an independent `swarm.send_message` is safe only if delivery is ordered and reliable. Rings routes messages through the ring (the handler itself has a `should_forward_from` relay path), and multi-hop relayed delivery has no ordering guarantee, so a legitimate multi-frame stream to a non-directly-connected DID can arrive reordered and fail with `E2eFrameSequenceMismatch`. The direct-peer test doesn't exercise this. Either document that E2E streaming requires an ordered reliable channel (and reject/guard relayed use), or buffer out-of-order frames by `sequence` until contiguous.

**3. Test synchronization via `Notify::notify_waiters()` has a lost-wakeup window.**
In `wait_processors_connected`, `wait_for_inbound_message`, and `wait_for_e2e_stream_frames`, the pattern is: check condition under lock, release, then `timeout(remaining, notify.notified())`. `notify_waiters()` stores no permit, so an event landing between the unlock and the `.notified()` poll is lost; if it was the final frame / the connecting event and nothing else follows, the wait runs to the 5s deadline and `.expect()` panics despite success. Low probability but real under CI contention — and exactly the test-determinism class #620 targets. Prefer `Notify::notify_one()` (stores one permit, enough since each wake re-checks the whole buffer) or a `tokio::sync::watch`.

### Minor

- `HandleMsg<E2eStreamFrame>` verifies the sender and `E2eStreamDecryptor::decrypt_next` verifies it again — double verification (defensible as early-reject, just noting).
- `prepare_processor_with_identity_key` largely duplicates `prepare_processor`; could share a builder.

RyanKung:
## Re-review — force-pushed `361d996`

Read the new revision fresh (e2e.rs +329, processor.rs +60). My round-2 findings are resolved; new findings concern the new reorder-buffer code.

### Resolved (with witnesses)

1. **Chunking is now a single lazy iterator (prev #1).** `E2ePlaintextFrames` owns the chunk/`is_final`/empty-payload logic once; `E2eStreamFrames` encrypts one frame per `next()`. The processor now drives `e2e::encrypt_stream_frames(...)` and sends each yielded frame, and `encrypt_stream*`/`encrypt_stream_with_rng` collapse to `.collect()` wrappers. The duplicated chunk loop in the processor is gone, and the iterator keeps sender memory bounded. (This is also the right answer to the `gen`/`yield` question from #628: a stable hand-written `Iterator`, no nightly feature.)
2. **Decryptor now tolerates reordering (prev #2).** `pending_frames: BTreeMap<u64, _>` buffers out-of-order frames and `decrypt_ready_frames` releases contiguous plaintext; `final_sequence` + `has_pending_frame_after` enforce that the final frame is the maximum. Tests `reordered_frames_are_buffered_until_contiguous`, `final_frame_can_arrive_before_gap_is_filled`, `frame_after_buffered_final_is_rejected` cover it, and the processor test now delivers frames reversed end-to-end. I traced the gap/final/duplicate-sequence cases — the contiguity and final-is-max logic is correct.
3. **Test lost-wakeup fixed (prev #3).** `notify_waiters()` → `notify_one()` in both notifies; the stored permit closes the check-then-wait race (one permit is enough since each wake re-scans the buffer under lock).

### New findings

**1. A duplicate/replayed frame is fatal — a one-packet stream-DoS over the relay path the reorder buffer is meant for.**
`validate_frame` rejects `frame.sequence < next_sequence` (already consumed) or an already-buffered sequence with `E2eFrameSequenceMismatch`, which propagates as a hard error and aborts the whole stream (`decrypt_stream` / `decrypt_e2e_stream_frame` return `Err`). But the *reason* this PR added a reorder buffer is that delivery isn't ordered — i.e. frames traverse untrusted relay hops. Any such hop can replay a previously-forwarded, validly-signed frame; one replay of an already-consumed sequence then kills the stream. A transport that reorders almost always also duplicates (at-least-once is the norm), so tolerating reordering while treating duplication as fatal is an inconsistent delivery model. A same-sequence frame is either byte-identical (benign) or signature-invalid (already rejected by `verify_sender`), so the safe behavior is **idempotent drop** of an already-seen/buffered sequence (return empty), not an error. If at-most-once is instead assumed, state that assumption and drop the reorder buffer.

**2. The reorder buffer is unbounded — one early gap buffers the rest of the stream in memory.**
`pending_frames` has no window or size cap, and it stores cloned *ciphertext* frames. A single missing/late low-sequence frame forces every later frame to accumulate until the gap fills; worst case the whole stream minus one frame sits in memory. That defeats the bounded-memory point of streaming and runs against the low-memory goal (#623). Add a bounded reorder window (reject or error frames more than `W` ahead of `next_sequence`, or cap `pending_frames.len()`), so memory is O(window) not O(stream).

### Minor

- `E2ePlaintextFrame` is a private module-internal struct, yet has `plaintext()`/`is_final()` accessor methods that only return the private fields and are called only within the module. Read the fields directly and drop the accessors.
- `encrypt_stream` (the eager `.collect()` wrapper, distinct from `encrypt_stream_with_rng`) now appears to have no production caller once the processor uses `encrypt_stream_frames`; confirm it's still needed or drop it.

The two new findings are about the delivery model: #1 (idempotent duplicates) and #2 (bounded buffer) should be decided together — both follow from pinning down exactly what delivery guarantees E2E streaming assumes. Not declaring it mergeable; that's your call.

RyanKung:
## Re-review — no new commits since `361d996`

Head is still `361d996` (identical to my last review), so there is no new code to evaluate. Instead of repeating, I did a fresh deep pass on the part #629 newly relies on but I hadn't audited: the secp256k1 byte↔point encoding (`encode_field_candidate` / `decode_field_bytes` / `lift_x`), which #629 now drives with **arbitrary binary** via `bytes_to_field`.

**That path is correct for binary** (no new defect):
- `[0xFF, 0x52(marker), len, payload(len ≤ 29), 0-pad]` round-trips; `decode_field_bytes` keys on `marker && len ≤ 29 && payload[len..] all-zero`, so embedded and trailing NUL bytes inside `len` are preserved (e.g. `[0x41,0x00]` decodes back exactly).
- `lift_x` mutates only byte 0 in its bias search, and `decode` discards byte 0, so the curve-lift search never corrupts the recovered bytes.
- `lift_x` is panic-free: the all-candidates-fail case returns `Err(Secp256k1PointLiftFailed)`, not a panic. Binary input doesn't change the ~2⁻²⁵⁵ miss probability, so #629's binary use doesn't add a failure mode.

**One pre-existing nit (in a file #629 modifies):** the `lift_x` doc says *"The panic at `Some(0)` is an invariant failure"*, but the code has no panic and no `Some(0)` arm — it returns `Err(Secp256k1PointLiftFailed)`. Stale comment; align it with the typed-error behavior.

**Still open from the previous round (unaddressed):**
1. Duplicate/replayed frame is fatal (`E2eFrameSequenceMismatch` aborts the stream) — a one-packet stream-DoS on the relay path the reorder buffer exists for; prefer idempotent drop of already-seen sequences.
2. The reorder buffer (`pending_frames`) is unbounded — one early gap buffers the rest of the stream in memory; add a bounded reorder window (vs goal #623).

Both are delivery-model decisions and should be settled together.

RyanKung:
## Re-review — `d2832ff` "Bound E2E stream reorder buffering"

(My previous note assumed no new commit — that was stale; this reviews the real `d2832ff`.) All open findings are resolved, and I traced the new decryptor state machine rather than just relying on the green tests.

### Resolved (with witnesses)

1. **Duplicate/replay is now idempotent, not fatal (round-3 #1).** `validate_frame` returns `FrameAcceptance::{New, Duplicate}`; `decrypt_next` returns `Ok(vec![])` on `Duplicate`. `is_consumed_duplicate` covers `sequence < next_sequence` and the post-final case (`seen_final && sequence <= final_sequence`); a same-sequence frame already in `pending_frames` is a duplicate iff byte-equal (`pending_frame == frame`). Tests `replayed_consumed_frame_is_idempotent`, `replayed_buffered_frame_is_idempotent`. The one-packet stream-DoS is gone.
2. **Reorder buffer is bounded (round-3 #2).** `reorder_window` (default `DEFAULT_E2E_REORDER_WINDOW_FRAMES = 64`, overridable via `with_reorder_window`) + `exceeds_reorder_window(seq) = seq.saturating_sub(next_sequence) > window` → `E2eFrameReorderWindowExceeded { next_sequence, actual, window }`. `pending_frames` is now capped at `window` entries. Test `future_frame_outside_reorder_window_is_rejected`.
3. **Minors:** the empty `E2ePlaintextFrame` accessors are gone (direct field access), the dead eager `encrypt_stream` wrapper is removed (no callers remain), and the stale `lift_x` "panic" doc now describes the typed-error behavior.

### Correctness I verified on the new state machine

- Check ordering in `validate_frame` is right: consumed-dup → buffered-dup(eq)/conflict(neq) → `seen_final` → window → `final_sequence` bounds → `is_final && has_pending_frame_after`. A replay after completion is dropped (consumed-dup clause) before the `seen_final` error can fire; a genuine new frame past `final` still errors `E2eFrameAfterFinal`.
- A same-sequence frame with **different** bytes returns `E2eFrameSequenceMismatch`. This is reachable only from the legitimate sender (every frame reaching the decryptor is already `MessagePayload`-signature-verified, so an on-path relay cannot forge a differing frame), so erroring is correct and not a new DoS surface.
- The window is a **sliding** frontier: `next_sequence` advances in `decrypt_ready_frames`, so a frame that is "too far" now can be accepted once the gap fills. Memory is O(window).
- The bounded-window abort does not add attack surface: a malicious relay carrying the stream can already DoS by dropping frames (→ `E2eMissingFinalFrame`); refusing to buffer beyond the window is the correct bounded-memory trade-off, since no scheme can both bound memory and tolerate unbounded adversarial reordering.

### Minor (optional)

- The window bounds frame **count** (64), so peak buffer memory is `64 × frame_size` and `frame_size` is caller-chosen via `max_plaintext_frame_len`. Fine for the default; if a caller sets a very large frame length, consider documenting that the reorder bound is in frames, not bytes.

No defects found in this round. Acceptance remains your call.

PR #630, opened 2026-06-27
Limit workflow token permissions

RyanKung:
## Summary
- Remove the obsolete `Notify New Issue to Developers` workflow instead of retaining a workflow with an empty token permission set.
- Add explicit least-privilege `GITHUB_TOKEN` permissions to the remaining workflows reported by Security and quality alerts.
- Keep release publishing permissions scoped to the release job only.
- Pin `crate-ci/typos` to the release tag `v1.47.2` instead of the moving `master` branch.
- Remove destination DID from the `send_message` info log to avoid cleartext sensitive identifier logging.

## Validation
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/qaci.yml .github/workflows/release-build.yml .github/workflows/auto-release.yml`
- `CARGO_HOME=/private/tmp/rings-security-quality-alerts-cargo-home CARGO_TARGET_DIR=/private/tmp/rings-security-quality-alerts-target cargo +nightly fmt --all -- --check`
- `CARGO_HOME=/private/tmp/rings-security-quality-alerts-cargo-home CARGO_TARGET_DIR=/private/tmp/rings-security-quality-alerts-target cargo test -p rings-node --lib --features node test_processor_handshake_msg`


RyanKung:
## Review — #630 "Limit workflow token permissions"

Small, focused security hardening. Verified correctness and completeness; no defects.

### Correct & complete

- **Workflow coverage is total.** The repo has exactly four workflows (`auto-release`, `issue`, `qaci`, `release-build`) and all four now declare `permissions`. No workflow is left on the default broad token.
- **`issue.yml` `permissions: {}` is the right call, not a break.** Despite the `label_issue` job name, the job never touches the GitHub API — it only calls `appleboy/telegram-action` with Telegram secrets and reads `github.actor` / `github.event.issue.html_url` from the event context. So it needs zero `GITHUB_TOKEN` scope; `{}` is exact least privilege. (The `label_issue` name is a pre-existing misnomer, out of scope here.)
- **`release` job scoping is right:** `contents: write` (create the GitHub release) + `actions: read` (artifact access), with everything else read-only. The publish privilege is confined to that one job.
- **Reusable-workflow token flow is consistent.** `release-build.yml` is `workflow_call`-only; its effective token is capped by the caller. Callers pass `contents: read` (qaci top-level; auto-release `build` job), matching release-build's own `contents: read`.
- **Log change is complete.** `destination` is removed from the `send_message` info log, and a grep of `crates/{node,core}/src` finds no other log macro that emits `destination` — so the who-talks-to-whom metadata isn't leaked elsewhere.

### Minor (optional, non-blocking)

- **Redundant job-level permission:** in `auto-release.yml`, top-level `permissions: contents: read` already applies to every job, so the explicit `build: permissions: contents: read` is redundant (harmless; `qaci.yml`'s reusable-calling job correctly relies on the top-level grant — the two files are just slightly inconsistent in style).
- **Debuggability vs privacy:** the log now drops `destination` entirely. If routing diagnosis ever needs it, consider keeping it at `trace!`/`debug!` (or a truncated/hashed form) rather than removing it from `info!` — `msg.len()` is still logged, so payload size remains visible regardless.

No correctness issues found. Acceptance is your call.

RyanKung:
## Supplementary (same security theme): unpinned third-party actions

Provenance note for the `appleboy/telegram-action` in `issue.yml`: it was added intentionally in #546 (2024-02, "Notify new issue to developers") to ping the dev Telegram group on new issues. The `label_issue` job name is a pre-existing misnomer (it notifies, it doesn't label). So it's legitimate — but it's the highest-risk action in the repo from a supply-chain view, and `permissions: {}` doesn't cover it:

- It consumes secrets (`RINGS_NETWORK_MONITOR_BOT_TOKEN`, `RINGS_NETWORK_TELEGRAM_GROUP_ID`).
- It is triggered by `issues: opened`, i.e. by **any external user** opening an issue.
- It is pinned to a mutable tag `@v0.1.1`. If that tag is moved / the upstream is compromised, an attacker opens an issue and the poisoned action runs with the bot token. `permissions: {}` removes `GITHUB_TOKEN` scope but not these secrets.

This is repo-wide, not unique to telegram — **every** `uses:` across the four workflows is tag/branch-pinned, none by commit SHA. The loosest is `crate-ci/typos@master` (`qaci.yml`) — pinned to a moving branch, which runs in PR CI. First-party `actions/*` are lower risk but still best-pinned by SHA.

Out of scope for this PR's size, but on the same hardening axis, a sensible follow-up priority order:
1. `appleboy/telegram-action` → pin to a 40-char SHA (secrets + external trigger).
2. `crate-ci/typos@master` → pin to a released tag, then a SHA.
3. Remaining third-party actions → SHA-pin (e.g. via Dependabot `github-actions` updates, which keeps SHA pins current).

Happy to keep this for a separate PR; flagging it here because #630 is the workflow-security PR.

RyanKung:
## Re-review — `a26c5d3` "Remove issue notification workflow"

Verified the new commit. It resolves the top item from my supplementary note by deleting `issue.yml` outright rather than SHA-pinning it.

- `issue.yml` is fully removed (file gone, not emptied), and there are **no dangling references** to it or to `RINGS_NETWORK_MONITOR_BOT_TOKEN` / `RINGS_NETWORK_TELEGRAM_GROUP_ID` anywhere in the tree.
- This eliminates the highest-risk surface I flagged: a secrets-handling, externally-triggered (`issues: opened`), tag-pinned third-party action. Removal is a stronger fix than pinning.
- The other three workflows' `permissions` blocks and the `send_message` log change are unchanged from the prior review (delta is `issue.yml` only). Final PR: least-privilege `permissions` on `auto-release`/`qaci`/`release-build`, the notification workflow removed, and the destination DID dropped from the log. All correct and complete.

### One hygiene follow-up (credential, not code)

Deleting the workflow does **not** invalidate the Telegram bot token — `RINGS_NETWORK_MONITOR_BOT_TOKEN` is still a live credential in repo secrets, and it was exposed to a mutable-tag action for ~2.5 years. Recommend **revoking/rotating that bot token** (via BotFather) and deleting the now-orphaned `RINGS_NETWORK_*` repo secrets to fully close the loop.

### Still open (optional, out of scope — from the earlier note)

- `crate-ci/typos@master` is pinned to a moving branch; other third-party actions are tag-pinned. SHA-pinning these (ideally via Dependabot `github-actions`) remains a sensible separate hardening PR.

No new defects. Acceptance is your call.

RyanKung:
## Re-review — `c1f5e7e` "Pin typos action to release tag"

Resolves the last in-scope item from my earlier note.

- `crate-ci/typos@master` → `crate-ci/typos@v1.47.2`. This removes the moving-branch pin (where any push to the action's `master` would run in PR CI) and also makes the typo check deterministic across runs.
- Confirmed there are **no remaining moving refs** (`@master`/`@main`/`@latest`) in any of the three workflows — typos was the last one. All actions are now uniformly tag-pinned.

Final PR state is coherent on the workflow-security axis: least-privilege `permissions` on all workflows, the secrets-handling externally-triggered notification workflow removed, and no branch-pinned actions.

### Still open (optional, separate PR / non-code)

- Tag pins are still technically mutable; SHA-pinning is the gold standard. But the repo is now internally consistent (all tag-pinned), so this is a uniform follow-up rather than fixing an outlier — well-suited to Dependabot `github-actions`.
- Credential hygiene: revoke/rotate the now-unused Telegram bot token (`RINGS_NETWORK_MONITOR_BOT_TOKEN`) and delete the orphaned `RINGS_NETWORK_*` secrets — deleting the workflow doesn't invalidate the token.

No new defects.

PR #633, opened 2026-06-27
Add manual cargo publish helper

### LINKED ISSUE #631: Make crates.io publishing crate-local and script-driven
## Background

We publish Rust crates manually. This should stay outside GitHub Actions: crates.io upload is an irreversible release-side effect and should not be folded into the shared release/QACI workflow.

The current workspace layout also makes `cargo publish` awkward. Internal crates are mostly inherited through root `[workspace.dependencies]` as path-only dependencies. When Cargo packages a crate, it removes `path`; if no `version` remains, packaging fails. For example, packaging a crate that depends on `rings-derive` through a path-only workspace dependency fails because crates.io needs an explicit versioned dependency.

## Direction

Move crates.io publishing concerns down to the publishable crates instead of treating publishing as a root-workspace operation.

Concretely:

- publishable crate manifests should carry crates.io-ready dependency metadata;
- internal dependencies used by publishable crates should include both local `path` and crates.io `version` at the crate level;
- manual publishing should be assisted by a small shell script, not by a workflow;
- the script should sequence package/dry-run/publish by dependency topology and leave upload under explicit human control.

## Proposed script

Add a small script, for example:

```sh
scripts/cargo-publish-crates.sh check
scripts/cargo-publish-crates.sh dry-run
scripts/cargo-publish-crates.sh publish
```

Expected behavior:

- `check`: verify every publishable crate has a concrete package version and versioned internal dependencies;
- `dry-run`: run `cargo publish --dry-run` or equivalent package checks in publish order;
- `publish`: run the same order with real `cargo publish`, stopping after each crate until crates.io has indexed it;
- no token handling beyond normal local `cargo login` / `CARGO_REGISTRY_TOKEN`;
- no GitHub Actions integration.

## Initial publish order

The script should encode the crate dependency order rather than relying on a workspace-wide publish operation:

```text
rings-derive
rings-transport
rings-snark
rings-core
rings-rpc
rings-node
```

This order may need adjustment if crate dependencies change; the script should keep the order explicit and easy to audit.

## Acceptance criteria

- No `cargo publish` step is added to GitHub Actions.
- Publishable crates can be packaged without missing-version errors for internal dependencies.
- Internal publish metadata is owned by the relevant crate manifests, not only by root workspace dependency inheritance.
- A small shell script exists for manual `check`, `dry-run`, and `publish` flows.
- The script fails fast and prints the exact crate being processed.
- The script documents publish order in code, not in a separate publishing document.

RyanKung:
## Summary

Closes #631.

This PR keeps crates.io publishing manual and outside GitHub Actions. It adds a small local bash helper for manual publish sequencing and moves publish-relevant crate metadata down into the publishable crate manifests.

## Changes

- Add `scripts/cargo-publish-crates.sh` with explicit `check`, `dry-run`, and `publish` modes.
- Encode publish order in the script: `rings-derive`, `rings-transport`, `rings-snark`, `rings-core`, `rings-rpc`, `rings-node`.
- Make publishable crates use concrete crate-local package versions instead of `version.workspace = true`.
- Give publishable crate manifests crate-local internal dependencies with both local `path` and crates.io `version`.
- Keep `cargo publish` entirely manual; no workflow or secret handling is added.

## Validation

- `scripts/cargo-publish-crates.sh check`
- `scripts/cargo-publish-crates.sh dry-run`
- `cargo publish --dry-run -p rings-snark --allow-dirty`
- `bash -n scripts/cargo-publish-crates.sh`
- `cargo check -p rings-core --features dummy --no-default-features`
- `cargo check -p rings-node --features node --no-default-features`
- `git diff --check`


RyanKung:
## Review — #633 manual cargo publish helper

Verified the publish mechanics against the workspace. Findings below.

### Verified correct
- **Publish order is a valid topological sort.** `rings-derive`, `rings-transport`, `rings-snark` have no internal `rings-*` deps; `rings-core` → {transport, derive}; `rings-rpc` → core; `rings-node` → all — each dependency precedes its dependents.
- **`ice_server.rs` cfg gate is exact.** `parse_ice_servers_or_warn` is called only from the `dummy` / `native_webrtc` / `web_sys_webrtc` connection modules, which `connections/mod.rs` gates with `#[cfg(feature = "dummy"|"native-webrtc"|"web-sys-webrtc")]`. The new `#[cfg(any(feature = "dummy", "native-webrtc", "web-sys-webrtc"))]` matches that union precisely — no over- or under-gating. (Keep it in sync if a backend module's gate ever changes.)
- Script hygiene is sound: `set -euo pipefail`, `run_check` before publish, clean-worktree guard, and skip-already-indexed for resumable re-runs.

### Maintainability — the main concern

`0.10.0` is now duplicated across **six `[package] version` literals plus ~ten internal-dependency `version = "0.10.0"` entries**. The actual publish requirement is only that *internal dependencies carry a `version`* (path-only deps can't publish). The root `[workspace.dependencies]` entries currently have `path` but no `version`, which is why `{ workspace = true }` wouldn't publish — but the minimal fix is to **add `version` to those six root lines and keep members on `{ workspace = true }`** (optionally `{ workspace = true, features = [...] }` for the `core_crate` case). Cargo resolves workspace inheritance — including path+version deps — into the published manifest, so that publishes correctly while keeping one source of truth.

Likewise `[package] version = "0.10.0"` is not required for publishing (`version.workspace = true` is resolved at publish time); it's only needed if independent per-crate versioning is an explicit goal. As written, a version bump now edits ~16 sites instead of one, and the next bump is an easy place to introduce a skew.

### Script gap that this duplication makes important

`check` asserts each internal dep has both `path` and `version`, but **not that the `version` equals the depended crate's `[package] version`**. With versions hand-maintained in many places now, a stale entry (e.g. `rings-rpc` pinning `rings-core = { version = "0.9.0" }`) passes `check` yet would publish `rings-rpc` against an *old* crates.io `rings-core`. Add a consistency assertion (dep `version` == that crate's package version) — it's exactly the guard the new literal-duplication needs.

### Minor
- `crate_version_indexed` keys readiness on `cargo search`, whose API cache can lag the sparse index that `cargo publish` actually reads — `wait_for_index` may pass prematurely or time out. Acceptable for a manual helper; querying `https://index.crates.io/...` is more reliable.
- `dependency_line` / `check_manifest` grep the whole manifest section-agnostically. Harmless today (no publishable crate has an internal `rings-*` dev-dependency), but a future path-only internal dev-dep would false-fail the check; scoping to `[dependencies]` would harden it.
- nit: `crate_version_indexed`'s grep leaves the version's `.` unescaped (regex any-char); negligible false-match risk.

No correctness defects in the publish flow itself; the headline is the version-literal duplication and the missing version-consistency check. Acceptance is your call.

PR #638, opened 2026-06-28
Add integration coverage for examples

### LINKED ISSUE #636: Add integration test coverage for all examples and move FFI header ownership out of examples
## Problem

The examples are not covered consistently. A quick inventory shows only two test files under `examples/` today:

- `examples/browser/tests/echo.rs`
- `examples/relay/tests/relay.rs`

Those are useful but they only cover narrow helper paths. The rest of the example modules do not have dedicated tests, and the current FFI path is only smoke-run by CI:

```bash
pip install wheel && pip install web3 cffi && python examples/ffi/rings.py
```

That smoke run is not enough integration coverage. It does not make `rings.py` import-safe, does not assert behavior, and does not catch drift between the Rust FFI declarations and the hand-maintained `examples/ffi/rings.h`.

There is also an ownership problem: `rings.h` should not be maintained by an example. The C ABI header should be generated or owned by the `rings-node` FFI build surface, while the example consumes it.

## Desired outcome

Every example should have meaningful tests, and the examples CI should run those tests as integration coverage rather than only building or smoke-running scripts. The FFI header should move out of example ownership.

## Required coverage

Add or expand test coverage for each example module:

- `examples/ffi`
  - Move canonical ownership of `rings.h` out of `examples/ffi/`. The example should consume a generated or crate-owned header instead of maintaining the ABI declaration by hand.
  - Make header generation/checking part of the FFI build flow, preferably from the `rings-node` FFI surface via cbindgen.
  - Refactor `rings.py` so it can be imported by tests without immediately reading files, opening the dynamic library, or running the example main path.
  - Add Python integration tests that run against the built `rings-node` cdylib and cover header loading, dynamic library path resolution, signer callback behavior, provider creation, `nodeInfo`, and `createOffer`.
- `examples/native`
  - Add integration tests or a deterministic harness for the native example workflow.
  - Cover seed URL argument handling, provider startup, message send path, and timeout behavior without relying on uncontrolled sleeps where possible.
- `examples/browser`
  - Keep the existing Rust `Echo` helper test, but add coverage for the actual browser example API surface.
  - Cover provider startup, seed connection request construction, DID connection request construction, backend message send, and protocol callback behavior.
- `examples/relay`
  - Keep the existing relay test, but make the module coverage explicit.
  - Cover the TCP and UDP relay entry points and failure cases expected from the example.
- `examples/dweb`
  - Add tests for the example library and/or wasm/webdriver flow so the published example behavior is checked in CI.
- `examples/proof-demo`
  - Add tests for the proof-demo library and/or wasm/webdriver flow so the example does not regress silently.
- `examples/snark`
  - Add tests for the example circuits/proof flow that can run in CI with deterministic fixtures.

## CI requirements

- The examples CI job should run the new test suites, not just `cargo build` or `python examples/ffi/rings.py`.
- The FFI CI path should build `rings-node --features ffi` and then run the Python integration tests.
- CI should fail if the generated/crate-owned FFI header drifts from the Rust FFI declarations.
- Tests should avoid depending on external network services unless the dependency is explicitly mocked, pinned, or skipped with a clear reason.

## Acceptance criteria

- Every directory under `examples/` has a test target or a documented integration test entry in CI.
- `examples/ffi/rings.h` is no longer the hand-maintained source of truth for the FFI ABI.
- `rings.py` can be imported without side effects that require the dynamic library or local header path.
- The FFI example remains runnable for users after the test refactor.
- Browser, native, relay, dweb, proof-demo, snark, and ffi example tests all run in GitHub Actions.
- Existing example docs are updated to describe how to run the tests locally.

## Non-goals

- Do not expand the FFI API as part of this issue.
- Do not change example behavior except where required to make the behavior testable.

## Notes

This issue intentionally broadens the original FFI-only testing gap: `rings.py` still needs integration tests, but the same standard should apply to every example module.

RyanKung:
## Summary

Closes #636.

This PR adds integration-level coverage across the examples and moves the FFI header ownership out of `examples/ffi` into the node crate.

## Changes

- Move `rings.h` to `crates/node/include/rings.h` and update the Python FFI example to load the crate-owned header.
- Add Python tests for `examples/ffi`, including real two-provider offer/answer connectivity and E2E handshake plus stream-frame delivery.
- Split reusable logic out of the native and browser examples so the example behavior has testable pure boundaries.
- Add native, relay, browser, dweb, proof-demo, and snark example tests.
- Add proof-demo browser coverage for two SNARK-registered browser nodes connecting through raw offer/answer.
- Add SNARK WAT coverage and replace the remaining U256 word conversion unwrap path with a typed error.
- Wire E2E send RPC methods through the RPC/node/FFI boundary so examples can exercise the implemented E2E flow.
- Update QACI so example coverage runs in CI.

## Validation

- `cargo test -p rings-node e2e`
- `cargo test -p rings-snark test_wat`
- `cargo check -p rings-rpc`
- `cargo check -p rings-node --features ffi`
- `cargo build -p rings-node --features ffi`
- `RINGS_FFI_REQUIRE_LIBRARY=1 python -m pytest examples/ffi/tests -q`
- `cargo test -p rings-native-example`
- `cargo test -p rings-snark-example`
- `cargo test -p rings-relay-example`
- `cargo test -p rings-browser-example`
- `node --test examples/browser/tests/app-core.test.mjs`
- `wasm-pack test --release --headless --chrome examples/proof-demo`
- `wasm-pack test --release --headless --chrome examples/dweb`
- `wasm-pack test --release --node examples/browser`
- `cargo +nightly fmt --manifest-path examples/proof-demo/Cargo.toml --all -- --check`
- `taplo format --check crates/snark/Cargo.toml examples/native/Cargo.toml examples/proof-demo/Cargo.toml examples/relay/Cargo.toml examples/snark/Cargo.toml`
- `git diff --check`


RyanKung:
## Review — #638 integration coverage for examples (closes #636)

Traced the production-impacting changes and gauged the new test quality. Correct and substantive; no defects found.

### Production changes — verified
- **E2E send RPC wiring is complete and consistent end-to-end.** New `Method::SendE2eHandshake`/`SendE2eMessage` (both `to_str`/`from_str` arms), request/response structs (`SendE2eMessageRequest{destination_did, recipient_public_key, data, max_plaintext_frame_len}` → `{stream_id}`, handshake → `{tx_id}`), `HandleRpc` impls in `rpc_impl.rs` that reuse the #629 Processor APIs (`s2d`/`s2pk`, base64-decode `data`, `frame_len == 0 → DEFAULT_E2E_PLAINTEXT_FRAME_LEN`, `send_e2e_message_with_frame_len`), and the dispatch is wired in `rings_node_handler.rs` (bounds + `Method::SendE2e* =>` arms) so it routes over both JSON-RPC and FFI. No orphaned method.
- **U256 panic removed.** `u256_from_vec_u32 -> Result<U256>` with `const U256_U32_WORDS = 8` replaces the two `try_into().unwrap()` paths, returning `WitnessInvalidU256WordLength{expected, actual}`.
- **Header move is clean.** `examples/ffi/rings.h` fully deleted, `crates/node/include/rings.h` added, and `rings.py` resolves `DEFAULT_HEADER_PATH = crates/node/include/rings.h` — no dangling reference. (E2E goes through the generic `request(method, params)` FFI entry, so no per-method header export is needed; the size delta is just regeneration.)

### Tests — substantive, not smoke
- FFI Python tests assert real behavior: header-path resolution, platform library-path logic, 65-byte signer output, `node_info` round-trip, `create_offer` over the FFI request path, **two FFI providers connecting via offer/answer**, and **two FFI providers exchanging an E2E handshake + stream frames** — i.e. the new RPC methods are exercised end-to-end across the FFI boundary. Library-requiring tests are gated behind `RINGS_FFI_REQUIRE_LIBRARY` / the `ffi_runtime` fixture; pure tests always run.
- Synchronization uses bounded poll-with-deadline (`while monotonic() < deadline … sleep(poll)`), not a blind fixed sleep — the correct pattern (no #620 regression).
- Relay tests assert real TCP/UDP round-trips; the SNARK WAT tests assert calculator metadata and the new typed-error-on-missing-export path.

### Notes (minor)
- Heads-up unrelated to this PR: the `F: PrimeField` diagnostics you may be seeing in `examples/snark` `lib.rs`/`macros.rs` are **local working-tree state on another branch**, not #638 — on this PR branch every generic helper in `examples/snark/src/lib.rs` carries `where F: PrimeField + From<u64>`, and `macros.rs` isn't part of this PR. Nothing to fix here.
- Scope: the PR bundles the example tests (the #636 goal) with a real functional addition (E2E send RPC methods) and a drive-by U256 panic removal. The RPC wiring is justified (the FFI/example tests need those methods to exist) and the panic removal is disclosed; just noting the diff is broader than "tests".

No correctness issues. Acceptance is your call.

RyanKung:
## Follow-up: the snark example's prove/verify is never executed by the test suite

Refining my earlier note. The snark example is the **one** example here whose core behavior isn't actually run by the tests, unlike the others:

- `examples/snark/src/lib.rs` `#[test]`s only assert bundled-asset existence and input *shape* (`*_inputs_match_the_example_shape`). They never call the proving pipeline.
- The real prove/verify (`gen_recursive_circuit` → `foldr` → `verify` → `compress_prove`/`compress_verify`) lives in `bn256.rs` / `merkle_tree.rs`, which are `[[example]]` targets. `cargo test -p rings-snark-example` (and QACI) only **compile** those — `main` is never executed, so their `assert_eq!`s don't run in CI.
- Contrast: `examples/relay/tests/relay.rs` actually runs TCP/UDP round-trips, and the FFI tests actually run a two-provider E2E exchange. The snark example got compile-checks + shape-checks only — a real coverage asymmetry, while the summary reads as if "snark example tests" cover the example.

(The crate's own `test_snark.rs` does run BN256 recursive prove/verify, so SNARK correctness is covered at the crate level — this gap is specifically about the *example's* end-to-end path.)

**Request:** add one fast integration test that actually executes the example's prove→verify path, using the **smallest** circuit and the **fewest folding steps**, so it stays cheap in CI:

- Reuse the new helpers: load `simple_bn256_*` assets, `simple_bn256_initial_input()`, and a **single** private input (one fold step instead of three), call `gen_recursive_circuit`, `foldr` once, then `verify` and assert it succeeds.
- That alone exercises witness calc + circuit gen + one fold + verify end-to-end at minimal cost. If even `compress_prove`/`compress_verify` is too heavy for the default suite, gate just the compress step behind `#[ignore]` or a feature, but keep the fold+verify running by default.

This makes the snark example's coverage match relay/ffi (its actual behavior runs, not just compiles).

RyanKung:
## Re-review — `8887aa8` (snark example coverage)

The requested gap is resolved, and I verified it actually exercises the example's path.

- New `simple_bn256_one_step_prove_verify()` + `#[tokio::test] simple_bn256_example_runs_one_step_prove_verify` runs the real path: load R1CS + witness calculator → `gen_circuit`/`gen_recursive_circuit` → `gen_pp` → `new` → `foldr` → `verify`. So `cargo test -p rings-snark-example` now executes the example's prove/verify, matching the relay/ffi examples (no longer compile-only).
- **Engine/field config is identical to the standalone `bn256.rs`** (`E1=VestaEngine`, `E2=PallasEngine`, `EE=EvaluationEngine`, `S1/S2=RelaxedR1CSSNARK`), so the test faithfully covers the example's actual configuration rather than a different one.
- Kept cheap exactly as suggested: **one** private input → **one** fold (`round = 1`), and compression (`compress_setup`/`compress_prove`/`compress_verify`) is intentionally left to the heavier standalone binary. `tokio` is a dev/runtime dep with `rt-multi-thread` so the `#[tokio::test]` runs.
- Minor native tweak in the same push is correct: drops the unused `init()`/`state == ()` assertion and replaces `effects[0]` with `effects.first().expect(...)` (indexing → checked access).

No new defects. The snark example's coverage is now symmetric with the other examples.

Issue #637, opened 2026-06-28
Introduce guest extensions with WASM and RISC-V runtimes


RyanKung:
## Motivation

Rings needs a dynamically loadable extension model that works in both native and browser runtimes. The current extension system already has the right host shape: each extension is a `(Protocol, Interpret)` pair registered under a namespace, and effects are executed through a namespace-scoped host capability. What is missing is a safe guest-program layer: a plugin that can be loaded at runtime without giving it direct access to transport, storage, files, sockets, clocks, randomness, or cross-namespace dispatch.

We should support both execution families:

- **WASM** for broad portability and browser/native execution.
- **RISC-V** for zkVM-friendly execution, since the RISC-V zkVM ecosystem is currently more mature than zkWASM.

The core abstraction should not depend on a single VM implementation. Rings should define a guest extension ABI and capability boundary, then add WASM and RISC-V runtimes as adapters.

## Proposed Model

Introduce a new guest extension layer:

```text
Guest Extension = Manifest + Guest Binary + GuestRuntime + Host Capability Shell
```

The guest program is a deterministic state transition:

```text
step : (State, Event, Context) -> Result<(State, Effects), GuestError>
```

The guest returns effects as data. The host interprets those effects through the existing extension shell model. The guest must not perform ambient IO directly.

## Runtime Backends

Add a runtime abstraction similar to:

```rust
trait GuestRuntime {
    fn instantiate(manifest: GuestManifest, binary: GuestBinary) -> Result<GuestInstance>;
    fn step(&mut self, input: GuestStepInput) -> Result<GuestStepOutput>;
}
```

Then implement adapters:

- `WasmGuestRuntime`
  - native execution through a WASM runtime
  - browser execution through browser-compatible WASM APIs
  - useful as the development and portability backend
- `RiscvGuestRuntime`
  - native execution/proving adapter
  - browser verification or remote proving path
  - backend-neutral over SP1 / RISC Zero / Jolt style implementations
- `GuestReceiptVerifier`
  - validates a proof receipt against `(program_hash, public_input, public_output)`
  - independent from the concrete proving backend

The core should only model:

```text
GuestProgramHash
GuestPublicInput
GuestPublicOutput
GuestReceipt
```

## Manifest

Define a validated guest manifest carrying at least:

```text
namespace
runtime = wasm | riscv
abi_version
module_hash
state_schema_hash
event_schema_hash
effect_schema_hash
capabilities
memory_limit
fuel_limit
[...]

PR #640, opened 2026-06-28
Introduce guest extension core model

RyanKung:
## Summary

Refs #637. Refs #643.

This PR introduces the runtime-neutral guest extension core model and keeps `rings-node` thin:

- adds a new `rings-runtime` crate for the guest manifest, state-transition model, ABI, receipt contract, runtime registry, and proof adapters
- keeps `rings-node` limited to `GuestProtocol`, `GuestShell`, and `register_guest_extension`, binding `rings-runtime` to the existing `Protocol + Interpret + Scope` extension boundary
- removes direct RISC Zero, arkworks, and bigint proof dependencies from `rings-node`; node `guest-*` features now forward to `rings-runtime` features
- keeps guest effects limited to v1 `Send` and `Inject`, both interpreted only through the namespace-scoped host `Scope`
- adds a validated `GuestManifest` with runtime, ABI version, namespace, hashes, declared capabilities, memory/fuel limits, and proof policy
- adds `GuestRuntime`, `GuestRuntimeAdapter`, `GuestRuntimeProfile`, and profile-based WASM/RISC-V adapter selection
- models proof policy as an adapter capability via `GuestRuntimeProfile`, so `wasm + none`, `wasm + verify_receipt`, `riscv + none`, and `riscv + verify_receipt` are all representable
- adds guest step input/output ABI encoding helpers for runtime adapters
- adds receipt claim verification policy independent from any concrete proving backend

This PR also adds concrete #643 RISC-V zkVM proof work:

- replaces the earlier guest R1CS/Spartan backend with a RISC Zero RISC-V zkVM backend
- adds an opt-in, default-off `guest-riscv-proof` verifier feature and `guest-riscv-prove` native host prover feature
- binds `GuestManifest::module_hash` to the RISC Zero ImageID of the guest ELF
- instantiates a native `Risc0ZkvmRuntime` from an ELF `GuestBinary`, rejects `RISC0_DEV_MODE`, proves with RISC Zero, and stores the serialized RISC Zero receipt in `GuestReceipt::proof`
- verifies receipts with the expected ImageID and checks the Rings receipt wrapper claim
- defines the public RISC Zero journal ABI as `public_input + output_abi`, so `GuestPublicInput` and `GuestPublicOutput` are both bound by the verified journal instead of only by host-side wrapper fields
- avoids `Journal::decode` on arbitrary bytes by validating word alignment before RISC Zero serde decoding
- adds native and wasm32-gated negative tests for malformed receipts/journals, plus pure tests for public input/output claim binding

This PR also implements the WASM verifier side without depending on the upstream `memmap`-bound verifier path:

- adds an opt-in, default-off `guest-wasm-proof` feature
- adds `ZkWasmAggregatorReceipt` for Delphinus/g1024 aggregator payloads: `proof`, `verify_instance`, `aux`, and `target_instances`
- adds `ZkWasmAggregatorVerifier`, a browser-capable verifier that ports the Delphinus continuation-batcher Solidity aggregator verifier into Rust over BN254/arkworks
- binds Rings receipts by requiring `zkwasm_aggregator_claim_scalar(claim)` as the first target instance exposed by the guest circuit
- preserves the generated verifier boundary separately from handwritten parsing, binding, field, curve, and pairing logic
- checks malformed payloads through typed `GuestError` paths, including generated field-subtraction underflow instead of panicking
- adds native negative tests and a wasm32/browser-gated verifier test target
- records the Delphinus Solidity verifier source as MIT-licensed; MIT is compatible with this GPL-3.0 repo

## Reviewer-directed structure

The latest structural review asked for the long-term stack `rings-algebra -> rings-proofs -> rings-runtime -> rings-node`. This update implements the `rings-runtime -> rings-node` part now:

- `rings-runtime` owns guest model/proof-adapter code that had been under `crates/node/src/extension/guest`
- `rings-node` now re-exports runtime APIs and retains only the extension registry/protocol glue
- heavy proof dependencies moved out of `rings-node/Cargo.toml` and are feature-gated in `rings-runtime`

The full `rings-algebra` extraction and `rings-proofs` absorption of `rings-snark` are intentionally left as follow-up work because they require moving core algebra/ecc/snark ownership across the workspace, not just the guest extension layer.

## Remaining proof acceptance fixtures

The RISC-V zkVM host adapter and verifier are implemented, but this PR does not add a positive end-to-end RISC Zero guest ELF fixture. That fixture should be a real guest program that reads `GuestStepInput::encode_abi()` and commits the `public_input + output_abi` journal ABI; this PR avoids adding a fake positive proof.

The WASM verifier no longer depends on the upstream crate that failed on `memmap`, but the full WASM prove/offload acceptance path still needs:

- Delphinus zkWasm aggregator receipts from the Rings prover/offload flow (`SnarkProtocol` / prover peer)
- a valid Delphinus proof fixture or service-backed integration test that exercises the positive verifier path

## Validation

- `taplo format --check`
- `cargo +nightly fmt --all --check`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo check -p rings-runtime`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo check -p rings-node --features node_default,guest-riscv-proof,guest-wasm-proof`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo test -p rings-runtime --features risc0-proof,wasm-proof`
- `RISC0_SKIP_BUILD_KERNELS=1 CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo check -p rings-runtime --features risc0-prove,wasm-proof`
- `RISC0_SKIP_BUILD_KERNELS=1 CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo test -p rings-runtime risc0_zkvm --features risc0-prove,wasm-proof`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo test -p rings-node extension::guest --features node_default,guest-riscv-proof,guest-wasm-proof`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-browser-target cargo check -p rings-node --no-default-features --features browser_default,guest-riscv-proof,guest-wasm-proof --target wasm32-unknown-unknown --tests`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo clippy -p rings-runtime --features risc0-proof,wasm-proof --tests -- -D warnings`
- `RISC0_SKIP_BUILD_KERNELS=1 CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo clippy -p rings-runtime --features risc0-prove,wasm-proof --tests -- -D warnings`
- `CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo clippy -p rings-node --features node_default,guest-riscv-proof,guest-wasm-proof --tests -- -D warnings`
- `RISC0_SKIP_BUILD_KERNELS=1 CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo check -p rings-node --features node_default,guest-riscv-prove,guest-wasm-proof`
- `RISC0_SKIP_BUILD_KERNELS=1 CARGO_TARGET_DIR=/private/tmp/rings-guest-extensions-637-target cargo clippy -p rings-node --features node_default,guest-riscv-prove,guest-wasm-proof --tests -- -D warnings`

`RISC0_SKIP_BUILD_KERNELS=1` was only needed locally because this macOS machine does not have the Xcode Metal Toolchain installed; it lets the native RISC Zero Rust API compile without building Metal kernels.


RyanKung:
## Review — #640 guest extension core model (refs #637)

Reviewed this as a capability/sandbox boundary. The model is sound, honestly abstracted, and well-tested; one validation hole worth closing before a real loader is wired.

### Verified correct
- **Sandbox is enforced on the execution path, not just declared.** `GuestProtocol::step` runs the runtime then routes through `accept_step_output` (resource limits → capabilities → receipt policy) *before* emitting any effect; on any failure it returns `Transition::pure(rejected(error))` with no effects. Test `guest_protocol_denies_undeclared_effect_before_shell_boundary` proves an undeclared `Send` never reaches the shell.
- **Effects are closed to `{Send, Inject}`** and `GuestShell` interprets them only through the namespace-scoped `Scope` (`scope.send` / re-inject), so a guest cannot act outside the namespace fixed by its manifest. Registry rejects duplicate namespace and duplicate adapter.
- **Adapters are honest abstractions, not panic stubs.** `WasmGuestRuntime<F>` / `RiscvGuestRuntime<F>` wrap an instantiation factory; a missing backend is a typed `RuntimeUnavailable`, not `unimplemented!()`. This is the right way to ship "core boundary, no concrete runtime" with no stopgaps.
- **ABI** is bincode with errors mapped to `AbiEncode`/`AbiDecode` (no `unwrap`/indexing); round-trip and reject-invalid-bytes are tested.
- **Receipt policy** binds `program_hash + public_input + public_output` into the claim, checks it against the receipt, then defers to the injected `GuestReceiptVerifier` — backend-independent, with required/mismatch/verify-failed/accepted all tested.
- Binary hash is matched against `manifest.module_hash()` before instantiate; manifest validation (namespace chars, abi version, duplicate capability, non-zero limits) and the full `accept_step_output` matrix (fuel/memory/capability/receipt/determinism) are covered by tests.

### Findings

**1. Validated types derive `Deserialize`, which bypasses their validating constructors (close before a loader exists).**
`GuestProgramHash::new` rejects an all-zero hash (`EmptyHash`), and `GuestManifest::validate` checks namespace/abi — but both types `#[derive(Deserialize)]`, so deserialization constructs them directly without those checks. `GuestManifestSpec` is documented as "raw manifest data supplied by a loader" and itself derives `Deserialize` with `GuestProgramHash` fields, so a manifest loaded from bytes carries hashes that never passed `EmptyHash`, and `validate()` does not re-check them (a zero `module_hash` would even match a zero binary hash in `validate_manifest`). No untrusted deserialization exists in this PR yet, so it's latent — but these types are `Serialize/Deserialize` precisely to be loaded from bytes, so it should be closed before the loader is written. Fix: a validating `Deserialize` for `GuestProgramHash` (or re-run `GuestProgramHash::new` on all four hashes inside `validate()`), and don't derive `Deserialize` on the validated `GuestManifest` — deserialize only `GuestManifestSpec` and always go through `validate`.

**2. Resource limits are checked against guest-self-reported `fuel_used` / `memory_pages_used`.**
`validate_resource_limits` compares the manifest limits to values the runtime *reports* in its output. A runaway guest is only actually bounded if the (future) adapter meters and traps during execution; this post-hoc check can't stop a guest that exceeds a limit but under-reports. Correct for a core boundary, but make it an explicit adapter contract: the manifest limit is authoritative and the adapter MUST enforce it in-flight; the host check is a backstop.

**3. Minor: the `state`/`event`/`effect` schema hashes are stored but never used** to validate the corresponding bytes at this layer — inert commitments for now. Fine if the runtime/proof layer consumes them later; worth a doc line that this core layer does not enforce schema conformance.

No correctness defects in the implemented logic. The headline is #1 (validating-constructor bypass on a security type). Acceptance is your call.

RyanKung:
## Follow-up: the WASM/RISC-V abstraction and its test coverage

### What the right algebraic shape is

The guest transition is a **deterministic Mealy machine / state-transition coalgebra**:
`step : State × Event × Context → Result<State × Effects (× receipt data)>`, and `assert_deterministic_replay` pins it to a pure morphism (no randomness/clock). "Runtime-neutral" is then the **one-signature / many-interpreters** pattern: `GuestRuntime::step` (plus the ABI boundary) is the signature, and `GuestRuntimeRegistry: Map<GuestRuntimeKind, Adapter>` is an **indexed family of interpreters** selected by a tag. That registry shape is the correct abstraction.

Where it doesn't hold up: `WasmGuestRuntime<F>` and `RiscvGuestRuntime<F>` are byte-identical — both are `{ factory: F }` whose only difference is `runtime()` returning the `Wasm` vs `Riscv` constant and then calling `(self.factory)(…)` the same way. So `GuestRuntimeKind` is being modeled twice: once as data (the registry key) and once as two empty types. The split:
- adds no type-level invariant — the `runtime()` tag is a self-declared label decoupled from what the factory actually builds (you can hand a riscv-building factory to `WasmGuestRuntime::new` and it registers under `Wasm`), and
- is pure duplication.

Two reasonable fixes:
1. Collapse to **one tagged adapter** — make the index data: `struct FnAdapter<F> { kind: GuestRuntimeKind, factory: F }`. That is the honest indexed-family-of-interpreters; the registry already keys on `kind`.
2. Or, if WASM and RISC-V are meant to differ, make the two types carry the **genuinely different structure** so the distinction earns a type. The real axis is the *verification strategy*: WASM ↔ replay-determinism (`ProofPolicy::None`, the host re-executes), RISC-V/zkVM ↔ receipt verification (`ProofPolicy::VerifyReceipt`, the receipt *is* the proof the step was computed correctly). Today `runtime` and `proof_policy` are **independent** manifest fields with no consistency constraint, so the manifest can express meaningless states — a WASM manifest with `VerifyReceipt` (but no prover), or a RISC-V/zkVM manifest with `ProofPolicy::None` (defeating the zkVM). The meaningful state space is a strict subset of that product; coupling the kind to its verification mode (or at least validating the combination) would close that.

### Are there WASM/RISC-V tests, and are they complete?

There are tests: `wasm_test_factory` / `riscv_test_factory` + `runtime_registry_selects_wasm_and_riscv_adapters_from_manifest`, which assert the registry routes a Wasm manifest to the Wasm adapter and a Riscv manifest to the Riscv adapter (both directions, distinct `"wasm"`/`"riscv"` output). But because the wrappers are tag-only, these can only verify **tag routing**, not backend behavior (there is none yet — by design).

For the backend boundary the coverage is **not complete**:
- `RuntimeUnavailable` — instantiating a manifest whose kind has no registered adapter — is **not** tested.
- `DuplicateRuntimeAdapter` — registering the same kind twice — is **not** tested.
- The tag↔behavior mismatch above is neither tested nor preventable.
- The `runtime` × `proof_policy` consistency is neither tested nor enforced.
- Real WASM/RISC-V execution is necessarily absent (no concrete runtime — acceptable for a core-boundary PR).

So: the selection happy-path is covered; the two registry error paths and the cross-field invariants are gaps worth filling now, and the WASM/RISC-V split should either collapse to a tagged adapter or grow the structure that actually differs (proof/verification).

RyanKung:
## Re-review — `72c5e10` (guest extension review feedback)

All findings from both the main review and the WASM/RISC-V follow-up are resolved, with tests. Verified the new code, not just its presence.

- **Validated types no longer bypass their constructors (main #1).** `GuestProgramHash` drops the derived `Deserialize` for a custom one that runs `new` (rejects zero) — test `program_hash_deserialization_rejects_zero_hash`. `GuestManifest` drops `Deserialize` entirely (doc: deserialize `GuestManifestSpec` and call `validate`). And `validate()` re-checks all four hashes via `validate_hash` with proper field labels — test `manifest_validation_rejects_hashes_that_bypass_constructor`. Triple-covered, and since `GuestManifestSpec`'s hash fields now use the validating `Deserialize`, the real load path rejects zero hashes at decode time.
- **runtime × proof_policy is now constrained (follow-up).** `GuestRuntimeKind::required_proof_policy` (Wasm→None, Riscv→VerifyReceipt) + `validate_runtime_proof_policy` reject inconsistent combinations with `InvalidProofPolicyForRuntime` — test covers both WASM+VerifyReceipt and RISC-V+None. The test manifests were correctly updated (riscv now uses `VerifyReceipt`).
- **WASM/RISC-V empty split collapsed (follow-up).** Both wrappers are replaced by one `GuestRuntimeFnAdapter { runtime: GuestRuntimeKind, factory }` whose `runtime()` returns the field — the kind is honest data, not two identical pseudo-typed structs.
- **Registry error paths now tested (follow-up).** Added `runtime_registry_rejects_unregistered_runtime` (`RuntimeUnavailable`) and `runtime_registry_rejects_duplicate_adapter_kind` (`DuplicateRuntimeAdapter`).
- **Docs added (main #2/#3).** `GuestRuntime` now states limits are authoritative and adapters must meter/trap in flight (host check is a backstop); `GuestManifestSpec` states the schema hashes are bound but not parsed for conformance at this layer.

### Minor (forward-looking, non-blocking)
The runtime↔proof_policy mapping is a strict bijection (Wasm⇔None, Riscv⇔VerifyReceipt). That's the right minimal constraint for v1, but it precludes proving-WASM (zkWASM, which would want VerifyReceipt) and plain non-proving RISC-V. If those become real, a capability check ("VerifyReceipt requires a proving-capable runtime") would generalize better than hardcoding kind→policy. Not needed now.

No new defects. Acceptance is your call.

RyanKung:
## Re-review — Spartan R1CS guest proof backend (`971b5dd`, `e2a11a4`, `930f0b4`)

(The PR rebased onto merged #641, so the `measure.rs`/`transport` lines in the range diff are #641 noise; this reviews the new guest-proof work.) This is a real, well-built R1CS proving backend — it closes the #640 proof hole. Verified, not rubber-stamped.

### Verified correct
- **Real prove + verify (the #640 hole is closed).** `prove` builds the `libspartan` `Instance`/`VarsAssignment`/`InputsAssignment`, checks `instance.is_sat` first, then `SpartanSnark::prove`; `verify` runs `SpartanSnark::verify` — actual cryptographic verification of `receipt.proof`, not the public-value-equality stub.
- **Sound claim binding.** `spartan_r1cs_claim_scalar` = `keccak256(domain ‖ program_hash ‖ len(public_input) ‖ len(public_output))`, injected as the **first public input** of the R1CS instance, so a proof is bound to its `(program_hash, public_input, public_output)` and can't be replayed across claims. `GuestReceiptVerifier::verify` also pre-checks `program_hash ==` and `matches_receipt` before the crypto verify.
- **Profile model (the requested kind↔proof fix).** `GuestRuntimeProfile { runtime, proof_policy }` replaces the strict `required_proof_policy` bijection; the registry keys adapters by profile, and `GuestRuntimeFnAdapter` is profile-tagged. Clean generalization.
- **wasm32 verification is tested.** `#[cfg(target_arch="wasm32")] mod wasm32` runs `wasm_bindgen_test` (`run_in_browser`) for both accept-real-receipt and reject-tampered-proof.
- **Negative tests** reject via the real verifier: tampered proof → `ProofDataDecode`/`ReceiptVerificationFailed`; wrong `public_output` → `ReceiptClaimMismatch`; mismatched `program_hash` → rejected.
- **Hygiene:** opt-in `guest-r1cs-proof` feature (spartan/curve25519-dalek/merlin/getrandom-js all `optional`, default-off); in-flight `enforce_in_flight_limits` before proving; production path panic-free (`unwrap_or` saturation; `expect` only in tests); `assert_deterministic_replay` correctly *not* applied to the proving runtime.

### Minor notes
1. **Non-deterministic proofs vs the #640 determinism model.** Spartan proving is randomized (`getrandom`/`js` is pulled in), so `step` is not reproducible in the `proof` bytes; `GuestStepOutput`'s `PartialEq` (includes `receipt`) and `assert_deterministic_replay` can't apply to proving runtimes. Only the logical output (`state`/`effects`/`public_output`) is deterministic — worth a doc line saying so.
2. **Profile consistency moved from validation to registration.** A manifest can validate with any `(kind, proof_policy)`; an unsupported profile only fails later at `instantiate` (`RuntimeUnavailable`). Acceptable, but a `validate`-time check would fail closer to the source.

### Structural direction (decided): `rings-algebra` → `rings-proofs` → `rings-runtime`, with a thin node

Split the conflated concerns into a layered stack so `rings-node` stays thin and the heavy crypto/algebra deps leave `rings-core`:

- **`rings-algebra`** (new foundational leaf): the trait tower (`Zero`/`One`/`AbelianGroup`/`CommutativeRing`/`Field`/`Module`/`JoinSemilattice`, today's `core/src/algebra.rs`) **and** all Field/Group/Ring implementations (today's `core/src/ecc/group.rs`: `CurveGroup`/`CurveScalarField`/`Point`/`Scalar`/`CyclicModule` + the secp256k1/secp256r1/bls12381/ristretto adapters), curve backends feature-gated. The heavy curve deps (`libsecp256k1`, `p256`, `bls12_381`, `curve25519`, `pasta`) move here, out of `rings-core`.
- **`rings-proofs`** (new; **absorbs `rings-snark`**; depends on `rings-algebra`): all proof systems behind one `ProofSystem` (`prove`/`verify`) trait — Nova/folding, Spartan/R1CS (move `r1cs_spartan`'s prove/verify primitives here), future zkWASM. Name is `proofs` not `zkp` (succinct ≠ zero-knowledge). Heavy prover deps (`spartan`, `nova-snark`, `merlin`, …) feature-gated here.
- **`rings-runtime`** (new; depends on `rings-proofs`): the guest-program runtime layer — `manifest`, the runtime trait(s), `GuestRuntimeProfile`/registry, the receipt/verifier contract (verifier a thin adapter over a `rings-proofs` `ProofSystem`). Naming note: `runtime` names the layer by its central trait but no longer signals the **capability/sandbox** nature and mildly overloads "runtime" vs the node's tokio/`RuntimeFlavor` — keep the sandbox/capability model explicit in this crate's docs (or use `rings-sandbox` if the security meaning should be in the name).
- **`rings-core`** (depends on `rings-algebra`): DHT/transport, with the algebra/curve weight removed.
- **`rings-node`** (depends on `rings-runtime`, `rings-core`): as thin as possible — only `GuestProtocol`/`GuestShell`/`register_guest_extension`, the glue binding `rings-runtime` to node's `extension::ext`. No proof/curve deps enter node's tree.

Dependency DAG (acyclic): `rings-algebra` (leaf) ← {`rings-proofs`, `rings-core`}; `rings-proofs` ← `rings-runtime` ← `rings-node`.

Boundary caveats:
- **`Did` stays in `rings-core`** and implements the `rings-algebra` traits there — moving its group impl into `rings-algebra` would make algebra depend on a core type (cycle). The rule "impls live with the type, traits come from `rings-algebra`" applies to any core-owned algebraic type.
- **ElGamal is a cryptosystem over groups, not a Field/Group/Ring** — it belongs in a crypto layer (co-located with the curve groups in `rings-algebra`, or a separate `rings-crypto`), not in the trait tower itself. Sub-decision; doesn't block.
- **External provers bring their own field types** (`nova-snark`/pasta, `spartan`/curve25519, `ff::PrimeField`). `rings-algebra` is rings' own algebra; the proof backends interoperate at the boundary (convert), not by re-expressing nova/spartan over `rings-algebra` traits.

Acceptance is your call.

crates/runtime/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+pub use manifest::GuestCapabilities;
+pub use manifest::GuestCapability;
+pub use manifest::GuestManifest;
+pub use manifest::GuestManifestSpec;
+pub use manifest::GuestProgramHash;
+pub use manifest::GuestResourceLimits;
+pub use manifest::GuestRuntimeKind;
+pub use manifest::ProofPolicy;
+pub use manifest::SUPPORTED_GUEST_ABI_VERSION;
+#[cfg(all(feature = "risc0-prove", not(target_arch = "wasm32")))]
```
RyanKung, line 28: since all parts art not compiled in wasm32, then ,why not set the cfg at mod layer.

RyanKung:
## Re-review — RISC Zero + zkWasm backends and the `rings-runtime` crate (`36269b3`, `889b116`, `82a63b7`)

Direction is right: dropping the R1CS/Spartan fixed-circuit backend for **RISC Zero (RISC-V zkVM)** as the `Riscv` runtime is the intended choice — risc0 for RISC-V is the correct use. Extracting `rings-runtime` is also the right move. But the layering and test coverage have real problems.

### Layering (this needs fixing)

1. **`rings-runtime` conflates the model with the proof backends.** It holds the runtime/sandbox model (`manifest`, runtime traits, `GuestRuntimeProfile`, receipt contract) **and** two heavy proof implementations (`risc0_zkvm`, `zkwasm_aggregator` + the 2,893-line generated verifier) in one crate. Per the agreed split, the backends belong in **`rings-proofs`** behind the unified `ProofSystem`; `rings-runtime` should hold the model and *depend on* `rings-proofs`. As is, the runtime crate is the new dumping ground that node used to be.
2. **No `rings-algebra` crate.** The algebra tower + Field/Group/Ring impls are still in `rings-core` (`algebra.rs`, `ecc/group.rs`); the foundational crate we agreed on doesn't exist yet. risc0/zkwasm pull `ark-bn254`/`ark-ec`/`ark-ff` into `rings-runtime` while the curve/field abstractions live elsewhere — exactly the duplication `rings-algebra` is meant to prevent.
3. **The 2,893-line generated verifier sits in `rings-runtime`.** A mechanically-transpiled crypto verifier of that size is a layer of its own; it belongs in the proofs layer (or its own crate), not next to the sandbox model.
4. **Scattered cfg.** `#[cfg(all(feature = "risc0-prove", not(target_arch = "wasm32")))]` is repeated ~20× across `risc0_zkvm.rs`. Consolidate behind one module/`cfg` boundary (e.g. a `prove` submodule gated once) instead of per-item.

Target after refactor: `rings-algebra` (leaf) ← `rings-proofs` (risc0 + zkwasm + nova behind `ProofSystem`) ← `rings-runtime` (model only) ← thin `rings-node`.

### Correctness / test gaps (more serious than the layering)

1. **The zkWasm verifier is never tested against a real Delphinus proof.** Every test in `zkwasm_aggregator.rs` builds a synthetic `ZkWasmAggregatorReceipt` in code and only exercises **rejection** paths (missing claim binding, malformed proof). There is no `include_bytes!` test vector from an actual Delphinus continuation-batcher proof, so the **acceptance** path of a 2,893-line verifier mechanically translated from Solidity is completely unverified — it could reject all real proofs, or accept invalid ones, and the suite would stay green. For a transpiled crypto verifier this is the highest risk in the PR. Add a real-proof vector: accept a genuine valid proof, and reject a tampered version of it.
2. **risc0 verify on `wasm32` is compiled-in but untested.** `Risc0ReceiptVerifier::verify` is *not* `not(wasm32)`-gated (good — browser nodes should verify), but there is **no `wasm32` verify test for risc0** (zkwasm has one; risc0 doesn't), and it is unconfirmed that `risc0-zkvm`'s verify path even builds on `wasm32`. Add a `wasm32` risc0 verify test, or at minimum confirm the verifier compiles under the browser clippy gate — otherwise the "verifier must run in the browser" requirement isn't met for the RISC-V runtime.

### Smaller
- Footprint: `Cargo.lock` +2,172 from risc0 + arkworks. Feature-gated (`risc0-proof`/`risc0-prove`/`wasm-proof` opt-in, default `std`), so default builds stay clean — good — but risc0 is a heavy tree when enabled.
- Claim binding is sound on both backends (risc0: journal must match `program_hash + public_input + public_output`; zkwasm: `keccak(domain ‖ program_hash ‖ len(input) ‖ len(output))` as the first target instance) — same pattern as the deleted Spartan backend. Good.

Bottom line: the backend choices are right, but (a) split proofs/algebra out of `rings-runtime`, and (b) the zkWasm verifier must be proven against a real proof and risc0 verify must be tested on `wasm32` before this is trustworthy. Acceptance is your call.

RyanKung:
Update after the latest review, pushed as `a29f68e5e47629d008d0161d7d5576ecc523c0c3`.

Implemented:
- Moved proof backends out of `rings-runtime` into a new `rings-proofs` crate.
- Moved the generated zkWasm verifier into `crates/proofs/src/zkwasm/generated.rs`.
- Kept `rings-runtime` as thin adapters over the runtime ABI and receipt model.
- Split RISC Zero native proving behind a `prove` module boundary; the wasm-capable verifier path remains compiled for `wasm32`.
- Added QACI browser gates for proof adapters:
  - `cargo clippy -p rings-runtime --features browser,risc0-proof,wasm-proof --target=wasm32-unknown-unknown --tests`
  - `cargo clippy -p rings-node --features browser_default,guest-riscv-proof,guest-wasm-proof --target=wasm32-unknown-unknown --tests`
  - `cargo test -p rings-runtime --release --target=wasm32-unknown-unknown --features browser,risc0-proof,wasm-proof`

Local validation:
- `cargo check -p rings-proofs --features risc0-proof,wasm-proof`
- `cargo test -p rings-proofs --features risc0-proof,wasm-proof`
- `cargo test -p rings-runtime --features risc0-proof,wasm-proof`
- `cargo clippy -p rings-runtime --features risc0-proof,wasm-proof --tests -- -D warnings`
- `cargo clippy -p rings-runtime --features browser,risc0-proof,wasm-proof --no-deps --no-default-features --target=wasm32-unknown-unknown --tests -- -D warnings`
- `cargo clippy -p rings-node --features browser_default,guest-riscv-proof,guest-wasm-proof --no-deps --no-default-features --target=wasm32-unknown-unknown --tests -- -D warnings`
- `RISC0_SKIP_BUILD_KERNELS=1 cargo check -p rings-runtime --features risc0-prove,wasm-proof`

Explicit remaining blocker:
- I did not add a fake zkWasm positive accept test. This repository does not contain a genuine Delphinus `proof / verify_instance / aux / target_instances` vector, and a shallow check of `zkcrossteam/g1024` also did not find a static valid proof fixture. A real accept-and-tamper test needs an actual Delphinus service output fixture first.

Local browser execution note:
- The wasm test binary compiles locally. Running it locally is blocked by this machine's webdriver setup: with matching `wasm-bindgen-cli 0.2.100`, `wasm-bindgen-test-runner` starts geckodriver and then fails session creation with status 500/SIGKILL. The PR CI step is now wrapped in `setup-xvfb`, so CI is the correct browser execution authority here.

RyanKung:
Follow-up CI fix pushed as `2a732a22c933201f3e730aa30df18b8c914bf1f7`: `rings-proofs` helper functions are now gated by backend features so the default-feature workspace build no longer trips `-D dead-code`. Local validation added after the fix: `cargo check -p rings-proofs`, `cargo build --all --verbose`, and `cargo test -p rings-proofs --features risc0-proof,wasm-proof`.

RyanKung:
Second CI follow-up pushed as `ad569d508f04228dc277a6573a734df672b9901e`: narrowed `data_encode` cfg so default-feature `cargo clippy --all --tests -- -D warnings` does not compile unused proof encode helpers. Local validation now includes the exact lint command: `cargo clippy --all --tests -- -D warnings`.

RyanKung:
## Re-review — `rings-proofs` split + feature gating (`a29f68e`, `2a732a2`)

The layering feedback is addressed well; two structural items and one real test gap remain.

### Resolved
- **`rings-proofs` crate with a unified `ProofSystem` trait.** Not just relocated backends: `ProofSystem { fn verify(&self, claim: &ProofClaim, proof) -> Result<Self::Verified, ProofError> }` is the shared verify interface that both `risc0` and `zkwasm` implement (proving stays backend-specific/native, e.g. `Risc0Prover::prove` — the right shape, since verify is the universal browser op). The 2,893-line generated verifier now lives in `proofs/src/zkwasm/generated.rs`.
- **`rings-runtime` is now thin.** `runtime/src/{risc0,zkwasm}.rs` are adapters that wrap a `rings-proofs` `ProofSystem` into `GuestReceiptVerifier`, plus `proofs.rs` boundary mappers (`proof_claim_from_guest`, `proof_error`). The crypto left the runtime crate.
- **cfg consolidated** into `risc0/prove.rs` instead of ~20 scattered per-item attributes; **risc0 now has a `wasm32` verify test**, confirming the verifier builds/runs on `wasm32`.

### Still open
1. **No `rings-algebra` crate.** The algebra tower + Field/Group/Ring impls (`core/src/algebra.rs`, `core/src/ecc/group.rs`) are still in `rings-core`. The foundational crate isn't started — this was an explicit ask.
2. **`rings-snark` is not merged into `rings-proofs`.** Two proof crates now coexist: `rings-proofs` (risc0/zkwasm, behind `ProofSystem`) and `rings-snark` (Nova, not behind it). The agreed direction is that `rings-proofs` absorbs `rings-snark` so all proof systems sit behind one `ProofSystem`. Right now Nova is the odd one out.

### Test gap (the real risk): neither verifier has a positive/acceptance test
Every test for both backends — native and `wasm32` — only exercises **rejection** (malformed ELF/receipt, missing claim binding, claim mismatch). Nothing feeds a **real, valid proof** and asserts it verifies. A verifier that rejects everything would pass the entire suite.
- **zkWasm (highest risk):** the 2,893-line verifier mechanically translated from Delphinus's Solidity has *no* acceptance test, and there's no in-repo prover to produce one — so it needs a captured real Delphinus continuation-batcher proof as an `include_bytes!` vector: assert a genuine proof verifies, and a tampered copy is rejected. Until then its accept path is entirely unverified.
- **risc0:** add a native, feature-gated prove→verify roundtrip via `Risc0Prover` (`#[ignore]` if too slow for default CI) so the accept path is covered at least once.

Bottom line: layering is in good shape (proofs split out, runtime thin). Remaining: extract `rings-algebra`, fold `rings-snark` into `rings-proofs`, and — most important — prove each verifier accepts a real proof, not just rejects synthetic ones. Acceptance is your call.

PR #641, opened 2026-06-28
Prioritize DHT connections by peer quality

### LINKED ISSUE #639: Use peer quality measurements to prioritize DHT connections
## Problem

Rings already has a peer measurement abstraction, but the current runtime does not use it as an active DHT signal.

Current witnesses:

- `crates/core/src/measure.rs` defines `MeasureCounter`, `Measure`, `BehaviourJudgement`, and default behaviour predicates.
- `crates/node/src/measure.rs` implements `PeriodicMeasure` with per-`Did` periodic counters and storage-backed persistence.
- `crates/node/src/provider/mod.rs` constructs `PeriodicMeasure` and wires it through `ProcessorBuilder` and `SwarmBuilder`.
- `crates/core/src/swarm/transport.rs` stores `measure: Option<MeasureImpl>`, but the field is currently dead.
- Production paths do not appear to call `Measure::incr` for transport events or `BehaviourJudgement::good` for DHT connection decisions.

As a result, Rings has the shape of a peer scoring system, but DHT connectivity does not currently prefer locally observed reliable peers.

## Goal

Make DHT connection scheduling prefer high-quality peers according to local observations.

The quality signal should influence connection priority only. It must not change Chord ring ownership, successor responsibility, storage placement, or correctness-critical successor/finger semantics.

## Required model

Introduce an explicit local peer-quality model instead of using raw counters directly at call sites.

Suggested state relation:

- State: local quality observations keyed by `Did` and event kind.
- Inputs: connection, disconnection, send success, send failure, receive success, receive failure.
- Derived value: a small semantic judgement such as `Healthy`, `Unknown`, or `Degraded`, or an equivalent ordered quality type.
- Output: a stable ordering or partition for candidate peers used by DHT connection scheduling.

Important laws:

- Unknown peers must not be permanently excluded.
- Degraded peers may be deprioritized, but they remain candidates when needed for liveness.
- Quality ordering is advisory: it selects which candidate to try first, not whether the candidate is a valid Chord peer.
- The candidate set must remain finite and must not gain or lose peers except through the original DHT transition that produced it.
- Observation updates are side effects at transport/callback boundaries; DHT candidate ordering should be a pure transformation from candidates plus quality evidence.

## Implementation outline

1. Define named peer-quality predicates or types.
   - Avoid naked comparisons such as raw failure-count checks at DHT call sites.
   - Make the judgement semantics explicit and documented.

2. Record observations at named effect boundaries.
   - Increment connection/disconnection counters from connection lifecycle callbacks.
   - Increment send/receive success and failure counters from transport send/receive paths where the outcome is known.
   - Keep transport I/O and storage effects outside pure DHT ordering logic.

3. Feed quality into DHT connection scheduling.
   - Prefer high-quality peers in connection actions generated by stabilization, successor/finger maintenance, and find-successor follow-up paths.
   - Preserve fallback behavior so DHT liveness does not depend on every peer being marked healthy.

4. Revisit the existing `ConnectBehaviour` predicate.
   - The current `(connect - disconnected) < threshold` shape appears inconsistent with the `CONNECT_FAILED_LIMIT` name and can make repeated disconnects easier to pass.
   - Replace it with a named proposition that matches the intended failure model.

5. Remove dead wiring.
   - `SwarmTransport.measure` should either be consumed by the runtime or be replaced by a smaller explicit quality-observation boundary.

## Tests and formal coverage

Acceptance requires tests that witness the model, not only direct counter increments.

Required coverage:

- Pure ordering tests: healthy peers are tried before unknown peers, unknown peers before degraded peers, and all original candidates remain present.
- Event-boundary tests: transport/callback outcomes increment the intended counters exactly once.
- Predicate tests: repeated disconnects or send failures eventually degrade a peer; successful/unknown peers are not falsely degraded.
- DHT schedule/topology tests: under a mixed-quality candidate set, connection scheduling prefers healthy peers while preserving convergence fallback when only degraded peers are available.
- Persistence tests: restarted nodes preserve enough recent evidence to keep the same local judgement, within the existing periodic-window semantics.

Formal comments should name the safety invariant at the ordering boundary, for example:

- Invariant: quality ordering is a permutation or stable partition of the DHT-produced candidate set.
- Preservation: applying quality order never changes ring responsibility or storage placement.
- Liveness expectation: degraded peers are deprioritized, not removed, so fallback attempts remain possible.

## Non-goals

- No global reputation system.
- No Sybil-resistance claim.
- No cryptographic trust score.
- No change to Chord key ownership, successor responsibility, or storage placement.

## Acceptance criteria

- Runtime code records peer-quality observations from real network outcomes.
- DHT connection scheduling consumes the peer-quality judgement when ordering candidates.
- The old dead `measure` wiring is removed or made live.
- `ConnectBehaviour` semantics are corrected or replaced by a clearly named predicate.
- The implementation has contract-focused docs/formal comments for the quality boundary.
- Tests cover ordering, event observation, fallback/liveness, and persistence semantics.
- No production `unwrap`, `expect`, `panic!`, unchecked indexing, or stringly typed domain errors are introduced.

RyanKung:
## Summary

- closes #639
- add an advisory peer-quality model derived from existing measure counters
- record peer observations at transport/callback side-effect boundaries
- order DHT-produced connection candidates by local quality without changing Chord ownership, successor, or storage placement semantics
- bump the workspace and tracked lockfiles to 0.12.0

## Correctness notes

- `PeerQuality` is local evidence only: `Healthy`, `Unknown`, or `Degraded`.
- Candidate ordering is a stable permutation, so it changes connection attempt order only.
- Disconnect observations are bounded to one `Disconnected` counter per connection epoch.
- Send/receive observations are recorded at transport delivery and callback validation boundaries.

## Validation

- `cargo test -p rings-core --features dummy measure::tests -- --nocapture`
- `cargo test -p rings-core --features dummy swarm::transport::tests -- --nocapture`
- `cargo test -p rings-node --features node measure::tests -- --nocapture`
- `cargo test -p rings-core --features dummy test_message_handler -- --nocapture`
- `cargo test -p rings-core --features dummy test_stabilization -- --nocapture`
- `cargo test -p rings-core --features dummy dht_schedule -- --nocapture`
- `cargo check --workspace --all-targets`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo doc --workspace --no-deps`
- `git diff --check`
- `rg "0\\.11\\.0" Cargo.toml Cargo.lock crates examples -S`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4185,7 +4185,7 @@ dependencies = [
 
 [[package]]
 name = "rings-browser-example"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "bytes",
  "js-sys",
@@ -4197,7 +4197,7 @@ dependencies = [
 
 [[package]]
 name = "rings-core"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "ark-bls12-381",
  "ark-ec",
@@ -4267,7 +4267,7 @@ dependencies = [
 
 [[package]]
 name = "rings-derive"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -4277,7 +4277,7 @@ dependencies = [
 
 [[package]]
 name = "rings-native-example"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "async-trait",
  "base64 0.13.1",
@@ -4291,7 +4291,7 @@ dependencies = [
 
 [[package]]
 name = "rings-node"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "anyhow",
  "arrayref",
@@ -4351,7 +4351,7 @@ dependencies = [
 
 [[package]]
 name = "rings-relay-example"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "rings-core",
  "rings-node",
@@ -4360,7 +4360,7 @@ dependencies = [
 
 [[package]]
 name = "rings-rpc"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "async-trait",
  "base64 0.13.1",
@@ -4376,7 +4376,7 @@ dependencies = [
 
 [[package]]
 name = "rings-snark"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "bellpepper-core",
  "byteorder",
@@ -4406,7 +4406,7 @@ dependencies = [
 
 [[package]]
 name = "rings-snark-example"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "rings-snark",
  "tokio",
@@ -4414,7 +4414,7 @@ dependencies = [
 
 [[package]]
 name = "rings-transport"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "async-trait",
  "bincode",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ resolver = "2"
 members = ["crates/*", "examples/browser", "examples/native", "examples/relay", "examples/snark"]
 
 [workspace.package]
-version = "0.11.0"
+version = "0.12.0"
 edition = "2021"
 license = "GPL-3.0"
 authors = ["RND <dev@ringsnetwork.io>"]
@@ -13,12 +13,12 @@ repository = "https://github.com/RingsNetwork/rings-node"
 async-trait = "0.1.77"
 js-sys = "0.3.64"
 jsonrpc-core = "18.0.0"
-rings-core = { path = "crates/core", version = "0.11.0", default-features = false }
-rings-derive = { path = "crates/derive", version = "0.11.0", default-features = false }
-rings-node = { path = "crates/node", version = "0.11.0" }
-rings-rpc = { path = "crates/rpc", version = "0.11.0", default-features = false }
-rings-snark = { path = "crates/snark", version = "0.11.0", default-features = false }
-rings-transport = { path = "crates/transport", version = "0.11.0" }
+rings-core = { path = "crates/core", version = "0.12.0", default-features = false }
+rings-derive = { path = "crates/derive", version = "0.12.0", default-features = false }
+rings-node = { path = "crates/node", version = "0.12.0" }
+rings-rpc = { path = "crates/rpc", version = "0.12.0", default-features = false }
+rings-snark = { path = "crates/snark", version = "0.12.0", default-features = false }
+rings-transport = { path = "crates/transport", version = "0.12.0" }
 serde-wasm-bindgen = "0.6.5"
 wasm-bindgen = "0.2.100"
 wasm-bindgen-futures = "0.4.37"
--- a/crates/core/src/measure.rs
+++ b/crates/core/src/measure.rs
@@ -55,11 +55,6 @@ impl PeerQuality {
             Self::Degraded => 2,
         }
     }
-
-    /// Return whether this quality still permits the normal preferred path.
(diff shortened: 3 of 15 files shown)


RyanKung:
## Review — #641 prioritize DHT connections by peer quality (closes #639)

Focused on the load-bearing claim: that quality only reorders connection attempts and never perturbs Chord. Verified — that part is correct.

### Verified correct (DHT semantics preserved)
- **Quality ordering is a stable permutation.** `order_peers_by_quality` sorts on `(connection_rank, original_index)`, so within equal quality the input (Chord-produced) order is preserved. Test `order_peers_by_quality_is_stable_permutation` confirms.
- **Ordering is confined to connection effects.** `run_prioritized_dht_effects` partitions effects, quality-orders only the `ConnectDhtPeer` peers, and runs every other effect (storage, successor, notify, …) unchanged. Chord's own distance ordering (`topology::…sort_by_key(dist)`) is untouched. So ownership / successor / placement semantics don't move — matches the PR's claim.
- **Observation recording is bounded.** Disconnect is one `Disconnected` per connection epoch (`measured_disconnects: BTreeSet<Did>` + test `disconnected_observation_is_once_per_connection_epoch`); send/recv observations are recorded only at the delivery-future / callback boundaries; the observation cache is TTL- and eviction-bounded.
- **Classification is sound** const-fn logic (failure threshold → `Degraded`, positive observation → `Healthy`, else `Unknown`), and the version bump to `0.12.0` is consistent across `[workspace.package]` and all six `[workspace.dependencies]`.

### Findings (minor)
1. **Half the quality model is unwired in production.** `connection_rank` is live (drives ordering), but `permits_preferred_path` is consumed only by `BehaviourJudgement::good()`, and `good()` has **no production callers** (empty on this branch and on `master`). So this PR adds `permits_preferred_path` as new surface that nothing in production uses — peer quality currently *orders* but *gates* nothing. Either wire quality into an actual decision (connection acceptance / preferred-path selection) or drop the unused boolean-judgement half. Related: the PR silently broadens `good()`'s meaning from disconnect-only to "not Degraded" (now also send/receive failures); harmless while `good()` is uncalled, but a latent behavior change the moment something calls it.
2. **`classify` test misses one failure dimension.** `peer_quality_evidence_classifies_unknown_healthy_and_degraded` exercises `Degraded` via `disconnected` and via `failed_to_send`, but never drives `failed_to_receive` to its threshold, so the `failed_to_receive >= limit` branch of `reaches_failure_limit` is untested. Add a case.

No correctness defects in the DHT path. Acceptance is your call.

RyanKung:
## Re-review — `833c261` (peer quality review feedback)

Both findings resolved, and resolved well.

- **Unwired/coupled quality model (finding 1):** `permits_preferred_path` is fully removed, and `good()` is now an abstract method **decoupled** from `quality()`, with explicit docs stating `quality` is advisory and "does not gate Chord membership, routing, ownership, or storage placement," and that collapsing the three-valued quality into a boolean would turn advisory scheduling into a hidden gate. `PeriodicMeasure::good()` is implemented independently from the legacy `ConnectBehaviour`/`MessageSendBehaviour`/`MessageRecvBehaviour` threshold judgements (AND-composed), so it keeps its legacy boolean meaning rather than being silently redefined as `!Degraded`. This is the right call, not just a mechanical fix. Verified both `BehaviourJudgement` impls (`PeriodicMeasure`, the test `RecordingMeasure`) now supply `good()`, and there are no remaining `permits_preferred_path` references.
- **Missing failure-dimension test (finding 2):** the `failed_to_receive >= limit → Degraded` branch is now covered (`PeerQualityEvidence::new(1,0,0,0,0,10) → Degraded`), so all three failure dimensions are independently exercised. The now-invalid `assert!(!good(...))` (which assumed `good` derives from `quality`) was correctly dropped.

No new issues; the DHT-safety guarantees from the prior round still hold. Acceptance is your call.

PR #642, opened 2026-06-28
Pin CodeQL analysis languages

RyanKung:
## Summary

- Add a checked-in CodeQL advanced setup workflow.
- Analyze only the languages that have source in this repository: Rust, GitHub Actions, and JavaScript/TypeScript.
- Omit C/C++ analysis explicitly.

## Root Cause

The failing master job was the generated/dynamic CodeQL configuration running `Analyze (c-cpp)`. It inferred C/C++ from `crates/node/include/rings.h`, but the repository has no C/C++ source files. CodeQL reported `0 source files, 1 header files` and failed extraction with `No source files found`.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/codeql-analysis.yml"); puts "yaml ok"'`
- `actionlint .github/workflows/codeql-analysis.yml`
- `git diff --cached --check` before commit

## Operational Note

This is the repository-side fix. If GitHub CodeQL default setup remains enabled in repository settings, the repo should be switched to advanced setup so this checked-in workflow is authoritative.


RyanKung:
Closing this PR. The checked-in CodeQL advanced workflow is the wrong fix while repository CodeQL default setup remains enabled: GitHub rejects CodeQL Action SARIF uploads with `CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled`.\n\nCurrent repository default setup already excludes `c-cpp` and the latest master default CodeQL run is green for actions, javascript-typescript, and rust. The earlier `Analyze (c-cpp)` failure was from a stale/default setup validation run, not something that can be repaired by adding an advanced workflow in this PR.

PR #644, opened 2026-07-03
Add Rings browser frontend and extension workflow

RyanKung:
## Summary

This PR adds `examples/frontend` as the Rings browser frontend. It replaces the historical browser connectivity example with a topology-first Rust/Yew app that can run as both a web page and a Chrome MV3 extension.

The standalone dweb and proof-demo surfaces are not removed or superseded by this PR. The frontend includes dweb, proof, custom-message, workbench, and peer-connection panels so a browser node can be operated from one screen, but it only replaces the old browser example surface.

It also adds extension packaging and wallet bridge support, improves SDP and HTTP peer connection flows, caches frontend settings, refreshes peer state automatically, and refactors the frontend into smaller modules after the UI work stabilized.

The PR also includes the supporting native/container pieces used by the frontend workflow:

- `docker/cluster/` for launching a local multi-node Rings cluster with deterministic ports and optional supplied private keys.
- `rings new-session --key-file` so cluster startup can avoid passing private keys on the command line.
- Browser dependency updates needed by the wasm/Yew and extension path, including the wasm-bindgen update and wasmer feature cleanup.
- TypeScript package-build updates for the generated wasm-bindgen declarations (`typescript` 5.9, `tsc -p tsconfig.json`, and explicit disposable-symbol libs).
- Small core CI fixes found while validating this PR: a clippy `vec!` lint in a DHT SEC test and an executable macro-expansion doctest for wasm docs.

## Review Fixes

The follow-up commits address review feedback by:

- preserving local disconnect cleanup status with a live node-generation guard.
- persisting extension seed-connect peer snapshots and stop cleanup messages so status polling does not discard richer state.
- guarding local and extension async snapshot/settle responses with generation tokens, including extension stop, refresh, and start-polling paths.
- centralizing connection actions behind a `NodeBackend` boundary instead of duplicating extension/local logic in every SDP/HTTP callback.
- removing peer `"unknown"` sentinels and dropping malformed peer rows at the boundary.
- replacing positional settings persistence with a named snapshot struct.
- moving extension bridge/service-worker JavaScript out of template strings into real source assets under `examples/frontend/extension-assets/`.
- renaming the browser app package from node demo to frontend while migrating previously saved local settings from the old keys.
- encoding strict frontend clippy lints in `examples/frontend/Cargo.toml` and QACI.
- making the wallet pure-function tests execute under the wasm browser runner.
- fixing the docker entrypoint unused node directory and clarifying comments-only key-file errors.

## User Impact

- Browser users get a topology-focused frontend for local node setup, peer connection, SDP exchange, workbench actions, and WebRTC/debug shortcuts.
- Extension users can load the generated MV3 package from `examples/frontend/dist-extension` and use the same interface backed by the extension bridge.
- Existing frontend settings saved under the earlier node-demo localStorage keys are migrated to the new `rings.frontend.*` keys on read.
- The topology view presents local and inferred Chord relationships with clearer visual hierarchy and animated data flow.
- Local cluster testing can use the new Docker image and `--key-file` path without exposing keys in process arguments.

## Validation

- `git diff --check`
- `rustup run 1.94-aarch64-apple-darwin cargo clippy --manifest-path examples/frontend/Cargo.toml --target wasm32-unknown-unknown --all-targets -- -D warnings`
- `CHROMEDRIVER=/usr/local/bin/chromedriver WASM_BINDGEN_USE_BROWSER=1 rustup run 1.94-aarch64-apple-darwin cargo test --manifest-path examples/frontend/Cargo.toml --target wasm32-unknown-unknown`
- `NO_COLOR=false RUSTUP_TOOLCHAIN=1.94-aarch64-apple-darwin trunk build --release` from `examples/frontend`
- `npm run test:frontend-extension-wallet`
- `node --check` for `examples/frontend/scripts/package-extension.mjs`, source extension assets, and generated `examples/frontend/dist-extension` assets
- `rustup run 1.94-aarch64-apple-darwin cargo test --doc`
- `rustup run 1.94-aarch64-apple-darwin cargo test -p rings-core --doc --no-default-features --features wasm`
- `NO_COLOR=false RUSTUP_TOOLCHAIN=1.94-aarch64-apple-darwin npm pack --dry-run`
- `RUSTUP_TOOLCHAIN=1.94-aarch64-apple-darwin cargo clippy --all --tests -- -D warnings`
- `RUSTUP_TOOLCHAIN=1.94-aarch64-apple-darwin cargo clippy -p rings-core --features wasm --no-default-features --target wasm32-unknown-unknown --tests -- -D warnings`
- `RUSTUP_TOOLCHAIN=1.94-aarch64-apple-darwin cargo clippy -p rings-node --features browser_default --no-deps --no-default-features --target=wasm32-unknown-unknown --tests -- -D warnings`
- `CHROMEDRIVER=/usr/local/bin/chromedriver WASM_BINDGEN_USE_BROWSER=1 RUSTUP_TOOLCHAIN=1.94-aarch64-apple-darwin cargo test -p rings-core --release --target=wasm32-unknown-unknown --features wasm --no-default-features --doc`
- `bash -n docker/cluster/rings-cluster-entrypoint.sh`
- `rustfmt --check crates/node/bin/rings.rs`


RyanKung:
Reviewed at e894a67 with the strict rubric. Verified locally: `cargo clippy --target wasm32-unknown-unknown --all-targets` with `-D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing` passes on `examples/node-demo`.

## Correctness

1. **The disconnect cleanup message is unreachable.** `examples/node-demo/src/app.rs:435` — `UseStateHandle::deref` reads the render-time snapshot, not live state. The callback was created in a render where the node was online, so `(*did)` is the old non-empty DID; `did.set(String::new())` earlier in the callback does not change what the captured handle derefs to. `if (*did).is_empty()` is therefore false on every reachable path, and the cleanup outcome ("closed N peer links" / "peer cleanup timed out" / "peer cleanup failed") is never shown. If the intent is "only report when the node was not restarted meanwhile", the guard needs a live source (e.g. a `use_mut_ref` generation counter), not a state-handle snapshot.

2. **The seed-start peer merge is computed and discarded in the extension path.** `examples/node-demo/src/extension.rs:308` binds the settled snapshot to `_snapshot` and drops it. The panel only sees peers through `rings.node.status`, and `headless_node_snapshot(state, ..., None, false)` (`extension.rs:184`) never merges the seed peer nor settles. So the `merge_required_peer` work in the start path can reach no observer; its only effect is a hidden ~7s delay (`[0, 1_000, 2_000, 4_000]`) before `starting` flips false. The webpage path (`app.rs:336`, via `sync_peers_after_handshake`) pushes the merged view into the UI — the two paths have already diverged. Either store the settled peers where status polls can serve them, or delete the dead call.

3. **`stop_headless_node` regresses its own message.** `extension.rs:329` writes `"background node stopped"` into `state.message` before cleanup; the richer outcome (closed count / timeout) exists only in the returned snapshot. The next status poll reads `state.message` and reverts to the generic text.

## Architecture / repetition

4. **The extension-vs-local boundary is re-derived at every call site.** `extension::extension_node_bridge()` is checked in seven-plus places (`app.rs` on_start / on_disconnect / refresh interval; `connect.rs` on_http_connect / on_create_offer / on_answer_offer / on_accept_answer), and each callback duplicates input validation and success handling in both arms (e.g. the `remote_did.trim().to_string()` + empty-check block appears verbatim twice inside `connect.rs:129-176` alone). The backend distinction deserves a type resolved once — `enum NodeBackend { Bridge(JsValue), Local(DemoNode) }` with `connect_http` / `create_offer` / `answer_offer` / `accept_answer` / `status` as methods — after which every callback has one arm and one validation.

5. **The node startup pipeline has two sources.** `app.rs:241-346` (webpage on_start) and `extension.rs:244-321` (`start_headless_node_inner`) are the same sequence — parse settings → wallet connect under `WALLET_CONNECT_TIMEOUT` → session auth under `SESSION_AUTH_TIMEOUT` → insert `/` page → `dweb::register` → `custom::register` ×2 → optional seed connect + required-peer merge — differing only in where effects land. The drift predicted by this duplication is already observable (finding 2). One start routine parameterized by an observer/effect sink gives it one source.

6. **The JS-interop vocabulary has three sources.** `browser_api.rs` defines `js_prop` / `js_string_field` / `js_set` / `js_method` / `await_js` / `is_callable` / `js_error_label`; `wallet.rs:105-170` re-implements all of them privately (`prop` / `string_prop` / `set` / `method` / `await_js` / `is_callable` / `js_err`, plus `call0..call3`); `controls.rs:845` carries a third `js_error_label` and `navigator_user_agent` (`controls.rs:836`) re-does raw `Reflect` instead of `js_string_field`. Likewise two hex decoders exist: `wallet.rs:183-214` (`hex_to_bytes` / `hex_nibble`) and `topology.rs:369-405` (`did_identifier` / `hex_nibble`). Each concept needs one home.

7. **Constants with multiple sources.** The demo namespace list `["custom", "example"]` appears at `app.rs:98` (default registered list), `app.rs:311`, and `extension.rs:281` — if they drift, the "Registered:" UI lies. The settle schedule `[0, 1_000, 2_000, 4_000]` appears at `peer_sync.rs:23` and `extension.rs:500-504`. The 240×750ms start-poll budget appears at `extension.rs:640` and again in the generated `node_bridge.js` (`watchNodeIconUntilSettled`).

8. `InferredEdge` and `InferredFinger` carry identical `endpoints` implementations (`topology.rs:220-230`); the finger is an edge plus an exponent.

## Model / types

9. **The sentinel survives where an `Option` was erased.** `node.rs:198-199` maps missing JSON fields to `"unknown"` / `"Unknown"`, then `disconnect_all` (`node.rs:158`) re-derives the lost distinction by string compare `peer.did == "unknown"`. `parse_peer_views` (`extension.rs:838-839`) repeats the same defaults. Keep the distinction in the type (fail the row, or `Option<Did>`), and if a sentinel must exist, name the predicate on `PeerView` instead of comparing naked strings at call sites.

10. **Errors are stringly throughout the crate.** Every fallible function returns `Result<_, String>`, and data rides inside format strings (`node.rs:172` `"closed {closed}/{attempted} peer links"`). The status line is a legitimate display boundary, but the conversion belongs at that boundary; as written, callers cannot distinguish timeout from rejection from transport failure without parsing prose.

11. **Dead stringly branch in `metric`.** `controls.rs:593` branches on `label == "Local DID"`; no caller passes that label (`local_did_metric` at `controls.rs:606` owns that case). The branch is dead and the dispatch is stringly.

12. **Undocumented lint suppression.** `node.rs:62` `#[allow(clippy::arc_with_non_send_sync)]` carries no justification. If the `Provider` API forces `Arc` in single-threaded wasm, state that in a comment at the smallest scope; otherwise fix the ownership shape.

13. **`WalletAccount` derives `Clone` while holding the signing capability.** `wallet.rs:30-39` — the `handle` is, in the WebCrypto case, the private `CryptoKey`. The duplication law (JS reference sharing; no key material copied; every clone can sign) is real but undocumented.

14. **Settings persistence is positional.** `app.rs:112-131` pairs a 7-tuple with `.0`–`.6` writes against seven key constants; the alignment is maintained by hand. A named struct (or iterating `(key, handle)` pairs) removes the transposition hazard.

## Tests / CI

15. **The wallet unit tests never execute anywhere.** `wallet.rs:484-505` (`mod tests`, plain `#[test]`): the native build of this crate fails (`rings-node` browser feature does not compile for the host target — verified locally), and the wasm runner (`wasm-bindgen-test-runner`, per `.cargo/config.toml`) only executes `#[wasm_bindgen_test]`. So `hex_signature_parser_accepts_prefixed_even_hex` and — critically — `rings_prefix_matches_core_secp256r1_transcript`, the compatibility witness for WebCrypto session signing, are dead in every pipeline. `topology.rs:639-640` already shows the correct dual `cfg_attr` pattern; `wallet.rs` instead duplicated one test into `wasm_tests` and left the rest stranded.

16. **The strict lint set is not encoded for the new workspace.** `examples/node-demo` is its own workspace, so the root `[workspace.lints]` (unwrap/expect/panic/todo/unimplemented/indexing_slicing, all `deny`) does not reach it, its `Cargo.toml` has no `[lints]` table, and qaci.yml runs only `cargo check` + `cargo test` for it. The PR's validation ran the flags by hand; nothing keeps them true. Add the `[lints]` tables to `examples/node-demo/Cargo.toml` (and preferably a clippy step in CI).

## Size

17. **Functions exceed the 100-line bound.** `app()` is ~550 lines (`app.rs:37-586`) and owns all state wiring, settings persistence, bridge polling, startup, disconnect, and assembly for every panel; `link_control` is ~282 (`connect.rs:66-348`); `control_sidebar` ~200 (`controls.rs:167-368`); `topology_component` ~173 (`topology.rs:25-198`). Findings 4 and 5 shrink `app()` and `link_control` mostly for free; the rest decompose along custom hooks / child components.

## Scope / packaging

18. **Undescribed load-bearing changes ride this PR.** The PR body covers the demo and extension workflow, but the diff also adds `docker/cluster/` (469 lines incl. a 362-line entrypoint), a new `rings --key-file` flag (`crates/node/bin/rings.rs:270-335`), and the wasmer 5→7 / wasm-bindgen 0.2.100→0.2.121 upgrades with `wasmer` removed from `rings-node` features. These are real behavior/dependency changes; describe them in the PR body at minimum, ideally split them out. (Nits in the entrypoint: `node_dir` is created at line 302 and never used; `read_secret_key_file` reports "empty" for a comments-only file.)

19. **~400 lines of generated JS live inside template strings.** `scripts/package-extension.mjs` embeds `service_worker.js`, `wallet_bridge.js`, and `node_bridge.js` as string literals; `node --check` validates only the generator, so nothing lints or parses the shipped code. Ship them as real `.js` assets copied at package time so tooling sees them.

---

Positive notes for the record: the pure Chord inference core in `topology.rs` is cleanly separated from rendering and carries genuine model tests (wrap-around, low-160-bit truncation, N-bounded finger targets); `dweb.rs` keeps the handler pure-shaped (state in, effects out); the empty-`sandbox` iframe and the URL-safe base64 handling carry over the earlier review outcomes correctly.

RyanKung:
Re-reviewed at 6cc9bfbe (fresh read of the new commit, not a closure check). Verified locally: `cargo clippy --target wasm32-unknown-unknown --all-targets -- -D warnings` passes with the new `[lints.clippy]` table active; the three extension assets parse (`node --check`) and match the previously embedded templates (node_bridge additionally names the 240×750ms constants). The PR body now describes docker/cluster, `--key-file`, and the dependency updates.

## Resolved from round 1

The generation guard replaces the snapshot-deref check (`app.rs:432-436`, `465`) and is correct: the counter is a live `RefCell`, incremented at both Start and Stop entry. The seed-settle snapshot is no longer discarded — `headless_node_snapshot` now persists peers/message into `HeadlessNodeState` (`extension.rs:534-538`) and the start path returns the settled message (`extension.rs:313-321`). `stop_headless_node` stores its final message (`extension.rs:364`). The four link operations now dispatch through `NodeBackend`/`PeerUpdate` with `required_input` as the single validation point (`connect.rs:67-173`) — and the extension arm now surfaces `snapshot.error`, which the old inline code dropped. The `"unknown"` sentinel is gone (`PeerView::from_fields`), `DEMO_NAMESPACES` / `PEER_SETTLE_DELAYS_MS` have one source, `InferredFinger` embeds `InferredEdge`, the JS-interop vocabulary is unified in `browser_api.rs`, `hex.rs` owns nibble parsing, the `allow` and the `Clone` law are documented, settings persistence is a named struct, the wallet tests are dual-target (the transcript witness now runs in the wasm pipeline), the dead `metric` branch is gone, and the strict lints are encoded in Cargo.toml plus a CI clippy step.

## New findings in the new code

1. **The snapshot write-back races with stop.** `headless_node_snapshot` writes `state.peers` / `state.message` after its awaits (`extension.rs:534-538`). A settle snapshot (connectHttp / acceptAnswer / seed start, up to ~7s in flight) that overlaps `stop_headless_node` will repopulate the state that stop just reset: `state.message` becomes "… connected; N peers visible" and `state.peers` non-empty while `state.node` is `None`. The offline branch then serves that stale message to the next status call. The same write-back also makes plain status polls mutate state — every poll overwrites `state.message` with the generic context, clobbering progressive start messages for the offline-branch reader. The query stopped being a query. Guard the write-back on the node still being installed (re-check `state.node` identity after the awaits, or reuse the generation-counter idea from app.rs), and skip the message write on non-settle polls.
   Invariant worth stating in the code: `state.node.is_none() ⇒ state.peers.is_empty() ∧ state.message reflects the stopped/offline outcome`.

2. **The generation guard stops at the disconnect message; the start pipeline never consults it.** Sequence: Start → Stop while "authorizing session key" (Stop is visible because `node_starting` is true) → Stop reports "node already offline" → the in-flight start continues, commits `*node_ref.borrow_mut() = Some(built)` and `did.set(...)` (`app.rs:341-343`), and the node comes online after the user was told it is offline. The mechanism to fix this now exists: capture the generation at Start entry and, before committing the built node, bail (and `built.stop()`) if it is stale.

3. **`PeerView::from_fields` is not the sole authority for its own invariant.** Callers still build seed peers by struct literal (`connect.rs:100-103`, `extension.rs:308-311`), and `merge_required_peer` re-implements the emptiness predicate inline (`peer_sync.rs:45`). Either route construction through the constructor (fields `pub` invite bypass) or name the predicate once on `PeerView` and use it in both places.

## Still open from round 1

- **The startup pipeline remains two sources.** `NodeBackend` covers the four link operations, but `app.rs` still branches on `extension_node_bridge()` inline at four sites (mount effect, on_start, on_disconnect, refresh interval), and the local protocol-install sequence (settings → wallet timeout → session auth → site insert → dweb register → namespaces → seed connect) still exists in both `app.rs:264-369` and `extension.rs:248-321`. Finding 1 above is this duplication's newest drift. Extending the same boundary to start/stop/status would collapse most of it.
- **Function sizes.** `app()` is now 568 lines (it grew), `link_control` 208, `control_sidebar` 202, `topology_component` 174 — all above the 100-line bound. The startup-boundary work above is also the main path to shrinking `app()`.
- **Errors remain `Result<_, String>` throughout**; variants (timeout vs rejection vs transport) are still only distinguishable by parsing prose.

## Nits

- `hex_to_bytes` maps the identical error string four times (`wallet.rs:136-152`); one `map_err` at the collect boundary suffices.
- `InferredFinger` grew a `source()` accessor but no `target()`; the test reaches `link.edge.target` directly (`topology.rs:730-731`) — pick one access style.
- `connect_headless_node_http` still re-validates the endpoint (`extension.rs:371-374`) that `required_input` already validated panel-side; as defense at the message boundary that is defensible, but then the offer/answer handlers' inline checks and the panel checks are the same proposition in two dialects — worth one shared source if kept.

RyanKung:
Re-reviewed at fae6d2dd (fresh read). Verified locally: strict clippy (`[lints.clippy]` table active, `-D warnings`, wasm32 target, all targets) passes.

## Resolved from round 2

- **Headless write-back race**: `HeadlessNodeState.generation` exists; `begin_headless_start` / `stop_headless_node` bump it; the snapshot write-back is guarded by `state.generation == generation && state.node.is_some()` and writes `message` only for settle snapshots (`extension.rs:669-677`); stop's late message write is generation-guarded (`extension.rs:427-432`); `headless_demo_node` returns a `HeadlessDemoNode` token carrying the generation, and the value-returning ops re-check it after their awaits (`ensure_headless_generation_current`). The invariant now holds.
- **Start pipeline cancellation**: `on_start` checkpoints the generation at every await boundary and `built.stop()`s on staleness (`app.rs:289-372`); Stop now bumps the generation even when `node_ref` is `None`, so Stop-during-authorization cancels the in-flight start. The same checkpoints exist headless-side. Bonus fix acknowledged: dweb/custom registration failures now `built.stop()` in both pipelines — that was a latent listen-loop leak neither review round had named.
- **`PeerView` sole authority**: fields are private; `from_fields` / `connected` / `did()` / `state()` / `is_addressable()` are the only access paths; all former struct-literal sites and `merge_required_peer` now go through them; tests construct via `connected`.
- Nits closed: `hex_to_bytes` maps one error at the collect boundary via `Result<_, ()>`; `InferredFinger::target()` added and used consistently.

## New findings

1. **The race just fixed headless-side still exists in the webpage path — the panel-side settle writers are unguarded.** `sync_peers_after_handshake` (`peer_sync.rs:12-41`) writes the `peers` and `status` UI handles across a ~7s settle window with no ownership check. Callers: the seed path in `app.rs:390-398` (generation checked *before* the call, never during) and `PeerUpdate::apply`'s Local arm (`connect.rs:141-150`). Scenario: Connect endpoint → Stop within the window → disconnect clears peers and sets "node disconnected" → the next settle sample calls `list_peers` on the stopped provider, `merge_required_peer` re-inserts the seed peer, and the UI shows a connected peer plus "HTTP endpoint connected; 1 peer visible" while the node is offline. Mirror the headless design: pass the generation (or a `still_current: impl Fn() -> bool`) into `sync_peers_after_handshake` and check before each write. The `Snapshot` arm's single post-await `peers.set` and the 180s `poll_extension_node_start` loop have the same shape with smaller/benign windows — one token type would cover all of them.

2. **The fix landed twice in parallel dialects; the boilerplate now measures the unpaid restructure.** `*node_generation.borrow() == start_generation` appears at ~8 checkpoints inside `on_start` (14 `node_generation.borrow` sites in app.rs), and the identical checkpoint sequence exists again headless-side through three overlapping mechanisms — `headless_generation_current`, `ensure_headless_generation_current`, and `set_headless_starting_for_generation`, whose `bool` return no caller reads (`start_headless_node_inner` checks `headless_generation_current` separately right before invoking it). This is the third round in which the start pipeline's two sources (`app.rs` local arm vs `extension.rs:286-397`) forced the same change to be written twice. One start routine with one cancellation-token type (`still_current()` / `abort(built)` / `commit()`) would have made rounds 2 and 3 single-site fixes. Related size drift: `app()` is now 605 lines, and `start_headless_node_inner` (112) and `headless_node_snapshot` (108) have crossed the 100-line bound that `link_control` (208) and `control_sidebar` (202) already exceed.

## Still open

- Errors remain `Result<_, String>` throughout; failure variants are distinguishable only by prose.

## Nits

- Stop during an in-flight start reports "node already offline" (`app.rs:465-469`) while it is actually cancelling a start; "start cancelled" would be truthful — `node_starting` is in scope to pick the message.
- `headless_node_snapshot`'s stale-generation branch (`extension.rs:601-625`) duplicates the offline-default-message logic of the `node = None` branch directly below it.
- `connect_headless_node_http` still re-validates the endpoint the panel already validated via `required_input` — same proposition in two dialects (carried over from round 2).

RyanKung:
Re-reviewed at aec4bacc (fresh read). Verified locally: strict clippy (wasm32, all targets, `-D warnings` with the `[lints.clippy]` table) passes.

## Resolved from round 3

- **Panel-side settle writers are guarded.** `generation.rs` names the mechanism (`GenerationClock` / `GenerationToken`, with real doc contracts — "proof that an async operation still owns the generation it started in" is exactly the right framing). `sync_peers_after_handshake` takes `still_current` and checks it around every suspension point; `NodeBackend::Local` / `PeerUpdate::Local` carry the token; the interval refresh guards its write; the seed path threads `start_token`.
- Nits closed: Stop during a start now reports "node start cancelled" (the `was_starting` read is a synchronous render-snapshot read at click time, which is correct — unlike the round-1 pattern, nothing mutates it earlier in the same callback); `retained_headless_message` and `required_message_field` deduplicate the offline-default and validation logic; the unused `bool` return is gone.

## The race has one instance left: extension mode, end-to-end

Three facts compose into the same stale-write, webpage-fixed but extension-reachable:

1. `headless_node_snapshot`'s final generation re-check gates only the **write-back**; the **returned snapshot** is built unconditionally with `online=true` and the pre-stop peers/message (`extension.rs:669-687`). A stop landing inside the ~7s settle window produces a response that claims an online node with connected peers.
2. The panel's `PeerUpdate::Snapshot` arm applies that response unguarded — `peers.set(snapshot.peers); status.set(...)` (`connect.rs:160-164`); only the `Local` arm carries a token.
3. The extension-mode disconnect arm never bumps the panel `GenerationClock` (`app.rs:424-457` returns before the `bump()`), so even a token in the `Snapshot` arm would currently pass.

Repro: extension mode → Connect endpoint → Stop within the settle window → the connect response arrives → panel shows the seed peer and "HTTP endpoint connected; N peers visible" while the background node is stopped. The cleanest single-authority fix is headless-side: after the settle loop, when `state.generation != generation` (or `state.node` is gone), return the retained offline snapshot instead of the computed one — the write-back guard already computes exactly this condition. Bumping the panel clock in the extension disconnect arm and tokenizing the `Snapshot` arm would additionally protect the panel against any future bridge that lies.

## The mechanism still has two dialects

`generation.rs` exists now, but only the panel uses it. The headless side still runs its own raw `u64` plus three bespoke helpers (`begin_headless_start`, `headless_generation_current`, `ensure_headless_generation_current`, `set_headless_starting_for_generation`). `HeadlessNodeState.generation` could be a `GenerationClock` (the cell just moves out of the struct), the `HeadlessDemoNode` handle a `GenerationToken`, and the helpers collapse into `is_current()`. The finding above is, once again, drift between the two dialects — the gated write-back was implemented against one representation while the panel's guard was implemented against the other, and the seam between them (the returned snapshot) is where the hole is.

## Still open

- The startup pipeline remains two sources (`app.rs` local arm vs `extension.rs:286-397`); `app()` is now 610 lines, `link_control` 213, `start_headless_node_inner` 112 (`headless_node_snapshot` dropped back to 96).
- Errors remain `Result<_, String>` throughout.

## Nits

- `app.rs` still holds both the raw cell and the clock (`node_generation = use_mut_ref(0u64)` feeding `GenerationClock::new(...)`, `app.rs:59-60`); nothing else reads the raw cell, so the clock could own its `Rc` (`GenerationClock::default()`) and the representation would stop leaking.
- `sync_peers_after_handshake` re-checks `still_current` at points with no intervening await (entry → `status.set` → re-check → `peers.set`); the discipline that matters is one check per resumption, before the writes that follow it — the extra checks blur where the suspension points actually are.
- Two more drive-bys rode this commit unannounced: the TypeScript 4.9→5.9 / `tsc -p` / `tsconfig.json` rework, and a `vec!`→array lint fix in `crates/core/src/dht/entry.rs` (a core test file in an examples PR). Both look harmless; both belong in the PR body's supporting-changes list.

RyanKung:
Re-reviewed at da7266e3 (fresh read). Verified locally: strict clippy (wasm32, all targets, `-D warnings`, `[lints.clippy]` table) passes.

## Resolved from round 4

- **The returned snapshot is now gated, not just the write-back.** After the settle loop, a changed generation or missing node returns `retained_live_headless_snapshot_js(&state)` under the same borrow that would have guarded the write (`extension.rs:662-665`) — single authority, as the condition and the response now come from one read. `retained_headless_snapshot_js` also zeroes peers/account when offline, so the response itself carries the `offline ⇒ no peers` invariant rather than trusting callers.
- **The `Snapshot` arm and the extension stop/refresh paths hold tokens.** `NodeBackend::Extension` and `PeerUpdate::Snapshot` carry a `GenerationToken`; `apply` checks it (`connect.rs:174-179`); the extension disconnect arm bumps the clock and guards both match arms (`app.rs:426`, `442-460`); the extension interval refresh checks its token after the await (`app.rs:524-532`). All three facts from round 4's composition are individually closed.
- **`sync_peers_after_handshake` now has exactly one check per resumption point** — entry, post-sleep, post-`list_peers` before the writes, and the `Err` arm. The suspension points are legible again.

## Residual, bounded

The last unguarded panel writers are the start-polling pair: `apply_extension_snapshot` and the `poll_extension_node_start` loop (`extension.rs`), which `on_start`'s bridge arm drives without a token for up to 180s. A Stop during extension-mode start can let one pre-stop snapshot repaint did/peers for ≤1 poll tick (750ms) before the next status — now served from retained headless state — corrects it; the loop's exit `Err(last_message)` then rewrites `status` with the same stop message text. Content converges because headless is now the single authority, so this is transient repaint, not persistent wrong state. For symmetry (and against any future bridge that lies), threading the existing `start_token` into the poll would finish the family; it is the one writer the token design hasn't reached.

## Still open (structural, unchanged for three rounds)

- The startup pipeline has two sources (`app.rs` local arm vs `extension.rs:286-397`), and the cancellation mechanism still has two dialects: `generation.rs` panel-side vs the raw `u64` + `begin_headless_start` / `headless_generation_current` / `ensure_headless_generation_current` / `set_headless_starting_for_generation` headless-side. `HeadlessNodeState.generation` could be a `GenerationClock` and the four helpers collapse into `bump()` / `is_current()`. Rounds 3, 4, and 5 each fixed one drift between the dialects; the seam keeps manufacturing them.
- `app()` is 617 lines; `link_control` 213; `start_headless_node_inner` 112.
- Errors remain `Result<_, String>` throughout.
- `app.rs:59-60` still holds both the raw cell and the clock built from it.

## Nits

- `crates/core/src/utils.rs:48` switches a doctest to `rust,ignore` — a core-crate drive-by that weakens a doc gate. If the illustrated expansion no longer compiles, updating the example to the macro's current output keeps the gate; `ignore` lets the docs drift silently. Also still unannounced in the PR body's supporting-changes list (as with the TS bump and `entry.rs` last round).

RyanKung:
Re-reviewed at 1a334b9b (fresh read). Verified locally: strict clippy passes; additionally `cargo test --doc -p rings-core --no-default-features --features wasm` compiles the rewritten doctest (`utils.rs - js_func::of - compile ... ok`).

## Resolved from round 5

- **The start-polling pair is guarded.** `poll_extension_node_start` takes the `GenerationToken`, checks it after the sleep, treats a stale status error as clean exit, and stops when `apply_extension_snapshot` reports staleness; `apply_extension_snapshot` itself checks the token and returns whether it wrote (`extension.rs:807-880`); both `on_start` bridge-arm error paths are token-guarded (`app.rs:236-269`).
- **The clock owns its cell.** `GenerationClock::default()` + `use_mut_ref(GenerationClock::default)`; `new(Rc)` is gone, the raw `u64` cell no longer leaks into `app()`.
- **The doctest drive-by became a real fix.** Instead of `ignore`, the `js_func::of` example is now `no_run` and actually compiles — and the rewrite fixed genuine bugs in the old illustration (missing `?`, `|_|` closures referencing an unbound `e`, wrong error conversions). Verified above under the wasm feature, which is the only cfg where the module exists.

## One writer left in the family

The mount effect (`app.rs:160-177`) applies the initial background-node snapshot — `did` / `peers` / `wallet_account` / `status` — with no token. A Start clicked while the mount status roundtrip is in flight gets its "connecting …" state overwritten by the pre-start snapshot when the response lands. The window is page-load-instant and the next poll self-corrects, so this is the same transient class as round 5's residual, smaller. It is also the literal last unguarded async writer in the crate: `generation.token()` captured at effect creation plus one `is_current()` before the writes finishes the family completely.

## Still open (structural; unchanged since round 3)

- **Two pipelines, two dialects.** The startup sequence still exists in both `app.rs` (local arm) and `extension.rs:286-397`, and the headless side still runs its raw `u64` + four bespoke helpers instead of the `GenerationClock`/`GenerationToken` it now shares a crate with. Every race fixed in rounds 3–6 was a drift between these two copies; the copies remain.
- `app()` is now 625 lines and has grown in every round of this PR (586 → 605 → 610 → 617 → 625); `link_control` 213, `start_headless_node_inner` 112.
- Errors remain `Result<_, String>` throughout.

## Note

The token/guard work across rounds 3–6 has converged to a coherent design: one clock, tokens as ownership proofs, retained snapshots as the single authority for stale reads. The two structural items above are what keeps requiring each fix to be written twice; they are refactors, not bug fixes, and where to draw the line for this PR versus a follow-up is your call.

RyanKung:
Re-reviewed at 0c9f5980 (fresh read of the rename commit). Strict clippy passes on the renamed crate; the `[lints.clippy]` table survived the rename.

## Blockers

1. **The rename breaks CI.** `.github/workflows/qaci.yml:68,72,78` still say `working-directory: ./examples/node-demo`; the directory no longer exists, so the frontend check, clippy, and browser-test steps will fail on checkout. The commit renamed the crate, package.json script, READMEs, and test fixtures but not the workflow.

2. **`.gitignore` still un-ignores the old path.** Lines 8–9 read `!examples/node-demo/extension-assets/` — dead rules now. The global `**/*.js` ignore therefore swallows anything new under `examples/frontend/extension-assets/`. The three tracked bridge files stay tracked, so this is silent until the next time someone adds or regenerates a bridge asset and it quietly never shows up in `git status`.

3. **The stated scope and the tree disagree.** The intent of this change is that `dweb` and `proof-demo` are retained and only the historical `browser` example is replaced — and the new frontend README asserts exactly that ("The standalone `dweb` and `proof-demo` surfaces remain conceptually separate", "alongside the retained proof demo"). But the branch still deletes both: vs master this PR removes `examples/dweb` and `examples/proof-demo` in full (12,627 deletions), and `git ls-tree` of the branch shows `examples/` containing only `ffi`, `frontend`, `native`, `relay`, `snark`. The README now documents directories that the same PR deletes. If retention is the intent, the restoration commit is missing — and it drags three more items with it: the qaci steps that ran the dweb/proof-demo browser tests (deleted earlier in this PR), the examples/README instructions for them, and the ecosystem-table rows for the external dweb/zkproof demos removed from the top-level README.

## Sound in this commit

- **The settings migration is the right shape.** `load_setting_with_legacy` reads the new key, falls back to the legacy `rings.node-demo.*` key, and writes forward (`extension.rs:93-99`). Because the save-effect has always persisted every setting (including the default storage name), existing users' `rings-node-demo` IndexedDB namespace carries over through the migrated `storageName` — stored DHT data survives the rename.
- **The `.stage` fallback in `package-extension.mjs` matches Trunk's post_build semantics** (dist is swapped in after the hook; the artifacts live in `dist/.stage` at hook time), and the error message follows `sourceRoot`.
- The crate rename is internally consistent: `rings-frontend`, new `main.rs`, manifest name/description, wallet-fixture storage name, `test:frontend-extension-wallet`, native/example READMEs.

## Carried forward

- The structural pair (startup pipeline in two sources; `GenerationClock` panel-side vs raw `u64` headless-side) and `app()` at 625 lines are unchanged; errors remain `Result<_, String>`.
- The PR body's Validation section still lists `examples/node-demo` paths and the old npm script name; worth refreshing together with the scope description once the dweb/proof-demo question is settled.

RyanKung:
Re-reviewed at 47dfb04e (fresh read of the three new commits). Verified locally: `examples/dweb` and `examples/proof-demo` both compile on wasm32 with the bumped pins; frontend strict clippy still passes.

## Round-7 blockers: all three closed

- **CI paths fixed and the demo gates restored.** qaci now points the three frontend steps at `./examples/frontend`, re-adds the wasm-pack install, and restores the dweb browser tests (with the original two-phase `--skip` / `--nocapture` split) and the proof-demo browser tests.
- **`.gitignore` un-ignore rules follow the rename** (`!examples/frontend/extension-assets/`).
- **The tree now matches the stated scope.** `dweb` and `proof-demo` are restored byte-identical to master in sources, HTML, webdriver.json, and READMEs; the only changes are the dependency pins aligned to the workspace (js-sys 0.3.98, wasm-bindgen 0.2.121, wasm-bindgen-test 0.3.71) with regenerated locks — and both compile. The examples table, ecosystem rows, and examples/README run instructions are back; the PR title/body were rewritten to match ("not removed or superseded" is now true).

## Ownership sweep: three functional leftovers

The link sweep covered READMEs, manifests, ROADMAP, and badges, but missed the references that do work rather than describe:

1. `cliff.toml:49` — the changelog `<REPO>` substitution still expands to `https://github.com/RingsNetwork/rings`, so every generated changelog entry will link to the old org.
2. `crates/snark/src/tests/native/test_witness.rs:81` and `crates/snark/src/tests/wasm/test_wasm_witness.rs:12` — both tests download `simple_bn256.wasm` from `RingsNetwork/rings/raw/master/...`. They pass today only through the transfer redirect, which GitHub drops the moment a new repo named `rings` appears under the old org. The native test is downloading a file that sits in its own repo (`crates/snark/src/tests/native/circoms/simple_bn256.wasm`) — reading it from disk removes the network dependency entirely; the wasm test genuinely needs a URL and should pin the new repo.

## Nit

- The crate READMEs embed the logo via `../../assets/logo/...`. That renders on GitHub, but crates.io rewrites relative URLs against the repository root without knowing the README lives in `crates/<name>/`, so the published crates' pages (rings-node, rings-core, etc. carry crates.io badges) will show a broken image. An absolute `raw.githubusercontent.com/RyanKung/rings/...` URL works in both renderers.
- Presumably deliberate, noting for the record: the Sponsor badge, the ecosystem table's external repos, the npm `--scope ringsnetwork`, and the `dev@ringsnetwork.io` authors line still reference RingsNetwork — those assets weren't transferred, so leaving them is consistent.

## Carried forward (unchanged)

The structural pair — startup pipeline in two sources, `GenerationClock` panel-side vs raw `u64` headless-side — plus `app()` at 625 lines and `Result<_, String>` errors throughout remain as previously discussed; they are refactors rather than defects, and whether they belong in this PR or a follow-up stays your call.

PR #648, opened 2026-07-03
Handle issues 645 and 647 with provider measurements

### LINKED ISSUE #645: Add a policy for Chord-routed report return
## Problem

`MessageRelay::report` currently depends on the forward-path stack in `MessageRelay.path`: the destination node reports to `path.last()` and uses `path.first()` as the original sender. That makes report delivery tied to path disclosure and prevents a policy where the response is routed normally through Chord.

For proxy and blinded-resource workflows we need a routing policy that allows response/report traffic to be addressed to a return DID/key and delivered through normal Chord routing, instead of walking the recorded forward path.

## Proposal

Add a relay/report policy that can choose between:

- path return: current behavior, reverse through `MessageRelay.path`;
- routed return: send the report as a fresh Chord-routed message to a declared return DID/key.

The first implementation can preserve the current default and add the routed return as an explicit opt-in policy.

## Scope

- Introduce an explicit report-return policy in the relay metadata or request message.
- Add a return destination field for routed reports.
- Update `send_report_message` / `MessageRelay::report` call sites so handlers do not implicitly assume path return.
- Keep normal DHT maintenance behavior compatible with the existing path-return policy unless a call site opts in.
- Add tests for both return policies.

## Security requirements

- A routed return destination must be covered by the request signature.
- Reports must bind to the original request id / transaction id to prevent replay or response confusion.
- Routed return must not allow reflection amplification to an arbitrary DID without proving control of the return destination or otherwise constraining the response size.
- Existing signed `MessagePayload` verification must remain intact.

## Non-goals

- Do not remove `MessageRelay.path` in this issue.
- Do not implement anonymous proxy semantics here.
- Do not change E2E encryption behavior.

### LINKED ISSUE #647: Add automatic online node registration and heartbeat descriptors
## Problem

Current service discovery only registers a service name to the local node DID when `register_service(name)` is explicitly called. It does not automatically publish a node descriptor when the node starts, and it does not expose enough metadata to answer basic network inventory questions such as:

- which nodes are currently online;
- which public key belongs to each online DID;
- whether the node is browser/native/ffi;
- which entries are stale because the node stopped heartbeating.

We need a first-class online-node registry instead of relying on ad-hoc service registration.

## Proposal

Add an automatic node registration/heartbeat mechanism using the DHT topic:

```text
online_nodes
```

Every node should publish a signed descriptor to this topic when it starts, and refresh it periodically while it is online.

Suggested descriptor shape:

```text
OnlineNodeDescriptor {
  did,
  public_key,
  node_type,        // browser | native | ffi
  network_id,
  capabilities,     // optional: relay, https-exit, tcp-exit, storage, snark, etc.
  endpoint_hint,    // optional and policy-controlled
  started_at_ms,
  heartbeat_at_ms,
  expires_at_ms,
  version,
  signature
}
```

The descriptor signature should cover all fields except `signature` and must be verifiable against the descriptor DID/public key.

## Scope

- Add a typed online-node descriptor instead of storing only raw DID strings.
- Automatically publish the descriptor at node startup.
- Periodically refresh `heartbeat_at_ms` / `expires_at_ms` while the node is running.
- Use `online_nodes` as the DHT topic for the registry.
- Add lookup/listing support that filters expired descriptors by default.
- Preserve current explicit service registration behavior separately from this online-node registry.
- Add tests for descriptor signing, refresh, expiry filtering, and multi-node registration.

## Node type requirements

At minimum, node type must distinguish:

- `browser`
- `native`
- `ffi`

The type should be part of the signed descriptor so peers cannot rewrite it in storage.

## Heartbeat requirements

- Heartbeat refresh interval should be configurable.
- Expiry/TTL should be explicit in the descriptor.
- Readers should treat descriptors with `expires_at_ms < now` as offline/stale.
- Stale records may remain in DHT storage, but normal listing APIs should hide them unless explicitly requested.

## Security requirements

- Do not trust unsigned DHT values.
- Do not accept a descriptor where `did` and `public_key` do not match.
- Do not let a storage node rewrite node type, capabilities, or heartbeat fields without invalidating the signature.
- Avoid making endpoint hints mandatory; browser nodes and privacy-sensitive nodes may not want to publish direct reachability metadata.

## Non-goals

- Do not replace service-specific registration in this issue.
- Do not implement HTTPS exit discovery here beyond allowing capabilities to advertise it.
- Do not require all node types to publish the same endpoint fields.

RyanKung:
## Summary

- Add signed report return policies for relay messages, including path-based and routed reports.
- Add signed online node descriptors in the node layer, a reusable node registration task abstraction, default-on `advertise_presence` controls, delta heartbeat publishing, network-scoped lookup support, configurable node type, and tombstone cleanup for superseded heartbeats.
- Preserve DHT storage placement across remote entry operations, validate placed operations against the entry's affine replica set, and cover split-owner redundant writes.
- Expose peer measurement counters through processor, internal RPC, native provider, and browser provider APIs without leaking the full ledger through external `nodeInfo`.
- Bump workspace and package versions to `0.13.0`.

Fixes #645.
Fixes #647.

## Compatibility

- Breaking protocol change: `0.13` nodes cannot verify `0.12` traffic and `0.12` nodes cannot verify `0.13` traffic because transaction signatures now cover the report-return policy tag.
- Breaking storage-message change: remote entry operations now carry their concrete DHT placement key, and receivers reject placements outside the operation entry's affine replica set so redundant writes do not collapse to one replica or pollute unrelated carriers.
- Breaking config/API change: presence advertisement is controlled by `advertise_presence` and defaults to `true`; the earlier `register_online_node` and `publish_online_node` config/builder spellings are removed.
- Breaking API/JSON change: peer measurements are no longer embedded in `nodeInfo`; `peerMeasurement` returns `measurement: null` for unobserved peers and `measurement.counters` for observed peers, while bulk measurement listing is the internal `listPeerMeasurements` method.
- Breaking browser lifecycle change: `Provider::listen()` is now a long-running listener; its Promise does not resolve during normal operation and must not be awaited as a readiness barrier.

## Follow-up

- #651 tracks storage-owner GC for expired online-node descriptors so crash or hard-stop sessions leave no long-term expired registry payloads.

## Validation

- `git diff --check`
- `cargo check -p rings-rpc --lib`
- `cargo check -p rings-core --lib`
- `cargo check -p rings-node --lib`
- `cargo check -p rings-node --no-default-features --features node --lib`
- `cargo test -p rings-core storage --lib`
- `cargo test -p rings-node custom_registration_task_publishes_through_shared_dht_sink --lib`
- `cargo test -p rings-node presence_advertisement --lib`
- `cargo test -p rings-node peer_measurement --lib`
- `cargo test -p rings-node online --lib`
- `cargo test -p rings-node --lib`
- `cargo test --release -p rings-node --lib`
- `cargo check -p rings-node --no-default-features --features browser --target wasm32-unknown-unknown --lib`


RyanKung:
Reviewed at 6aa4ca68 against the requirements of #645 and #647. Logic review only; gates are CI's to report.

## Blockers

1. **The registry write model loses updates and can erase the registry.** `publish_online_node_descriptor` (`processor.rs:465-483`) is a read-modify-write against one shared entry, and `storage_store` is `EntryOperation::Overwrite` (`handlers/storage.rs:360-366`). Two consequences:
   - *Lost updates*: N nodes heartbeat the same entry every 30s; two concurrent publishers each fetch, append self, and Overwrite — last writer erases the other's descriptor until its next beat. With TTL 90s and 30s beats, contention makes live nodes flap offline in listings.
   - *Cold-cache wipe*: `storage_fetch` for a remotely-owned entry only initiates the lookup — the `FoundEntry` response lands in the local cache asynchronously (`handlers/storage.rs:349-357`, `FoundEntry` handler). A publisher whose cache hasn't been filled yet reads nothing, and Overwrites the registry down to exactly one descriptor: its own.
   The single-node tests cannot witness either failure — the local-storage hit path is synchronous, and there is one writer. The crate already carries the correct primitives: `EntryOperation::Extend`/`Touch` are delta operations (entry.rs proves strong eventual consistency for data deltas), and `Tombstone` is the two-phase remove for pruning superseded records. The publisher should `storage_touch_data(ONLINE_NODES_TOPIC, descriptor)` — publish only its own delta — and leave merge to the entry CRDT; expiry/newest-per-did filtering already happens read-side in `latest_valid_by_did`. A topology test (two publishers, one storage owner, interleaved beats) would then witness what the current tests cannot.

2. **The hash change is a silent lockstep-upgrade requirement.** `hash_transaction` now appends the policy tag for *every* message, including default `Path` (`payload.rs:69-86`), so 0.12↔0.13 signature verification fails in both directions — decoding survives (`#[serde(default)]` works because the wire codec is gzip+JSON), but every cross-version payload is rejected as a bad signature, and mixed-version peers additionally rack up `failed_to_receive` and get down-ranked by the quality ordering. The tagged preimage is the cryptographically right choice — keeping the legacy preimage for `Path` would be malleable across policies (a signed `Path` transaction whose `data` begins `0x01 ‖ did` reinterprets as `Routed` with truncated data) — so the break is justified; but it must be stated in the PR body/release notes as "0.13 nodes cannot verify 0.12 traffic and vice versa", because nothing in the diff or the version bump says it.

## Model gaps

3. **`lookup_online_nodes` ignores `network_id`.** A signature-valid descriptor from another network passes the filter and is returned (`processor.rs:487-505`). The field is signed and locally known (`swarm.network_id()`); filter by it by default.

4. **`online_node_type()` derives a deployment property from compile features** (`processor.rs:410-423`). Feature unification distorts it: any build with `ffi` enabled reports `Ffi` (e.g. `--all-features`, or a native binary compiled in a build that also enables the ffi surface). Issue #647 makes node type a signed, load-bearing field. Make it a builder/config input with the cfg only as the default.

5. **Unenforced timing invariant and no opt-out.** Nothing checks `online_node_heartbeat_interval < online_node_ttl`; a config with interval > TTL makes every node oscillate offline between beats — enforce it at `ProcessorConfig` construction (`Invariant: heartbeat_at + interval < expires_at`). Separately, publication is unconditional for every node once `listen()` runs; #647 explicitly worries about privacy-sensitive nodes and endpoint hints, and pubkey/version/node-type broadcast has the same character — a disable flag is cheap now and breaking later.

## Design

6. **`OnlineNodeDescriptor::new_signed` stacks four smells that one shape removes** (`online.rs:80-118`): 11 positional parameters under `#[allow(clippy::too_many_arguments)]` (the repo convention is to fix the root cause, not allow); a placeholder `MessageVerification::new(&[], ...)` that performs a real, discarded signing operation over the empty transcript; and the `OnlineNodeDescriptorSigningData` mirror struct, where a future field added to the descriptor but not the mirror silently escapes the signature. Restructure as `descriptor body + signature`: the body struct *is* the signing data (`sign : Body × SessionSk → Signed<Body>`), the constructor takes one value, the placeholder disappears, and a new field cannot bypass the preimage. If the flat struct must stay, destructure `let Self { signature: _, .. } = self` in `signing_data` so field additions become compile errors, and let a params struct replace the 11 arguments. `Swarm::online_node_descriptor`'s 7-argument passthrough collapses with it.

7. **`PeerMeasurement` duplicates `PeerQualityEvidence`** — the same six counters and a copy-pasted `from_measure` (`measure.rs:95-148` vs `152-185`). Compose (`PeerMeasurement { did, evidence: PeerQualityEvidence }` or a shared counters struct with two views); two sources for one set of counters will drift.

8. Minor: `latest_valid_by_did` clones inside `and_modify` (a `match entry` avoids it); `send_report_message` re-validates the policy that `Transaction` verification already enforces — defensible as defense-in-depth at the send boundary, but worth a one-line comment saying it is deliberate.

## Sound — for the record

- The #645 security requirements are genuinely met by construction: the policy lives inside the signed preimage with an injective tagged encoding (`Path`/`Routed` disambiguated by the first byte); `validate_authorized_by` pins the routed destination to the request signer, which closes reflection amplification by requiring proof of control; reports reuse the request `tx_id`, satisfying the binding requirement; `Path` remains the default and no call site switches, matching the opt-in scoping.
- The descriptor verification chain (`did == public_key.did() == session.account_did()` and session pubkey equality) satisfies #647's "storage node cannot rewrite fields" requirement.
- The tests witness the right propositions: policy tamper flips verification; foreign routed destinations are rejected at construction; both relay shapes and the missing-next-hop case; descriptor tamper, encode/decode round trip, expiry filtering with `include_expired`, newest-per-did selection, multi-node listing; and the measurement path end-to-end through `node_info` and the RPC method.
- The RPC additions (method, request/response, handler dispatch, browser/native provider surfaces) are consistent with every existing method's shape.

RyanKung:
Re-reviewed at be8ccc20 (fresh read of the five new commits). Logic only; gates are CI's.

## Resolved from round 1 — all of it, and mostly by the strong form

- **The registry write model is now delta-based.** `publish_online_node_descriptor` is a single `storage_touch_data(ONLINE_NODES_TOPIC, descriptor)` (`processor.rs:548-557`) — no read-modify-write, no `Overwrite`. Both failure modes (concurrent lost updates, cold-cache wipe) are structurally gone.
- **The compatibility break is now stated in the PR body** in exactly the required form ("0.13 nodes cannot verify 0.12 traffic and vice versa"), plus the JSON shape change.
- **`lookup_online_nodes` filters by network** via the named predicate `matches_network` (`processor.rs:572-574`), with a two-network test.
- **Node type is a config/builder input** with the cfg derivation demoted to a default (`default_online_node_type`), tested via `online_node_type_is_configurable`.
- **The timing invariant is enforced** at both construction paths (`validate_online_node_timing` in `TryFrom` and `from_config`), and **publication has an opt-out** (`publish_online_node`) threaded through Config → Serialized → Builder → `listen()`, each with a test.
- **`new_signed` took the strong form**: `OnlineNodeDescriptorBody` is the signing data (`sign : Body × SessionSk → Descriptor`), the placeholder double-signing is gone, the `too_many_arguments` allow is gone, and `body()` destructures `Self { signature: _, .. }` — a future field addition now fails to compile instead of silently escaping the signature. `OnlineNodeDescriptorParams` collapses the swarm-layer passthrough the same way.
- `PeerMeasurement` composes `PeerQualityEvidence` (one source for the six counters), and `latest_valid_by_did` uses the map entry match instead of the cloning `and_modify`.

## New findings

1. **The Touch fix traded lost updates for unbounded growth.** Every heartbeat appends a permanently distinct value (fresh `heartbeat_at_ms`), and nothing ever removes one — the refresh test now *asserts* the accumulation (`stored.len() == 2` after two beats, `processor.rs:1022`). At the default 30s cadence that is 2,880 descriptors per node per day, forever, in one entry that `FoundEntry` transfers whole on every fetch; the registry becomes unusable within days on a real network. The crate already has the removal primitive: `EntryOperation::Tombstone` is a two-phase remove keyed by payload bytes, and the publisher knows its previous descriptor's exact bytes — tombstone the superseded self-record after touching the new one (or add storage-side GC of expired records). If pruning is deliberately deferred, it needs a filed follow-up issue referenced from the PR, because the current behavior is load-bearing and tested.

2. **The browser heartbeat cannot be stopped and leaks the node.** `spawn_interval` is fire-and-forget — `spawn_local` around an infinite loop, returning no handle (`utils.rs:233-248`). The browser arm of `online_node_heartbeat_daemon` (`processor.rs:597-612`) publishes once, spawns that detached loop with a cloned `Processor`, and returns. Consequences: dropping or aborting the `listen()` future — exactly what the frontend's node stop does — kills the native loop but not this one, so a stopped browser node keeps signing and publishing fresh "online" descriptors until the document dies; the captured `Arc<Swarm>` is kept alive; and every start/stop cycle stacks another parallel heartbeat loop. The native arm is lifetime-correct precisely because the loop lives *inside* the listen future. Fix by splitting the timer, not the daemon: one `loop { publish; sleep }` shape with a cfg'd cross-platform sleep (`window_sleep` on wasm, `futures_timer::Delay` native) keeps the browser heartbeat inside the listen future and cancellable with it.

3. Minor: `validate_online_node_timing` returns bare `Error::InvalidData` — the caller learns "invalid" but not *which* invariant (heartbeat_interval ≥ TTL) or by how much; a message-carrying variant names the proposition.

4. Note, related to 1–2: there is no unpublish on graceful stop — a cleanly stopped node stays listed as online for up to TTL (indefinitely in browser until 2 is fixed). A final descriptor with `expires_at_ms = now` (or a tombstone) on shutdown would make stops visible immediately; reasonable as part of the same follow-up as finding 1.

## Sound — for the record

The `body()` destructuring totality trick, the `matches_network` predicate, the params struct at both layers, and the test additions (timing rejection, opt-out, foreign-network filtering, configurable type) all land the round-1 items in their strongest form rather than the minimal patch. The RPC error mapping simplification (`Error::from` instead of the double map) and the `register_service` reuse of the new `storage_touch_data` passthrough are consistent cleanups.

RyanKung:
Re-reviewed at 7c3c27b6 (fresh read). Logic only.

## Resolved from round 2

1. **Growth is pruned, and at the right layer.** `EntryOperation::Tombstone` now covers Data entries: `DataTopicBuffer` carries the two-phase `removes` set, the constructor postcondition drops tombstoned adds and floors removes alongside values, join unions removes, and `RelayMessageSet` folds the buffer removes in — with the semilattice law test extended by a tombstoned sample and a no-resurrection test (`data_tombstone_removes_observed_payload_by_join`: re-joining the stale add does not bring the value back). The publisher tracks its own published encodings (`online_node_published_descriptors`), touches the new descriptor, tombstones the superseded ones, and the refresh test asserts `stored.len() == 1` again. The two-node integration test (`storage_tombstone_data_removes_observed_payload`) witnesses the op end-to-end through the storage owner. This is the strong form of the fix.
2. **The browser heartbeat is cancellable.** The daemon is one shape — `loop { publish; sleep }` inside the listen future — with only the sleep cfg-split (`futures_timer::Delay` native, `window_sleep` wasm). `spawn_interval` is no longer involved; aborting `listen()` now kills the heartbeat with it, which is exactly what the frontend's stop does.
3. `Error::InvalidConfig(String)` names the violated invariant, and the test asserts the message mentions both knobs.

## One compatibility fact to surface

**Browser `Provider::listen()` changed contract: its Promise never settles now.** Before this commit the wasm stabilizer scheduled itself and returned, so `await provider.listen()` resolved — the deleted test code (`try_join!(JsFuture::from(provider.listen()), …).unwrap()`) is the proof. With the heartbeat loop joined in, the Promise pends forever, and the tests were rewritten to stop awaiting it. That is the *correct* shape (it is what makes abort-based cancellation work), but any downstream JS caller that awaits `listen()` before continuing now hangs. This belongs in the PR body's Compatibility section next to the other two breaks, plus a line in the `listen` rustdoc ("runs until dropped/aborted; do not await completion").

## Bounded residuals — fine as follow-ups, worth naming

- **One garbage descriptor per node-session.** Pruning is publisher-driven, so a crashed or hard-stopped node's last descriptor stays in the entry forever (expired, hidden from listings, but stored and transferred). That is a per-session constant, not per-heartbeat growth — a ~3000× improvement — and the eventual complement is storage-side GC of expired descriptors (the storage owner can prove expiry from the signed `expires_at_ms`). Worth a filed follow-up.
- **By-value tombstones are order-sensitive.** A tombstone that reaches a replica before the add it removes is a no-op there (no observed dot matches), and the late add then survives; the publisher won't retry because the stale encoding already left its local set. Rare, bounded to one leaked (expired) value per reordering event, and the same storage-side GC covers it.
- `window_sleep` failure in the wasm sleep arm logs and returns immediately — a persistent failure (e.g. a context without timers) turns the heartbeat into a tight publish loop. Treat a sleep error as fatal to the daemon (return) rather than a skipped delay.
- A manual `publish_online_node_descriptor` racing the daemon can tombstone the other in-flight publication's fresh descriptor (both record before either retains); it self-heals on the next beat, but if the manual path stays public, a note or an internal serialization would pin it.

## Sound — for the record

The CRDT extension is where the semantics belong (carrier-level two-phase set with law tests), not in the registry code; the daemon/timer split matches the invariant that cancellation follows the future; the wasm test adaptations (`spawn_local`, dropped promises) are consistent with the new listen contract. All three round-2 items closed in their strong forms.

RyanKung:
Re-reviewed at 42ba983c — full adversarial pass over the final state, not just the delta. Logic only.

## Resolved from round 3

Sleep failure is fatal to the daemon (`processor.rs:643-651`); manual publish and heartbeat are serialized under an async lock; the listen contract is documented at all three layers, stated in the PR body's Compatibility section, and #651 is filed for storage-owner GC. All four residuals from round 3 are closed.

## New findings from the adversarial pass

1. **Descriptor verification has two time authorities and silently ignores one.** `OnlineNodeDescriptor::verify_signature` calls `MessageVerification::verify(&data)` directly (`online.rs:160-178`), bypassing `MessageVerificationExt::is_expired`. Every descriptor therefore carries a *signed* `ts_ms`/`ttl_ms` (DEFAULT_TTL_MS = 600s) that is never enforced, while `expires_at_ms` is enforced separately. The bypass is load-bearing: with `include_expired = true`, descriptors older than 10 minutes would fail an Ext-style verify, so switching to it would break the API — and nothing in the code says any of this. State the model: the descriptor's sole liveness authority is `expires_at_ms`; the embedded message TTL is vestigial. Document it on `verify_signature`, or implement `MessageVerificationExt` and set the message TTL from the descriptor TTL so there is one authority.

2. **The predicate is named but its one call site reconstructs it.** `is_live_at` (`online.rs:186-189`) has zero callers — grep finds only the definition — while `latest_valid_by_did`'s `!include_expired` filter is *literally* `verify_signature() && !is_expired_at(now)` spelled out inline (`online.rs:199-206`). Either the filter uses the predicate or the predicate is deleted; a dead public predicate adjacent to its own inlined reconstruction is the exact failure mode the naming was for.

3. **Growth is still unbounded — it moved into the tombstone set.** Each heartbeat nets: values stay at one per node, `removes` gains one `EntryDot`, forever — the 2P-set keeps its tombstones by construction, so the registry entry grows by ~2,880 dots per node per day instead of full descriptors. Reduced ~100× in bytes; still monotone. The bounding mechanism already exists in the CRDT you just extended: `DataTopicBuffer::new` floors **both** values and removes below the `register` reset floor (`crdt.rs:170-180`), so a periodic compacting `Overwrite` (live descriptors, fresh version) prunes tombstones soundly. #651's text covers "expired registry payloads" only — its scope must include tombstone-dot compaction, and the register floor is the named mechanism to cite.

4. **Two locks guard one invariant.** `online_node_publish_lock: AsyncMutex<()>` serializes publishes, and `online_node_published_descriptors: Mutex<BTreeSet<Encoded>>` guards state that is only ever touched inside that serialized section (`record_…`/`retain_…` have no other callers). A lock over `()` next to state under a second lock is the lock-the-code shape; `AsyncMutex<BTreeSet<Encoded>>` is lock-the-data — one lock, no poisoning path, and `Error::Lock` plus both `map_err`s disappear.

5. **The serialization fix is unwitnessed.** No test drives two concurrent `publish_online_node_descriptor` calls (`futures::join!`) and asserts the postcondition the lock exists for: exactly one live self-descriptor stored, no cross-tombstoned fresh record. The round-3 race was found by reading; the lock that fixes it deserves a witness so it cannot be removed silently.

6. **The registry has no true multi-node network test.** `online_node_registry_lists_multiple_nodes` seeds the second node's descriptor by writing storage directly; `storage_tombstone_data_removes_observed_payload` is the only two-node network test, and it lives at the storage layer. Missing: two connected processors, both `publish`, both `lookup`, both visible — the composed path (publish → touch → owner join → fetch → filter) is currently proven only in fragments. For the skill's induction framing: the N=2 base case of the registry protocol is untested end-to-end.

7. **`body()` allocates per verification.** `latest_valid_by_did` verifies every candidate, and each `verify_signature` clones the full field set (capabilities vector, strings) through `body()` (`online.rs:117-146`). The destructuring gave you totality; a borrowed serialize view keeps it and drops the clones: `#[derive(Serialize)] struct BodyRef<'a> { … }` built from the same `let Self { .. } = self` destructure. Zero-copy and the compile-time field check are not in tension.

8. **Capabilities are naked strings.** `"storage"` / `"snark"` literals in `online_node_capabilities` (`processor.rs:521-529`) with no constant or enum on either the write or read side; a typo advertises a nonexistent capability and no reader can match reliably. Issue #647 wants open-ended labels — fine — but the labels this crate itself emits should be named constants next to `ONLINE_NODES_TOPIC`.

9. **`Processor` derives `Clone` with an undocumented sharing law.** Clones share the swarm, the publish lock, and the published-descriptor set (all `Arc`), which is correct handle semantics — but the skill's rule is that `Clone` on a handle documents its law. One doc line: "cloning shares the node; publishes from any clone are serialized against each other."

10. Vacuous fallibility: the native `sleep_online_node_heartbeat_interval` returns `Result` only for cfg symmetry with the wasm arm (`processor.rs:95-99`). Acceptable; a comment saying the native arm is infallible would stop a reader from hunting for its error path.

11. **Pre-existing, adjacent, worth an issue: `is_expired` has an underflow panic path.** `self.verification().ts_ms - TS_OFFSET_TOLERANCE_MS > now` (`verify.rs:87`) underflows u128 when a peer sends `ts_ms < 3000`: panic in debug builds (a hostile payload aborts a debug node), wraparound in release (the guard misfires but the later `now > ts + ttl` check still rejects). Not introduced here, but this PR embeds `MessageVerification` in a new attacker-writable, long-lived surface (DHT-stored descriptors), so the hostile-`ts_ms` input class is now easier to reach. `checked_sub` is the fix; separate issue is fine.

## Verdict shape

Findings 1–4 are model/design debts in code this PR introduces and should land here; 5–6 are missing witnesses for invariants this PR created; 7–10 are hygiene at the severity this codebase holds itself to; 11 is a filed-issue candidate. Nothing found this round invalidates the protocol design itself: the signed policy, the descriptor binding chain, and the CRDT extension all held up under the adversarial pass.

RyanKung:
Re-reviewed at e7b4df86 (fresh adversarial read of the new commit). Logic only.

## Round 4: 11 of 11 closed, most in the strong form

1. **Time authority documented.** `verify_signature`'s rustdoc now states that the embedded `MessageVerification` ts/TTL is not a liveness rule and that `expires_at_ms` is the sole authority, pointing at `is_live_at` for enforcement.
2. **The predicate has its call site.** `latest_valid_by_did`'s live branch is now `is_live_at(now_ms)` instead of the inlined reconstruction.
3. **#651 rescoped correctly.** The issue text now names both growth vectors (expired payloads *and* tombstone dots per heartbeat) and cites the exact mechanism — a periodic compacting `Overwrite` so the `DataTopicBuffer` register floor prunes old dots.
4. **One lock, lock-the-data.** `AsyncMutex<BTreeSet<Encoded>>` replaces the `()`-mutex + std-Mutex pair; `record_`/`retain_` are inlined under the single guard; `Error::Lock` and the poisoning path are gone. The reordering (touch → insert → per-stale tombstone → remove) keeps the self-healing property on partial failure.
5. **The lock is witnessed.** `online_node_concurrent_publish_keeps_one_self_record` drives two publishes through clones under `try_join!` and asserts the postcondition (one stored record, one listed, heartbeat from one of the two) — which also exercises the documented Clone law.
6. **The N=2 base case is proven end-to-end.** `online_node_registry_lists_two_publishers_over_network` runs a real connect, stabilizes to mutual topology, pins registry ownership deterministically (rejection-sampling keys until the owner covers every `rotate_affine` placement, with a comment explaining *why* — the first-placement-hit fetch semantics), then has both nodes publish and both converge through lookups. This is the topology-aware test the protocol needed.
7. **Zero-copy with retained totality.** `OnlineNodeDescriptorBodyRef<'a>` serializes borrows; `body_ref()` still destructures `Self { signature: _, .. }`, and a new field now forces updates through *both* mirrors at compile time (the pattern and the `new_signed` construction).
8. Capability labels are named constants (`ONLINE_NODE_CAPABILITY_STORAGE`/`_SNARK`), exported and used on both sides.
9. The `Processor` Clone law is documented — and tested, via finding 5's test.
10. The native sleep's vacuous `Result` carries its explanatory comment.
11. **The underflow was fixed here rather than deferred**: `saturating_sub` plus a regression test that pins `ts_ms = 0 ⇒ is_expired() == true` without panic. Better than the issue I asked for.

## Remaining, all small

1. **`NETWORK_TEST_LOCK` guards an unnamed resource.** The new static mutex serializes six network tests — including three pre-existing ones — but no comment says *what* is contended. If the two-node networks are truly isolated in-process, the lock is unnecessary; if they are not (UDP port pressure, STUN pacing, tokio timing), the contended resource should be named where the lock is defined. A mutual exclusion whose reason lives only in someone's memory is the seed of the next flake investigation.
2. **`signing_data` has two sources whose byte-agreement is load-bearing.** `new_signed` signs `Body::signing_data()` (owned serialize); `verify_signature` checks `BodyRef::signing_data()` (borrowed serialize). Serde guarantees the encodings match today, and every sign→verify test witnesses it implicitly — but the clean shape is one source: have `OnlineNodeDescriptorBody::signing_data` construct a `BodyRef` from `&self` and delegate. One function owns the preimage; the equivalence question disappears instead of being answered.
3. Cosmetic: `BodyRef.version: &'a String` (and `endpoint_hint: &'a Option<String>`) — `&'a str` / `Option<&'a str>` serialize identically and drop the double indirection; requires adjusting the destructure bindings only.

## Verdict shape

Every finding from the adversarial round is closed with a witness, not a promise: the two structural fixes (single lock, borrowed preimage) came with the compile-time totality preserved, the two missing tests exist and test the right propositions, the pre-existing underflow got a regression test, and the follow-up issue names the mechanism rather than the symptom. The three items above are single-digit-line changes; nothing found this round touches the protocol design.

RyanKung:
Re-reviewed at `e42c58d` (the commit after round 4's `e7b4df8`). This round I also ran the gates locally on the pinned 1.94 toolchain rather than reviewing logic only: `cargo fmt --all -- --check` clean; targeted `rings-core` suite (tombstone, `data_topic_buffer`, `report_return`, descriptor/online) **12/12**; targeted `rings-node` suite (`online_node`, `peer_measurement`) **11/11**.

## What `e42c58d` does — all of it correct

1. **The `listen()` breaking change's blast radius is now fully swept.** All three examples get a real teardown story: `dweb` and `proof-demo` grow a node struct holding an `AbortHandle` with `stop()`, and — the part that's easy to forget — the abort is wired into the Yew effect *destructor*, so unmounting the component actually stops the listener instead of leaking it. The `dweb` wasm test stops both nodes at the end. `frontend` is simplified to run `Abortable::new(processor.listen(), …)` directly instead of round-tripping through the JS promise, which is strictly better: aborting the Rust future is meaningful, aborting a `JsFuture` wrapper never was.
2. **The `#[allow(clippy::arc_with_non_send_sync)]` is justified where it's used** — the comment explains the constraint (upstream `Provider` takes `Arc<Processor>`; browser code is single-threaded) rather than silencing the lint bare. This is the right way to carry an allow.
3. **`online_node_registry_lists_multiple_nodes` now goes through the production write path** (`storage_touch_data` delta) instead of `storage_store(registry_entry(…))`. I checked the other two remaining direct-store call sites and they are *correctly* left alone: the expired-descriptor and foreign-network tests must seed states the production publish path would rightly refuse to create — direct store is the only honest way to arrange those preconditions.
4. **The two-publishers test wait order is now causally right**: wait for the *owner* to observe each publish before asserting the publisher's view, and drop the pre-convergence publisher-side wait that raced propagation. This is a flake fix, not a test weakening — every assertion that existed still exists.
5. The `ONLINE_NODE_CAPABILITY_SNARK` import is `cfg(feature = "snark")`-gated, matching its only use.

## Round-4 leftovers — status unchanged

The three small items from the previous round remain open, and remain optional: (1) `NETWORK_TEST_LOCK` still doesn't name the contended resource; (2) `OnlineNodeDescriptorBody::signing_data` and `BodyRef::signing_data` are still two independently-serialized sources whose byte-agreement is load-bearing — delegating one to the other is a five-line change that deletes the equivalence question; (3) `BodyRef.version: &'a String` / `endpoint_hint: &'a Option<String>` still carry the double indirection. None of these block; if #651 or any follow-up touches these files, fold them in.

## Verdict

Nothing new found. The delta is a disciplined cleanup of exactly the collateral the breaking `listen()` contract created, the one test that bypassed the production write path now uses it, and the tests that *must* bypass it were correctly recognized and left alone. With the gates run green locally on top of five rounds of logic review, this is ready to merge from my side; the three cosmetic leftovers can ride along with #651.

crates/node/src/online.rs
```diff
@@ -0,0 +1,390 @@
+#![warn(missing_docs)]
+//! Signed online-node descriptors stored in the DHT.
```
RyanKung, line 2: I dont think online module should be here, instead, it should be placed as part of Node.
Core is only focus on protocol

RyanKung:
Layering review at e7b4df86, prompted by the question of whether `online` belongs in core. It does not, and pulling that thread exposes three placement errors that share one root.

## 1. `core/src/dht/online.rs` is node-layer content in protocol-layer clothing

The module's own vocabulary testifies against its placement:

- `OnlineNodeType::{Browser, Native, Ffi}` enumerates **rings-node's compile features**. Core has no browser/ffi axis; that distinction exists two layers up, in the crate that has `[features] browser` and `ffi`.
- `ONLINE_NODE_CAPABILITY_SNARK` makes rings-core name a concept from **rings-snark — a crate core does not know exists**. The dependency arrow points upward in vocabulary even though it can't in `Cargo.toml`.
- `version` is populated from rings-node's `build_version()`; `endpoint_hint` from rings-node's `external_address` config; the heartbeat interval, TTL, publication policy, and opt-out all already live in `Processor` — correctly. The type is in core while every producer, consumer, and policy is in node.
- Nothing in core consumes it: chord, stabilization, and entry never mention descriptors. The only core-side consumer is `Swarm::online_node_descriptor` — which this PR itself added, so the placement justifies itself circularly.
- The existing precedent already decides this: `register_service` implements the *service* registry entirely in `Processor` over generic core storage ops, with no core type. The online registry is the same pattern with a richer schema.
- Even within core the placement is a category error: `dht/` holds Chord-structural models (subring is a finger structure); a descriptor is stored *content*, not DHT structure.

The move is clean because the module only consumes exported core primitives (`SessionSk`, `MessageVerification`, `Encoder`/`Decoder`, `Did`) — and consuming core primitives from node is the correct direction. `ONLINE_NODES_TOPIC`, the capability constants, and the prelude exports this PR added to both core and node preludes all travel with it.

## 2. `Swarm::online_node_descriptor` + `OnlineNodeDescriptorParams` smear application signing into the swarm

The swarm method exists only to reach four values: `did()`, `network_id()`, the account verification pubkey, and the session key. This PR already added the first three as accessors, and `Processor` holds its own `SessionSk`. So the swarm gains an application-payload signing method it never needed: the node layer can build and sign the descriptor entirely from accessors. Both the method and the params struct should be deleted from `swarm/mod.rs`; `OnlineNodeDescriptorParams` is a node-layer parameter object that currently lives in core because the builder it feeds was put there.

## 3. The RPC wire schema broke its own documented contract

`protos/rings_node.rs` opens with: the types are "plain serde structs… kept identical to the previous prost-generated output to preserve the on-the-wire JSON shape." Before this PR the module had **zero** `rings_core` imports; `SwarmInfo` is the established pattern — a proto-local DTO mirroring core's `SwarmInspect`, converted at the boundary (`swarm.inspect().await.into()`).

This PR imports `OnlineNodeDescriptor` and `PeerMeasurement` directly into the wire schema. Three consequences:

- The public JSON API is now coupled to core's internal representation — refactoring the core type silently changes the wire shape the module doc promises to preserve.
- `OnlineNodeDescriptor` now serves **three compatibility contracts with one Rust type**: the bincode signing preimage, the bincode DHT storage encoding, and the JSON RPC response. These have different lifecycles (a signature preimage must never change; a wire response may want to evolve) and must not be the same type.
- It is also the one thing that blocks relocation 1: rings-rpc cannot name a rings-node type. With proto-local DTOs (`OnlineNodeDescriptorInfo`, `PeerMeasurementInfo`) and `From` conversions in rings-node — exactly the `SwarmInfo` pattern — the constraint disappears.

Supporting tell: the `Serialize`/`Deserialize` derives this PR added to core's measure types exist solely to serve the RPC path; with the DTO discipline restored they become unnecessary.

## What is correctly placed — for the record

`ReportReturnPolicy` in `message/protocols` is protocol-layer by definition (it lives inside the signed transaction preimage) — correct. The `Entry` tombstone/CRDT extension is carrier semantics — correct, and the right call was made putting it there rather than in registry code. `Measure`/`PeerQualityEvidence` are transport-observation infrastructure consumed by core's own quality ordering — correct. `PeerMeasurement` (did + evidence) is constructed by core's transport, so its core placement is defensible; its *direct use in the wire schema* is the part to fix.

## Shape of the fix

Move `online.rs` (with topic + capability constants) to `crates/node`; delete `Swarm::online_node_descriptor`/`OnlineNodeDescriptorParams` and build+sign in `Processor` from swarm accessors; restore the proto-DTO discipline for both new response types. No protocol behavior changes; the bincode storage encoding and signature preimage stay byte-identical since the serde shapes move unchanged. The diff is mostly `use`-path churn plus two small `From` impls.

RyanKung:
Follow-up review at e7b4df86: is the measurement exposure correct? No — on one security-relevant axis and three model axes.

## 1. The traffic ledger leaks through the external RPC surface

`ExternalRpcHandler` serves `NodeInfo` (`rings_node_handler.rs:178`, `193-198`), and this PR embedded `measurements: Vec<PeerMeasurement>` unconditionally into `NodeInfoResponse`. Any remote caller of the external endpoint therefore receives the node's complete per-peer ledger: every connected peer DID with connect/disconnect/sent/failed/received counts — the node's contact graph plus traffic volumes, reconnaissance-grade operational metadata.

The contradiction proves this wasn't the intent: the dedicated `PeerMeasurement` method was added to `InternalRpcHandler` only (external requests fall through to "method is not allowed"). Gating the single-peer query while broadcasting the full table through `nodeInfo` gives an attacker strictly more than the gated method would. Fix: remove `measurements` from `NodeInfoResponse` — the external contract — and expose the bulk listing as its own *internal* method (`listPeerMeasurements`), symmetric with the single-peer one.

## 2. `empty()` erases the state the codebase already models one function away

`SwarmTransport::peer_measurement` returns `PeerMeasurement::empty(peer)` when `self.measure` is `None` — all-zero counters. The adjacent `quality()` in the same file returns `PeerQuality::Unknown` for exactly this case (`transport.rs:395-397`): the codebase already knows "no measurer" is a distinct state, and the new API un-models it. Consequences compound in the bulk path: a measure-less node reports an all-zero row for *every* connected peer — "we never exchanged a message with anyone", which is false. `Option<PeerMeasurement>` (or a `Measured`/`Unmeasured` enum) is the honest signature; the RPC layer can then say `null` instead of lying with zeros.

## 3. Never-seen DIDs answer with zeros

`peer_measurement(did)` for an arbitrary DID — the RPC accepts any string — returns zero counters, conflating "peer never observed" with "peer observed, zero events" on the query axis, same erasure as finding 2. Absence should be absent.

## 4. Two surfaces, one dataset, coupled costs

The bulk data rides `nodeInfo` while the single-peer data has a dedicated method. Every `nodeInfo` call now performs 6×N sequential storage reads (`Processor::peer_measurements` awaits each counter of each peer in series) whether or not the caller wants measurements — inspection cost silently scales with peer count. One dedicated measurement surface (single + bulk, both internal) keeps `nodeInfo` inspection-shaped and makes the cost opt-in. If the bulk loop stays, `join_all` over peers is the cheap improvement; the six counters per peer also come from six independent awaits, so a row is not a consistent snapshot — either document the non-atomicity on `PeerMeasurement` or add a multi-counter snapshot read to `Measure`.

## 5. Wire vocabulary (ties to the layering review)

The public JSON now reads `measurement.evidence.sent` — "evidence" is this crate's internal quality-ordering vocabulary, leaked verbatim because the core type is serialized directly. The proto-DTO mirror from the layering review would rename at the boundary (`counters`) and decouple the wire shape from the internal composition.

## Correctly shaped, for the record

The layering of the accessors is right: `pub(crate)` on the transport, a thin accessor on `Swarm`, passthroughs on `Processor`/`Provider`. Browser `get_peer_measurement` matches the browser provider's `get_node_info` convention. And putting the dedicated method on the internal handler only was the correct instinct — finding 1 is the embedding that circumvents it.

RyanKung:
Re-reviewed at f6a38631 (fresh full read of the three commits, including placement and exposure audits). Logic only.

## Resolved — the layering, exposure, and residual items, all of them

- **`online.rs` lives in `crates/node`**, consuming only exported core primitives; `OnlineNodeDescriptorParams` and `Swarm::online_node_descriptor` are deleted; `Processor` holds its own `SessionSk` and signs the descriptor itself from swarm accessors. The vocabulary and the crate now agree.
- **The proto module honors its own contract again**: zero core imports, four DTO mirrors (`OnlineNodeDescriptorInfo`, `OnlineNodeTypeInfo`, `PeerMeasurementInfo`, `PeerMeasurementCountersInfo` — the wire says `counters`, not `evidence`), conversions in the new `rpc_dto.rs`, and the serde derives dropped from core's measure types now that nothing serializes them.
- **The external leak is closed the right way round**: `NodeInfoResponse.measurements` is gone; the bulk listing is a dedicated `listPeerMeasurements` on `InternalRpcHandler` only (verified: `ExternalRpcHandler`'s method set is untouched); the single-peer response is `Option` — `null`, not fabricated zeros; `from_measure` returns `None` for never-observed peers with the law documented; the bulk path is `join_all` instead of serial.
- Round-5 residuals: `NETWORK_TEST_LOCK` now names the contended resource (process-global ICE/UDP and timing-sensitive callbacks); `signing_data` has one source (`Body::signing_data` delegates through `body_ref()`); `version: &'a str`.
- **The listen-contract change was propagated to every consumer**: dweb and proof-demo now mirror the frontend's `Abortable` pattern, components stop their node on teardown (and on rebuild), tests stop what they build, and the `arc_with_non_send_sync` allows carry the same justification comment the frontend established.

## New findings

1. **`PlacedEntryOperation` trusts an attacker-controlled placement key.** The old `HandleMsg<EntryOperation>` derived placement from the operation's own entry key (`entry_operate::<_, 1>`), so a peer could not lie about where data lands. The new handler applies `msg.op` at `msg.placement` after checking only *ownership* of the placement (`handle_placed_entry_operation` → `operate_entry_at_placement`, `storage.rs:166-196`) — never that the placement **belongs to the operation's entry**. Any peer can send `PlacedEntryOperation { placement: K, op }` for an arbitrary key K it can route to and have the owner store the entry under K. Two consequences: (a) storage-slot pollution under keys unrelated to the entry's carrier; (b) **denial of service by carrier squatting** — occupy a legitimate topic's replica placement with a mismatched-carrier entry before the topic's first write, and every subsequent legitimate op at that placement fails `validate_same_carrier` forever. The receiver has everything needed to validate: `msg.placement ∈ op_entry_key.rotate_affine(REDUNDANT)` (its own redundancy bound). That check restores the old implicit invariant — *the sender chooses which replica, never where a replica may live*. State it as the type's invariant and enforce it in the handler.

2. **The write-side redundancy fix deserves its own witness and a Compatibility line.** The change this commit quietly makes is substantial: previously a replica op arriving at placement #k's owner was re-derived with redundancy 1 and effectively converged on the primary key — write-side redundancy was broken, and every existing test passed because no test ever gave placements distinct owners. The new two-publisher test asserts per-placement storage, but its setup *rejection-samples until one node owns every placement* — deliberately avoiding the split-ownership topology that motivates `PlacedEntryOperation`. The scenario that distinguishes new code from old (placement #k owned by a different node than the primary) still has no test; a three-node arrangement (or two nodes with placements straddling their arcs) is the base case that would have caught the original bug. Separately: `Message::OperateEntry` changed wire shape, which is a second independent 0.12↔0.13 break — the PR body currently attributes the break solely to the signature-preimage change; add the message-schema line so nobody assumes restoring the hash restores interop. The Summary should also name the placement fix explicitly — it is a core routing correction, not layering fallout.

## Notes, small

- `OnlineNodeDescriptorInfo.did: Value` / `public_key: Value` / `signature: Value` punt those fields' wire schema back to core's serde — the coupling the DTO exists to break, in miniature. `did` serializes as a string; make it `String`. The two structured fields are a pragmatic `Value` for now, but a comment saying their shape follows core's serde would make the debt visible.
- The DTO timestamps are `u128`; serde_json errors at runtime above `u64::MAX`. Millisecond epochs fit `u64` for 584 million years — `u64` is the honest wire type.
- `evidence == PeerQualityEvidence::default()` as "never observed" is a naked comparison where a named predicate (`PeerQualityEvidence::is_unobserved`) would carry the proposition; and `None` now covers both "no measurer configured" and "peer never observed" — the lie is gone, but if callers ever need the distinction, a node-level `measurements_enabled` flag on the list response is the cheap place for it.
- `Processor` now holds a second `SessionSk` alongside the one in the swarm transport — same session, two holders; one doc line on the field ("same session the swarm signs with; kept for node-layer signing") would pre-empt the question.

Finding 1 needs to land in this PR — it is introduced by this PR's own refactor. Finding 2's test can reasonably ride the same commit as finding 1's validation, since the validation needs the split-ownership topology to be witnessed anyway.

RyanKung:
Re-reviewed at cc8ea2fd (fresh read; re-ran the placement/exposure audits over the new code). Logic only.

## Round 8 — both findings closed with witnesses, all four notes closed

- **The placement-trust hole is closed at the trust boundary.** `PlacedEntryOperation::validate_placement` enforces `placement ∈ op.did().rotate_affine(redundancy)`, called first in `handle_placed_entry_operation` — i.e. on the inbound-message path, before `find_successor` and before any storage touch. The invariant is stated on the type ("sender may choose a replica, cannot choose where the replica set lives"), and the check derives redundancy from the receiver's own `transport.storage_redundancy()`, not from anything the sender sends. The local store API path correctly does *not* validate — it computes placements by construction and isn't a trust boundary. Placement of the check is exactly right.
  - I checked the regression I was most worried about: the old handler hard-coded redundancy 1 for the operate path ("For relay message, set redundant to 1"), so I looked for a sender still emitting placements under redundancy 1 that the new redundancy-N validation would now reject. It does not exist — that hard-coded operate path was *deleted*, which is the write-side-redundancy collapse this PR fixes. The only remaining `<_, 1>` is `entry_lookup` on the search path (`storage.rs:403`), unrelated. No relay-message regression.
- **Both witnesses landed, and they test the right propositions.** `placed_entry_operation_rejects_non_affine_placement` drives a forged placement through the real handler and asserts rejection *and* that neither key was written. `remote_redundant_store_writes_split_replica_at_affine_placement` builds the split-ownership topology (rejection-sampled until primary and replica placements have distinct owners) that the round-2 test deliberately avoided, and asserts the replica lands at its affine placement on the *other* owner and not at the primary — the exact scenario that distinguishes the fixed code from the collapsing original.
- **Compatibility + small notes**: the `OperateEntry` wire-schema break is now its own Compatibility line; `did`/measurement `did` are `String`; timestamps are `u64` via a checked conversion; `is_unobserved()` is a named predicate replacing the naked `== default()`; the second `SessionSk` on `Processor` is documented; `public_key`/`signature` stay `Value` but now carry a comment that their shape follows core's serde.

## One small observation, not a blocker

`validate_placement` uses the *receiver's* configured redundancy, and the sender derived the placement under *its own* const `REDUNDANT`. Today these are identical everywhere (`DATA_REDUNDANT = 6` on every node-layer path, and `ensure_storage_redundancy` pins each call's const to the local config), so the invariant holds. But it encodes an unstated network-homogeneity assumption: a peer legitimately configured with a different storage redundancy would have its valid replicas rejected. That is a reasonable assumption for this network — worth one sentence on `validate_placement` naming it ("assumes network-uniform storage redundancy; a replica valid under a different redundancy is treated as misplaced"), so a future heterogeneous-redundancy change knows this is a site to revisit.

## Verdict

The security fix is correct, minimally scoped, and checked at the boundary; the two missing tests now exist and cover the forged-placement and split-ownership cases; every round-8 note is closed. Nothing new surfaced under the adversarial re-read beyond the homogeneity-assumption doc nit. From my side the logic, layering, exposure, and model concerns raised across the review are resolved; remaining acceptance (elegance/completeness bar) is yours to call.

PR #650, opened 2026-07-03
Refresh Rings paper target architecture

### LINKED ISSUE #649: Refresh the Rings whitepaper as a forward-looking target architecture
## Context

The repository-owned paper under `papers/` is a forward-looking whitepaper: it should describe Rings' ideal/target architecture, not merely the current implementation snapshot. The problem is not that every unimplemented mechanism must be removed. The problem is that the paper currently mixes three categories without clear boundaries:

1. directions that have been superseded or explicitly abandoned by the current protocol model;
2. target mechanisms that are not fully implemented yet but should remain in the paper with a complete logical design;
3. current implementation details that can serve as grounding evidence but should not reduce the paper to an implementation report.

This issue was opened after auditing `papers/rings.tex` from a clean worktree based on local `origin/master` at `542eeee3` (`Add Rings browser frontend and extension workflow (#644)`). A remote refresh attempt failed locally with an HTTP/2 framing error, so recheck against latest remote head before final merge.

## Update abandoned or superseded directions

- Replace the paper's DID-as-finite-ring/field language with the current and target-safe model: DID is a point in the Chord identifier circle / additive `Z / 2^160` space for routing and placement. Cryptographic fields/groups used by signatures, ElGamal, zk, or secret sharing are separate carriers and must not be conflated with Chord IDs.
- Avoid claiming support for arbitrary DID proof algorithms. The target can be pluggable, but it needs an explicit verifier interface, domain separation, recoverable vs non-recoverable proof handling, and compatibility rules.
- Remove RSA as a Rings E2E target unless we intentionally reintroduce it. The privacy target is better expressed around EC ElGamal/KEM-like handshakes, signed envelopes, and future homomorphic/private-compute use.
- Stop calling the message layer an MSRP implementation. If the target is MSRP-inspired chunk/report semantics, define Rings' own relay/framing semantics instead of claiming IETF MSRP compliance.
- Replace historical demo names (`Chatter`, hidden-service demo, BNS) with target application classes or currently maintained examples, depending on whether the section is vision or evidence.

## Complete forward-looking mechanisms instead of deleting them

- **Mailbox / offline delivery**: specify resource-ID derivation, encrypted append-only entry semantics, retrieval proof, anti-spam/admission rules, TTL/replication, and metadata leakage boundaries.
- **Hidden/service provision**: specify service descriptors, rendezvous lookup, capability/ACL model, relay path selection, and what anonymity claim is actually provided.
- **Subrings**: explain fixed point, membership, join/update operations, local copy vs DHT-persisted state, and convergence assumptions.
- **Sidecar Sequencer**: clarify whether it gives a partial order or total order, what the witness proves, how cross-chain/block witnesses compare, how replay prevention uses it, and what it does not solve.
- **Ranking / peer quality**: distinguish local quality counters from a target global reputation/ranking protocol; define evidence, scoring, Sybil resistance, challenge/appeal, freeze/slashing, and how routing decisions consume the score.
- **Secret sharing / private storage**: keep it as a target only with a correct cryptographic carrier model, threshold reconstruction story, placement/redundancy, and recovery/failure assumptions.
- **Performance**: label Chord-cited tables as theoretical/reference data and add a path for Rings-owned benchmarks/simulation before using availability claims.

## Proposed branch work

1. Keep the paper as a perfect-mode architecture document.
2. Update superseded math/protocol claims so they no longer contradict the current design direction.
3. Rewrite unimplemented sections as coherent target designs with explicit assumptions and threat boundaries.
4. Rebuild `papers/rings.pdf` and review the rendered output before PR.

RyanKung:
## Summary

- Reframe the Rings paper as a forward-looking target architecture rather than a snapshot of the current implementation.
- Rewrite the Algebraic Network section to introduce classic Chord first, then CorrectChord, then Rings-specific extensions for DID/resource rings, monadic protocol transitions, and verification-oriented invariants.
- Add the CorrectChord citation and include created / last-updated dates on the title page.
- Regenerate `papers/rings.pdf` from the updated LaTeX source.

## Validation

- `latexmk -xelatex rings.tex`
- `git diff --check`
- Checked `rings.log` for overfull boxes, undefined citations, and undefined references
- Rendered the first PDF page with `pdftoppm` and visually verified the title-page dates

Closes #649


RyanKung:
Reviewed the full diff against the checklist in #649, plus structural checks on the head revision of `rings.tex` (no dangling `\ref`s, all environments balanced, no leftover superseded terms — `sequancer`, `Albegraic`, `Reliablility`, `effeicent`, and `stabilizationization` are all gone, and every remaining `finite field` / `MSRP` / `RSA` mention is a deliberate negation, not a stale claim). This is a genuine correctness pass on the paper's model, not just a reframing. A few findings below, mostly in the new pseudocode; nothing blocking.

## What the rewrite gets right

- **The carrier correction is the important one, and it's done properly.** The old text claimed DIDs form a finite ring that "can even be treated as a finite field" and ran SSSS directly over it — but $Z/2^{160}$ has zero divisors and is not a field, and Shamir needs a field, so the old "SSSS over a finite ring" algorithm even had to smuggle in a prime $p$, contradicting its own title. The new model — Chord IDs as an additive cyclic group used only for offsets/bias/placement, with multiplication explicitly outside the routing model and crypto living in separate scalar-field/curve-group carriers — is the correct algebra, and the paper now says so in exactly the right terms.
- **The Zave citation checks out**: IEEE TSE 43(12):1144–1156, 2017, DOI `10.1109/TSE.2017.2655056`, arXiv 1610.01140 all match the published record. The claims attributed to it are accurate: original Chord's ring maintenance incorrect under its own operating assumptions, successor lists plus a stable base of $r+1$ central to the invariant, join/rectify/stabilize as the corrected transition family.
- **Real errors removed, not just softened.** The old sidecar text had the witness direction wrong (embedding a block hash proves a message came *after* the block, not "not later than"); the new before/after-bound formulation plus the partial-order/finality-domain caveat fixes it. The ElGamal-based "secret sharing" (encrypt-to-each-participant, which is not a threshold scheme) and the double-comma weighted-median ranking math are gone. "MITM entirely infeasible" and the inherited "99.85% availability" claim are now properly scoped.
- **Honest deviation naming.** "Sparse no-wrap fingers … a routing-state fact, not a Chord theorem" and the correctness-vs-performance split (successor list as safety root, fingers as optimization) are exactly the kind of statements #649 asked for.
- **LaTeX hygiene**: the three `\label`-before-`\caption` fixes mean `\ref{dht}`, `\ref{path}`, `\ref{failure}` now bind to the right counters, and `Figure \ref{failure}` → `Table \ref{failure}` fixes a genuine cross-reference type error.

## Findings

1. **`alg:lookup`: unreachable effect after `\Return`.** The remote branch reads `\Return $Remote(next)$` followed by `\textbf{effect} $FindSuccessor(key)$` — as written the effect line is dead code. Since the Monadic Effect Boundary section defines transitions as *returning* action values, the consistent form is a return value that carries the effect (e.g. `Remote(next, FindSuccessor(key))`), or emit-then-return.
2. **`JoinThenSync` violates the paper's own effect boundary.** `s ← known.find_successor(n.did)` and `t ← query_topology(s)` are synchronous remote calls inside the transition, directly contradicting "A DHT transition should not directly perform network IO" three subsections earlier — and `emit QuerySuccList(s)` is then redundant with the already-fetched `t`. Either present join as a composite of reply-driven transitions (which matches the described architecture) or annotate the sketch as collapsing several steps.
3. **`Lookup`'s `next = none` fallback asserts ownership it can't justify.** When `key ∉ (node.did, s]` and no closer preceding node is known, returning `Local(s)` claims `s` owns the key. On a converged ring this branch is unreachable (in that case `s` itself precedes the key), so the branch is either dead or unsafe under partial state — worth one sentence saying which, especially in a paper that elsewhere distinguishes safety from optimization carefully.
4. **Layer naming: the figure and the body disagree.** The list/figure name six layers — Identity, Transport, Overlay, Extension Runtime, Protocol, Application — but the subsections are still "Network Layer" (not Overlay), there is no Extension Runtime subsection (its content lives as a paragraph under Application Layer), and "Identity and Runtime Surfaces" spans two concerns. A reader cannot map the architecture figure onto the document structure. Renaming the subsections (or adding a sentence mapping old names to new) would close it.
5. **Bound symbol drift: $r$ vs $K$.** The Correct Chord section defines $r$ as the successor-list length ("stable base of at least $r+1$"); the pseudocode bounds the list with `take(K, …)` and $K$ is never defined. One symbol should serve both.
6. **Created date changed without an in-repo witness.** The old title page said `Feb, 10, 2023`; the new one says `Created: February 8, 2023`. The file's history only reaches back to the #644 move, so the repo can't confirm either date — if February 8 comes from an external record (first publication/announcement), fine; otherwise the only prior in-repo record said the 10th.

## Nits

- `@article` + `howpublished`: `unsrt` ignores `howpublished` on `@article`, so the arXiv URL will not render — use `note = {\url{…}}` if you want it visible.
- "A DID is a $2^{160}$-point Chord identifier" — the *space* has $2^{160}$ points; the identifier is 160-bit. The next clause says it correctly.
- The algorithm captioned "ElGamal Encryption Algorithm" now contains both `Encryption` and `Decryption` procedures (the standalone decryption block was deleted) — caption could say "ElGamal Encryption and Decryption".
- Notation inconsistency: `$Z / 2^{160}$` uses plain `Z` while the E2E section writes `$\mathbb{Z}_q$` — blackboard bold in both would match.
- Abstract: "Rings Network is a decentralized peer-to-peer network … **It describes** a target architecture" — the antecedent of "It" is the network, but the describing is done by the paper.
- The Conclusion still reads as present-tense capability claims ("offers a secure and efficient solution…"); one clause echoing the target-architecture framing would keep it consistent with the new abstract.
- `alg:join`, `alg:lookup`, `alg:biasdid`, and `ElGamal Encryption` are labeled but never `\ref`'d — the prose says "the **join** algorithm" instead. Harmless, but numbered references would tie the text to the right block.

## Verdict

The paper's algebra is now correct where it used to be wrong, the CorrectChord grounding is accurately cited and honestly scoped, and the forward-looking sections (mailbox, hidden services, subrings, ranking, sequencer) now carry the derivations, field lists, and threat boundaries that #649 asked for instead of hand-waving. The findings above are all local: two pseudocode blocks that contradict the paper's own effect-boundary discipline, a figure/body naming mismatch, and small notation drift. Fix those in this PR or a follow-up — either way this is a clear improvement to merge.

RyanKung:
Re-reviewed `2306860`. Every finding from the previous review is resolved, and several fixes go beyond what was asked. I re-verified the head structurally: the six subsections now match the six-layer figure exactly (Identity, Transport, Overlay, Extension Runtime, Protocol, Application), there are no dangling `\ref`s, and all environments remain balanced.

Per-finding:

1. **Lookup dead code** — fixed the right way: `a ← FindSuccessor(key); Return Remote(next, a)`. The action now travels in the return value, which is exactly what the Monadic Effect Boundary section promises.
2. **JoinThenSync synchronous IO** — fixed better than requested: instead of annotating the sketch, the join is now genuinely reply-driven (`BeginJoin` → `InstallSuccessor` → `InstallSuccList`), each transition returning `Remote`/`Multi`/`Noop` action values, with prose stating "No join transition directly calls a remote peer or mutates remote state." `take(r, s :: list)` is also the correct CorrectChord successor-list construction (self's successor followed by the successor's list, truncated to $r$).
3. **`next = none` ownership claim** — fixed with a `Repair(QuerySuccList(…))` action plus a new `s = none` guard, and the prose now names the branch's status precisely ("not an ownership claim… unreachable in a converged ring… under partial local state it returns a repair action"). That's the safety/optimization distinction stated the way the rest of the paper states it.
4. **Layer naming** — subsections renamed and restructured to match the figure, and the Wasm runtime material moved from the identity section into the new Extension Runtime Layer subsection, which is also its correct home. The dead commented-out OSI figure is gone too.
5. **$r$ vs $K$** — unified on $r$ everywhere, including stabilization.
6. **Created date** — now carries an external witness (`RingsNetwork/whitepaper` repository, timestamp `2023-02-08T13:27:55Z`), which settles the Feb 8 vs Feb 10 question with a source rather than an assertion.

All seven nits were handled as well: `note = {\url{…}}` in the bib entry, "160-bit Chord identifier in a space with $2^{160}$ points", the ElGamal caption/label (`alg:elgamal`, now referenced from prose), $\mathbb{Z}/2^{160}\mathbb{Z}$ notation throughout, the abstract's dangling "It", the conclusion reframed as target architecture with a closing sentence about validation work, and all four algorithm labels now referenced by number from the text.

Two truly minor residuals, neither worth another round:

- The abstract now opens two consecutive sentences with "This paper describes… This paper presents…" — trivially smoothable whenever the file is next touched.
- The Stabilization algorithm still uses inline `emit` while join/lookup now return action values. Since the Monadic Effect Boundary section defines multi-actions, `emit` reads as accumulation into one, so it's defensible — but ending Stabilize with a single `Return Multi(…)` would make all four sketches use one convention.

LGTM — merge when ready.

RyanKung:
Fourth pass, at `5c4a6b6`, with a different brief: treat this as a submission to a theory venue and audit (a) mathematical/terminological precision, (b) residual generated-text phrasing, (c) overall academic register. The new "Rings as an Algebraic Object" section is a real upgrade — the writer monad is the right minimal effect model, $\mathbf{Set}/R$ is the right home for namespaces, Law 2 is a genuine theorem, and the Moggi citation checks out (Inf. & Comp. 93(1):55–92, DOI correct). But the paper is now bimodal: the new formal core reads like a specification while the surrounding 2023 prose reads like a product page, and the formal core itself has several precision defects. Findings below, worst first.

## A. Mathematical and terminological precision

**A1. `Inv_place` excludes the owner from its own replica set — a formal bug.** $\operatorname{Succ}^r_N(n)$ is defined over $N \setminus \{n\}$, so it never contains $n$. The placement invariant then requires $replicas(H_\nu(x)) \subseteq \operatorname{Succ}^r_N(P^N_\nu(x))$ — i.e. the owner itself is *forbidden* from holding the entry. In Chord the successor (= owner) is precisely the node that stores the key. The invariant should read $replicas(H_\nu(x)) \subseteq \{P^N_\nu(x)\} \cup \operatorname{Succ}^{r-1}_N(P^N_\nu(x))$, or $\operatorname{Succ}$ should be redefined to include the node at index 0.

**A2. The interval convention breaks the degenerate ring, silently.** Definition 1 gives $(a,a]_R = \emptyset$ (since $\delta_a(a)=0$), whereas classic Chord treats $(n,n]$ as the full circle so a singleton ring owns everything. Under your definitions a one-node ring has $L_n = \emptyset$, so `Lookup` returns `Repair(QuerySuccList(node))` — a self-query that can never terminate. Since you already adopt Zave's stable base, the honest fix is one standing assumption stated once: $|N| \ge r+1$ throughout, with the bootstrap case explicitly out of scope. Right now the model neither handles the case nor excludes it.

**A3. $\operatorname{pred}$ is used but never defined.** It appears in the Rings-object tuple and in $\operatorname{Inv}_{ring}$ ($p_n = \operatorname{pred}_N(n)$) with no formula, while $\operatorname{owner}$ and $\operatorname{Succ}^r$ get definitions. One line: $\operatorname{pred}_N(n) = \arg\min_{p \in N\setminus\{n\}} \delta_p(n)$. While there: $\operatorname{owner}_N$ needs $N \neq \emptyset$, and a clause noting the argmin is unique because $\delta_k$ is injective on $N$.

**A4. The categorical vocabulary is currently decorative where it could be precise.** "Viewing $X_\nu$ and $N$ as discrete categories, $P^N_\nu$ is the placement functor" — every function is a functor between discrete categories, so this sentence asserts nothing. And the displayed equation $P^N_\mu \circ f = P^N_\nu$ is called "natural," but naturality is a property of a transformation between two named functors; what you actually have is simpler and better: a derivation-preserving $f$ is *exactly* a morphism of $\mathbf{Set}/R$, and post-composition with $\operatorname{owner}_N$ is a functor $\mathbf{Set}/R \to \mathbf{Set}/N$; your equation is that functor's action on morphisms. State it that way and the category theory earns its place — or drop the vocabulary. In the same family: "a present finger is a performance *morphism* toward the owner relation" — there is no category in which a finger entry is a morphism; the word you already use correctly elsewhere is *witness*. The abstract's "placement functors into the live Chord topology" inherits A4: as written the paper does not yet support "functor" in a non-vacuous sense.

**A5. The effect monad is the writer monad — say so, and resolve the error-composition ambiguity.** $\mathsf{T}_A(X) = X \times A^*$ is the writer monad over the free monoid $A^*$; naming it buys you the monad laws for free (they are the monoid laws) and one citation instead of a bespoke construction. More importantly, "typed errors add a sum component, $\mathsf{E}+\mathsf{T}_A(X)$, but do not change the effect-composition law" is not yet a definition: $E + (X \times A^*)$ and $(E + X) \times A^*$ are different monads — the first discards actions accumulated before a failure, the second preserves them — i.e. `ExceptT` over `Writer` versus `WriterT` over `Except`. Which one holds determines whether the interpreter executes actions emitted before an error, which is operationally load-bearing for a network protocol. Name the transformer stack and its law.

**A6. "Effect transparency" is vacuous as stated.** $\pi_S(\tau(s,i))$ is independent of $I_A$ *by construction* — $\tau$'s type never mentions the interpreter. The intended (and valuable) claim is about the implementation: every production code path factors as $I_A \circ \tau$, i.e. the node refines the pure transition composed with an interpreter. That is a refinement/factorization theorem obligation, not an invariant of the model.

**A7. One order, two notations.** Definition 1 introduces $\delta_a$; the State Algebra subsection re-defines the same function as $\operatorname{dist}_n(x)$; the BiasDID algorithm computes it a third way as $A-C$. Unify on $\delta$. In fact BiasDID stops being an algorithm at all once the formalism exists: $A \le_C B \iff \delta_C(A) \le \delta_C(B)$ — the biased order is the pullback of the linear order on $[0,M)$ along $\delta_C$, total for each fixed $C$. Replacing the algorithm block with that one-line definition (plus the remark that it is *not* translation-invariant in $C$, which is the whole point) would be strictly more precise and more academic.

**A8. `Law` conflates two modalities.** Law 2 (rotation equivariance) is a provable proposition — one line from translation-invariance of $\delta$. Law 1 (carrier separation) is a design constraint; it cannot be proved, only obeyed. Housing both in one theorem-like environment muddies what the paper claims versus what it legislates. Keep `law` (or `proposition`) for provable statements *with proofs* — `amsthm` is already loaded and the declared `proposition` environment is never used, so the paper currently proves nothing — and introduce a separate `constraint`/`discipline` environment for normative rules.

**A9. Protocol morphisms are underspecified.** "A morphism of protocols preserves placement and maps transitions to transitions" — the second clause needs the commuting square: maps $\sigma_X, \sigma_S, \sigma_I, \sigma_O$ plus a monoid homomorphism $\sigma_A^* : A_\nu^* \to A_\mu^*$ with $H_\mu \circ \sigma_X = H_\nu$ and $\mathsf{T}_{\sigma_A}(\sigma_S \times \sigma_O) \circ \tau_\nu = \tau_\mu \circ (\sigma_S \times \sigma_I)$. Without it, "protocol refactoring must commute with $\tau_\nu$" has no checkable content.

**A10. Smaller precision items.** The state-variable list includes a pending action list $Q$ that no displayed operator or invariant ever touches. $I_A : A^* \to IO$ is called a monoid homomorphism but $IO$ is not presented as a monoid — say "into IO actions under sequential composition." $\operatorname{Succ}^r_N(\cdot)$ is a list but $\operatorname{Inv}_{place}$ uses $\subseteq$ — coerce to a set or use list membership. The legacy Overlay Layer prose contradicts the formal section: "the information regarding the closest successor or predecessor of a node is stored in the finger table" (the paper's own safety-root claim puts them in $(p_n, L_n)$, not $F$), and "identifier ring, which has at most $2^m$ nodes" conflates identifiers with nodes. `$O(log N)$` renders as the product $l\cdot o\cdot g$ — `$O(\log N)$`; likewise `target_i`, `finger_i`, `replicas` need `\operatorname`. The title is missing an article: "for *the* sovereign age."

## B. Generated-text register (AI 句式)

The formal sections are clean. The residue is concentrated in the 2023 front matter, and it is recognizable:

1. **Three consecutive throat-clearing paragraphs** (Introduction): "The objective of this paper is to introduce…", "These findings will contribute to the ongoing discussions about the future of the internet…", "This paper aims to make a meaningful contribution to the discourse surrounding…" — three paragraphs, zero technical content. This is the strongest boilerplate signal in the paper. Replace all three with a bulleted contributions list (which the paper, notably, still lacks).
2. **Nominalization filler**: "facilitates the development of secure and trusted methods… This, in turn, promotes the establishment of secure and confidential communications and data exchange based on trusted relationships between parties" — an entire paragraph that survives deletion with no information loss.
3. **Unsupported comparatives in brochure register**: "surpassing the privacy and security offerings of traditional centralized networks"; "grants Rings Network a higher degree of scalability, efficiency, and robustness compared to unstructured P2P networks" (structured-vs-unstructured is a tradeoff — unstructured networks are famously *more* churn-tolerant — so this needs hedging or a citation); "a unique feature that sets it apart"; "complete control over their data sovereignty"; "smooth user experience"; "vast user base"; "ideal for video conferencing, online gaming, and live streaming."
4. **A leaked editorial note**: "Concrete demos may change over time; **the paper should describe** the application class and protocol requirement rather than depending on historical demo names." This is an instruction to the authors — issue-tracker language — sitting in the paper body. It must become prose ("We characterize applications by protocol class rather than by demo") or be deleted.
5. **The normative "should" register**: 13 lines instruct "Rings should…" / "the receiver…must…". A specification states what the design *is* and what holds under which assumptions; a review states what someone *should* do. Convert each to a definition, a numbered requirement with declared keyword semantics, or move it to Future Work.
6. **Duplicated content blocks**: the WebRTC handshake is explained twice (a seven-bullet sequence, then a three-bullet summary of the same thing, then prose repeating both), and two figures carry the identical caption "Exchange SDP" — they should be "Standard SDP/ICE exchange" vs. "Rings single round-trip handshake." The abstract still opens consecutive sentences with "This paper describes… This paper presents…".

## C. Academic structure

- **State assumptions once, formally.** $|N| \ge r+1$ (Zave's stable base), fail-stop overlay (Byzantine out of scope — currently said in prose twice), and the delivery assumptions behind $\Downarrow$ belong in a single "Model and Assumptions" block that Definitions and invariants reference. $\Downarrow$ itself deserves a one-paragraph definition (configurations, a step relation for interpreting `Remote`/`Repair` actions, $\Downarrow$ as reachability of `Local(\cdot)`) — it is the paper's main safety statement and currently rests on "under the modeled delivery assumptions."
- **Prove the provable.** Law 2, placement functoriality (A4), and the monad laws (by reduction to the writer monad) are each 1–3 lines. A theory paper that declares `definition`/`law`/`proposition` environments and contains zero `proof` environments undercuts its own register.
- **Missing citations for named artifacts**: join-semilattice replication is CRDT — cite Shapiro et al.; "TLA-like" — cite Lamport; Stateright is named with no reference; MSRP is discussed by name — cite RFC 4975. Mac Lane 1971 is fine (consider the 1998 2nd ed.).
- **Relocate the tutorials.** The WebAssembly explainer ("provides a way for developers to write high-performance code…") and the WebRTC video-conferencing paragraph are textbook content orthogonal to every claim in the paper; compress to two sentences plus citations, or move to an appendix on the implementation substrate.
- **Reference the formalism from the legacy sections.** Definition 1 is never cited by number; the Overlay Layer and Protocol Layer sections re-explain in prose what §Algebraic Network now defines. The paper will read as one document when the prose sections *use* $\delta$, $\operatorname{owner}_N$, $(a,b]_R$ instead of paraphrasing them.

## Verdict

The direction is right and the new core is worth keeping: carrier + slice-category placement + writer-monad effects + scoped verification boundary is a coherent and honest formal skeleton. To reach the theoretical register this paper is now aiming for: fix the two actual formal defects (A1, A2), define what is used (A3, A9), make the categorical and monadic language carry non-vacuous content (A4, A5, A6), separate theorems from design rules and prove the theorems (A8), and then delete or rewrite the 2023 marketing/boilerplate residue (B1–B6), which is now the largest single gap between this document and a paper one could submit.

RyanKung:
Fifth pass, at `04280c7`. Every item from the theory review — all ten precision findings (A1–A10), all six register findings (B1–B6), and all five structural recommendations — is resolved, and the fixes are the strong versions, not the minimal ones. I re-verified the head mechanically: **zero** occurrences of the normative "should" remain in the paper, no dangling `\ref`s, all environments balanced, and all four new bibliography entries check out against the published records (Shapiro–Preguiça–Baquero–Zawirski is the canonical CRDT paper with the correct LNCS volume and DOI; Lamport 2002 ISBN correct; RFC 4975 DOI correct).

Highlights of how the findings were closed:

- **A1** — `Rep^r_N(k) = {owner_N(k)} ∪ set(Succ^{r-1}_N(owner_N(k)))` is exactly right, and generalizing the successor-sequence definition to `Succ^q` so the instances at $r$ and $r-1$ don't collide was a good instinct. `Inv_place` now quantifies over the correct set.
- **A2** — the new Assumption (Overlay model) block states $|N| \ge r+1$, fail-stop, and the delivery scoping once, and explicitly classifies bootstrap rings as initialization states rather than counterexamples. This is the honest fix.
- **A4** — $\Omega_N : \mathbf{Set}/R \to \mathbf{Set}/N$ as an explicit post-composition functor, with placement functoriality as a proved Proposition, is precisely the non-vacuous formulation; the abstract's rewritten formal sentence now matches what the paper actually proves.
- **A5** — naming the writer monad, proving its laws from the monoid laws, and above all **committing to the error-over-writer stack with the semantic consequence stated in prose** ("a failed transition returns an error and no executable action list; … a recoverable fault report is an explicit successful action") resolves the operationally load-bearing ambiguity completely.
- **A6** — effect transparency restated as the refinement obligation $\operatorname{Impl}_A(s,i) \preceq I_A^\sharp(\tau(s,i))$ with $\preceq$ scoped: now a real proof obligation on the implementation instead of a tautology.
- **A7/A8** — BiasDID as a one-line pullback order definition (with the observer-dependence remark doing the conceptual work the old algorithm block couldn't), `dist` eliminated in favor of $\delta$, `Constraint` vs proved `Proposition` separating legislation from theorems. The paper now contains three proofs where it previously contained none.
- **B** — the three throat-clearing paragraphs are replaced by a four-item contributions list and a genuinely sharper problem statement ("how can browser and native runtimes share one peer-to-peer overlay without treating identity, routing, storage, and application protocols as centralized services?"). The leaked editorial note is gone, the structured-vs-unstructured comparison now states the tradeoff honestly instead of claiming dominance, the duplicate WebRTC exposition is collapsed with the two figures properly distinguished, and the "changes the Rings signaling protocol, not the WebRTC security model" clarification is a nice precision bonus. Stabilize now accumulates and returns `Multi(actions)`, closing the last convention residual from the previous round.

Four micro-nits, none requiring another round:

1. **$r$ does double duty.** The Assumption declares $r$ as the successor-list length; four lines later "for a replication factor $r$" introduces a second parameter with the same letter. Either state "we take the replication factor equal to the successor-list length" (which `Rep^r ⊆ {owner} ∪ Succ^{r-1} ⊂` list coverage in fact requires) or rename one.
2. **`::` points both ways.** `s :: list` prepends an element in `InstallSuccList`, but `actions :: FindSuccessor(c)` appends one in `Stabilize`. One convention (e.g. `actions ++ [a]`) would keep the pseudocode algebra honest.
3. The Stateright entry's `year = {2026}` reads as a publication year; `note = {Accessed July 2026}` is the usual form for a tool URL.
4. The Assumption's final clause defers delivery fairness to "the selected verification artifact" — mildly circular; naming the two artifact classes it means (exhaustive finite interleavings; converged-ring trace replay) would close the loop.

The paper is now a coherent document in one register: definitions are used where they are defined, claims are either proved, scoped by a named assumption, or marked as constraints, and the prose no longer contains instructions to its own authors. From my side this is done — LGTM.

RyanKung:
Reviewed at the PR head against #649 and the current implementation. This is a paper PR, so the review is technical/mathematical accuracy, internal consistency, and paper-vs-code fidelity — not lints.

## The rewrite resolves #649's concerns, and does so well

Each superseded direction the issue named is fixed at the model level, not just softened in prose:
- **DID-as-finite-ring/field is gone.** The identifier carrier is now `R = Z/2^160 Z` as an *additive cyclic group*, with an explicit `Constraint[Carrier separation]` (`Mult_R ∉ RouteOps_R`, `PlacementOps ∩ CryptoOps = ∅`, `CryptoCarrier ∈ {F_q, G, Z_q}`) and a carrier-boundary table. This is exactly the correction #649 asked for, and it is now a stated law rather than an aside.
- **Secret sharing no longer conflates carriers.** The old "SSSS over a finite ring" algorithm — which used the DID ring as the polynomial field — is deleted and replaced by `Secret sharing placement`: shares live in `F_q` (`sh_i ∈ F_q^2`), and only their *placement* `H_S(sh_i) ∈ R` maps into the overlay. Correct separation, correctly cited to Shamir.
- **RSA removed** from the E2E target; the carrier is the EC group `G` with `G ≠ R` stated explicitly.
- **MSRP compliance claim removed.** It is now "the split vocabulary follows MSRP" citing RFC 4975 — a vocabulary influence, not a conformance claim.
- **Demo names** (Chatter/BNS/onion hidden service) are replaced by formal target definitions (`Hidden service descriptor`, `Mailbox`, `Subring`) with descriptors, lookup maps, and auth predicates — the "complete the mechanism, don't delete it" path the issue preferred.
- **Performance tables are labeled Chord reference data** and paired with an `Evaluation morphism` obligation for Rings-owned benchmarks, addressing the "theoretical/reference vs Rings-measured" split.

The category/effects core is faithful to the code: the writer monad over action lists, the typed failure stack `E + (X × A*)` with `RecoverableReport ∈ A`, and the transition shape `τ : S × I → T^E_A(S × O)` are the effect-returning handler model the implementation actually uses; the join/lookup/stabilize algorithms were rewritten to return `Local/Remote/Repair/Multi` effects instead of performing IO, which matches the handlers. Rotation equivariance and placement functoriality are stated and correctly proved.

## One substantive error: the replica-set model contradicts the shipping placement

`Definition[Replica set]` defines
\[ \mathrm{Rep}^r_N(k) = \{\mathrm{owner}_N(k)\} \cup \mathrm{Succ}^{r-1}_N(\mathrm{owner}_N(k)), \]
i.e. **successor-list replication** (the owner plus its r−1 consecutive successors), and `Invariant[Placement]` then asserts `replicas(H_ν(x)) ⊆ Rep^r_N(H_ν(x))`.

The implementation does not place replicas this way. Storage redundancy uses **affine key rotation**: `Did::rotate_affine(r)` returns `k + floor(2^160 · i / r)` for `i ∈ 0..r` (`crates/core/src/dht/did.rs`, documented as the placement model with an algebraic law), and `entry_operate` stores at each of those `r` equidistant keys, each under its own owner (`crates/core/src/dht/chord.rs`). PR #648 just hardened exactly this: `PlacedEntryOperation::validate_placement` checks `placement ∈ entry_key.rotate_affine(redundancy)`.

These are different node sets. Successor-list replicas are r *adjacent* nodes; affine replicas are r *equidistant* points around the ring with generally distinct, non-adjacent owners (the design chooses equidistant placement precisely to avoid correlated adjacent-node failure). So the stated `Placement` invariant is false for the current system — affine placements are not, in general, in `Succ^{r-1}(owner(k))`.

This matters because the paper positions itself as describing the implementation ("Rings is implemented…", "the implementation realizes these laws"), and affine placement is not an unbuilt target that may legitimately differ — it is the current, shipping, just-hardened mechanism. The paper here is *behind* the code, describing a simpler replication scheme than what runs. The fix is small and the paper already has the machinery: define
\[ \mathrm{Rep}^r_N(k) = \{\, \mathrm{owner}_N(k + \lfloor 2^{160} i / r\rfloor) : i \in 0..r \,\}, \]
which is just `owner_N` applied to the affine offsets the code uses, and reword the `Placement` invariant accordingly. `Succ^r_N` remains correct for its actual role (successor-list resilience / topology maintenance in `Inv_ring`); it just should not be the *storage* replica set.

If instead the successor-list replication is a deliberate future target that supersedes affine placement, that is the one place in the paper where a target-vs-current note is genuinely needed (the issue's category-1/2 boundary) — say so explicitly, because a reader diffing against the code will otherwise read it as an error.

## Minor

- `@Stateright` in `cites.bib` is dated `year = {2026}` (future); the other 7 new entries (Zave CorrectChord, Mac Lane, Moggi, Shapiro CRDT, Lamport TLA+, MSRP RFC 4975) are correct and appropriate. Fix the year.
- Deleted algorithms `lamport ts`, `ElGamal SSSS`, `FR SSSS` — confirm no surviving `\ref{}` targets them (the validation notes checking `rings.log` for undefined references; worth an explicit grep since three labeled blocks were removed).
- `Assumption[System scope]` fixes `fault = fail-stop`, but the `Security predicates` table and ranking protocol reason about Sybil/MITM/misbehaving raters — i.e. Byzantine participants at the application/identity layer. That is a coherent split (crash-fault overlay, adversarial application layer), but a one-line note that the fail-stop assumption scopes the *overlay maintenance* model, not the security model, would prevent a reader from seeing the two as contradictory.
- Pre-existing `cites.bib` entries outside this PR's diff contain fabricated-looking citations (e.g. `amp` "Author-Merlin Protocol" attributed to Chaum/van Antwerpen; a `NymNetwork` journal that does not exist). Out of scope for this PR, but if any are still `\cite`d by the surviving text they are worth a follow-up cleanup pass.

## Verdict

The mathematical spine is careful and, on the points I cross-checked against the code (transition/effect model, biased order, owner/pred/succ, rotation equivariance, carrier separation, secret-sharing placement), accurate. The one real defect is the replica-set definition describing successor-list replication where the system does affine placement; it is a small edit with machinery already in the paper, but it should land because the document claims implementation fidelity. Everything #649 asked for is addressed.

PR #657, opened 2026-07-04
Add onion exit registry and route builder

RyanKung:
Refs #654

## Summary
- Add signed onion exit descriptors and an application-layer onion_exits DHT registry.
- Add route selection that combines online-node presence for relay candidates with the exit registry for exit candidates.
- Expose lookupOnionExits and buildOnionRoute JSON-RPC methods plus DTO/client support.

## Testing
- CARGO_HOME=/private/tmp/codex-cargo-home-onion-654 CARGO_TARGET_DIR=/private/tmp/codex-target-onion-654 cargo check -p rings-node --no-default-features --features node
- CARGO_HOME=/private/tmp/codex-cargo-home-onion-654 CARGO_TARGET_DIR=/private/tmp/codex-target-onion-654 cargo test -p rings-node onion

Note: cargo test -p rings-node --no-default-features --features node onion is blocked by the existing crates/node/src/tests/native/snark.rs cfg issue because it references crate::extension::snark while snark is disabled.


RyanKung:
## Review — findings ranked by severity

**1. `crates/node/src/onion.rs:506,574` — route selection is fully deterministic; a low-Did adversary captures every circuit (security)**

`order_peers_by_quality` is documented as a *stable permutation* of its input, and both inputs here iterate in ascending-Did order (`BTreeMap::keys()` / `BTreeSet`). Exits and relays are almost never directly measured peers, so they all classify as `PeerQuality::Unknown` — one quality class, ties everywhere. The result: **every client, on every call, selects the same exit and the same relays: the numerically smallest DIDs.** For an anonymity route builder this is fatal — an adversary grinds a small Did, publishes an exit descriptor plus an online descriptor, and becomes both the first hop and the exit of every circuit in the network. Route selection needs (weighted) random sampling over the eligible set; the deterministic ordering built for DHT connection scheduling is the wrong tool here.

**2. `crates/node/src/onion.rs:559` (constants at :44, :47) — every online node is conscripted as a relay; `ONION_RELAY_CAPABILITY` / `ONION_EXIT_CAPABILITY` are defined and exported but never checked (correctness)**

`eligible_relay_dids` filters only on network/local/exit. Since `default_advertise_presence()` is `true`, any ordinary node — including browser clients — becomes a relay candidate for other people's onion circuits without ever opting into forwarding anonymized traffic. Both capability constants are dead code as shipped. Relay eligibility should require `descriptor.capabilities` to contain `onion-relay` (and arguably exits should advertise `onion-exit`); otherwise the constants should not land in this PR.

**3. `crates/node/src/onion.rs:95-439` — ~300 lines copied nearly verbatim from `online.rs` / `registration.rs` (simplification)**

`validate_signer` / `body_ref` / `signing_data` / `new_signed` / `verify_signature` / `is_expired_at` / `is_live_at` / `latest_valid_by_did` / `Encoder` / `Decoder`, plus `OnionExitRegistration` mirroring `OnlineNodeRegistration` (`new` / `validate_enabled_schedule` / `descriptor_at` / `publish_descriptor` / `descriptors_from_entry` / the `RegistrationTask` impl) — even the test `latest_valid_by_did_filters_expired_and_keeps_newest` is duplicated; only the body field list differs. The module doc of `registration.rs` states the publisher exists "so new registries do not reimplement that state", yet this PR reimplements everything above the publisher. The signed envelope wants to be a generic `SignedDescriptor<B: DescriptorBody>` (did/heartbeat/expiry via the trait), and the registration task parameterized over body construction. With a second registry now copying the first, this is the moment to consolidate — before a third appears.

**4. `crates/node/src/processor.rs` (`build_onion_route`) — every descriptor's ECDSA signature is verified twice per call (efficiency)**

`lookup_online_nodes(false)` and `lookup_onion_exits(&service, false)` each already run `latest_valid_by_did` (per-descriptor signature verification + network/service filtering). `select_onion_route` then runs `eligible_exits` / `eligible_relay_dids`, which repeat exactly the same validation and filtering over the same data (with a second, later `get_epoch_ms()`). One RPC call performs 2n elliptic-curve verifications. `select_onion_route` should accept already-validated sets (distinguished by type, e.g. a `Validated<T>` newtype), or be the sole owner of validation.

**5. `crates/node/src/processor.rs` (`onion_route_peer_qualities`) — duplicates the thresholds+classify pipeline of `PeriodicMeasure::quality` (reuse)**

`PeerQualityThresholds::new(CONNECT_FAILED_LIMIT, MSG_SEND_FAILED_LIMIT, MSG_RECV_FAILED_LIMIT)` + `classify` already exists at `crates/node/src/measure.rs:216-224`. The node's quality thresholds now have two sources of truth; extract a single `pub(crate)` constructor and use it from both sites.

**6. No production path publishes exit descriptors (scope/completeness)**

`OnionExitRegistration` is never constructed outside tests in this PR: there is no config plumbing (unlike the online registry's `advertise_presence` / heartbeat / ttl), and the only entry point is the library-level `ProcessorBuilder::registration_task`. A node running the shipped binaries cannot become an exit, so the registry is permanently empty in production. If this is a deliberate slice of #654, the PR description should say so; otherwise `validate_enabled_schedule` and friends are unreachable public API.

**7. `crates/rpc/src/jsonrpc.rs:49` — drive-by `where`-clause reformat violates `rustfmt.toml`'s `where_single_line = true` (conventions)**

The deleted single-line form is exactly what nightly rustfmt produces under this repo's config; the new form is stable-rustfmt default style. `cargo +nightly fmt` will revert it and the fmt CI gate will fail. This unrelated hunk should be dropped.

**8. `crates/node/src/onion.rs:533` — the `has_duplicate_dids` check is unreachable by construction (vacuous invariant)**

Relays come from a `BTreeSet` that already excludes `local` and `exit_did` (elements pairwise distinct), and `exit_did` is appended exactly once — a duplicate cannot occur. This runtime error branch is dead code. The invariant holds by construction; if it should be documented, that is a comment plus `debug_assert!`, not an `Err` that can never fire.

Minor, same pattern as 8: `BuildOnionRouteResponse.exit` is an `Option` that is always `Some`, and `exits_by_did.remove(...).ok_or_else(…"disappeared during route selection")` in `select_onion_route` is another by-construction-unreachable error branch.

Findings 1 and 2 go to the heart of what this feature claims to be (anonymity, and consent to relay); they deserve a decision first. The consolidation direction in 3 also sets the cost of every future registry.

RyanKung:
## Re-review (commits c458f63e, 33eb334c, 00d6c86e)

Full re-read of the revised branch. The previous round's items are addressed: weighted random selection (`RouteEntropy` / `pick_weighted_index` / `quality_weight`), relay eligibility gated on `ONION_RELAY_CAPABILITY`, a shared `SignedDescriptor` trait in `descriptor.rs`, a single `peer_quality_thresholds()` source, full config + CLI plumbing with an `advertise_onion_relay ⇒ advertise_presence` check, the reverted `where`-clause hunk, and `debug_assert!` for the by-construction dedup invariant. Good.

New findings from this read, most severe first:

**1. `crates/node/src/onion.rs:975` — browser path-selection RNG is not cryptographic (security)**

`RouteEntropy for SystemRouteEntropy` uses `js_sys::Math::random()` on the browser arm, while the native arm uses `rand::random()` (ThreadRng, a ChaCha CSPRNG). `Math.random()` is explicitly not a CSPRNG — its output is predictable/reproducible to an observer who can model the engine's PRNG state. For an onion route builder, predictable relay/exit selection is a deanonymization vector: an adversary who can predict which relays and exit a browser client picks can position itself on the path. This PR *already* adds `rand` with the `getrandom` feature, and the `getrandom` 0.2 js backend is present in the lock graph, so the browser arm can and should draw from a CSPRNG (`getrandom` / `crypto.getRandomValues`) instead of `Math.random()`. The native/browser asymmetry makes this an easy, contained fix.

**2. `crates/node/src/provider/browser/provider.rs:2507` + `onion.rs:506` — browser auto-exit ships an empty, unenforced "closed" policy (security/design)**

`Provider::new_https_exit_instance` → `enable_https_onion_exit` publishes an exit descriptor with `default_onion_exit_policy()` = `OnionExitPolicy::default()`: empty `allowed_targets`, empty `denied_targets`, all limits `0` ("unspecified"). `validate_onion_role_config` only requires `onion_exit_services` to be non-empty — it never requires an exit to declare any target policy. The doc comment claims the default is "intentionally closed until the operator configures targets," but nothing enforces closure; the meaning of an empty allow-list is deferred to a not-yet-existing exit implementation. When enforcement lands, a browser using this constructor becomes an open HTTPS proxy if empty-allow is later read as allow-all. Recommend either enforcing it now (require a non-empty `allowed_targets` when `advertise_onion_exit` is set) or making the closed-by-default semantics explicit and testable rather than a comment.

**3. `crates/node/src/processor.rs:657` — `ONION_EXIT_CAPABILITY` is advertised but never consumed (dead wiring)**

The builder pushes `ONION_EXIT_CAPABILITY` into the online-node descriptor's capabilities when `advertise_onion_exit` is set, but no selector reads it: exit discovery uses the dedicated `onion_exits` registry, and relay eligibility (`has_onion_relay_capability`) checks only `ONION_RELAY_CAPABILITY`. It is doubly inert when `advertise_onion_exit` is set with `advertise_presence` false, since then the online descriptor carrying the label is never published. Either wire a consumer or drop the capability from the online descriptor.

**4. `crates/node/src/onion.rs:578-651` — the signing envelope is still duplicated from `online.rs` (simplification, residual of prior finding)**

`SignedDescriptor` now shares the read side (verify / expiry / encode / decode), but `OnionExitDescriptorBody` / `body_ref` / `signing_data` / `new_signed` and the `OnionExitDescriptorBodyRef<'a>` borrow-struct remain copied field-for-field from their `OnlineNode*` twins (~130 lines). This is exactly the construct/sign half a body trait + generic `new_signed` would remove; the extraction captured verification but not signing.

**5. `crates/node/src/onion.rs:996` — `from_validated_descriptors` re-applies the network filter `lookup_*` already applied (efficiency, minor)**

`lookup_online_nodes` / `lookup_onion_exits` already filter by `matches_network`; `OnionRouteCandidates::from_validated_descriptors` filters by network again (the capability / service / local filters it adds are genuinely new). Harmless but redundant — drop the network predicate from the candidate builder, or make the builder its sole owner.

Items 1 and 2 are the substantive ones — both go to the anonymity/consent properties the feature is meant to provide.

RyanKung:
## Re-review (commit 3c9b7ed4 "Advertise default TCP and HTTPS exits")

Only one commit landed since the last round. It changes `default_onion_exit_services()` from `[https]` to `[https, tcp]` and adds `OnionExitService::tcp()`. None of the five findings from the previous re-review were addressed — items 1 (non-CSPRNG browser path RNG), 2 (empty/unenforced exit policy), 3 (unconsumed `ONION_EXIT_CAPABILITY`), 4 (residual signing-envelope duplication), and 5 (redundant network filter) remain open. This commit also introduces a new issue.

**New — `crates/node/src/onion.rs:76` + `crates/node/src/provider/browser/provider.rs` — browser exit now advertises a native TCP service it cannot serve (correctness)**

`Provider::new_https_exit_instance` (the browser constructor) → `ProcessorConfig::enable_https_onion_exit` now sets `onion_exit_services = default_onion_exit_services() = [https, tcp]`. A browser node publishes its descriptor with `node_type = Browser`, but `OnionExitTransport::Tcp` is documented as "Native TCP service" — a browser runtime cannot open native TCP sockets. So a browser exit advertises a `tcp` service it fundamentally cannot fulfill, and route selection (`offers_service("tcp")`) will happily pick that browser as a `tcp` exit. There is no transport/runtime compatibility check anywhere between the advertised transport and the publisher's `node_type`. The default TCP service belongs to native exits only; the browser constructor should keep advertising HTTPS (browser-fulfillable) and not the native TCP service.

**New — same commit — `enable_https_onion_exit` is now misnamed (naming)**

Both `ProcessorConfig::enable_https_onion_exit` and `ProcessorConfigSerialized::enable_https_onion_exit` now enable HTTPS **and** TCP; the doc comments were updated but the method name still says only HTTPS. The name should reflect what it does (e.g. `enable_default_onion_exit`), or the browser path should have its own HTTPS-only enabler distinct from the native default.

The two are related: if the browser constructor stops pulling in the native TCP default, the "https" name becomes accurate again for the browser path and the runtime/transport mismatch disappears.

RyanKung:
## Open items — consolidated status (as of 3c9b7ed4)

Tracking what is still open across the review rounds. The original 8 findings were addressed in `c458f63e` (weighted random selection, relay-capability gating, the shared `SignedDescriptor` trait, single `peer_quality_thresholds()`, config/CLI plumbing, reverted `where` hunk, `debug_assert!`). The items below remain.

### Correctness / security

- **Browser path-selection RNG is not cryptographic.** `onion.rs` `RouteEntropy for SystemRouteEntropy` uses `js_sys::Math::random()` on wasm while native uses `rand::random()` (ChaCha CSPRNG). Predictable relay/exit selection is a deanonymization vector. The `rand`/`getrandom` js backend is already in the graph — draw from a CSPRNG (`getrandom` / `crypto.getRandomValues`) on the browser arm.
- **Browser exit advertises a native TCP service it cannot serve.** After `3c9b7ed4`, `new_https_exit_instance` → `enable_https_onion_exit` advertises `[https, tcp]`, but a `node_type = Browser` node cannot open native TCP sockets (`OnionExitTransport::Tcp` = "Native TCP service"). `offers_service("tcp")` will still select it. No transport/`node_type` compatibility check exists. Browser should advertise HTTPS only.
- **Auto-exit ships an empty, unenforced "closed" policy.** `default_onion_exit_policy()` is `OnionExitPolicy::default()` (empty allow/deny, limits `0`), and `validate_onion_role_config` only checks that services is non-empty. The "intentionally closed" claim is a comment with nothing enforcing it; when exit enforcement lands, an empty allow-list read as allow-all makes the browser constructor an open proxy. Enforce a non-empty `allowed_targets` when `advertise_onion_exit` is set, or make closed-by-default explicit and tested.

### Cleanup / conventions

- **`ONION_EXIT_CAPABILITY` is advertised but never consumed.** Pushed into the online-node descriptor when `advertise_onion_exit` is set, but no selector reads it (exit discovery goes through the `onion_exits` registry; relay eligibility checks only `ONION_RELAY_CAPABILITY`). Doubly inert when `advertise_presence` is false. Wire a consumer or drop it.
- **`enable_https_onion_exit` is misnamed.** It now enables HTTPS **and** TCP on both `ProcessorConfig` and `ProcessorConfigSerialized`; only the docs were updated. Rename to reflect behavior, or give the browser path an HTTPS-only enabler.
- **Signing envelope still duplicated between `online.rs` and `onion.rs`.** `SignedDescriptor` shares the read side; `*DescriptorBody` / `body_ref` / `signing_data` / `new_signed` and the `*BodyRef<'a>` borrow-struct are still copied field-for-field (~130 lines). A body trait + generic `new_signed` removes the construct/sign half too.
- **Redundant network filter (minor).** `lookup_online_nodes` / `lookup_onion_exits` already filter by `matches_network`; `OnionRouteCandidates::from_validated_descriptors` filters by network again. Drop it from the candidate builder or make the builder its sole owner.

Priorities: the first three (the two browser items + the policy default) go to the anonymity/consent properties the feature is meant to provide; the rest are cleanup.

RyanKung:
## Re-review (commits cabaaa5a "Address onion exit review feedback", 5fa31edb "Add onion proxy client ingress")

### Open items now resolved (cabaaa5a)

- **Signing-envelope duplication** — new `SignedDescriptorBody` trait + `sign_descriptor_body` helper; both `online.rs` and `onion.rs` delegate `new_signed` to it and drop their `validate_signer` copies. Resolved.
- **`ONION_EXIT_CAPABILITY` dead wiring** — const, prelude re-export, and builder push all removed. Resolved.
- **Redundant network filter** — `from_validated_descriptors` no longer takes/repeats `network_id`. Resolved.
- **`enable_https_onion_exit` misnamed / browser TCP mismatch** — split into HTTPS-only `enable_https_onion_exit` and `enable_default_onion_exit` (https+tcp); the browser constructor now advertises HTTPS only, so a browser no longer advertises a native TCP service it can't serve. Resolved.
- **Browser path-selection RNG** — the `Math.random()` arm is gone; `SystemRouteEntropy` now uses `rand::random()` on both targets (CSPRNG via the getrandom js backend on wasm). Resolved.
- **Empty/unenforced exit policy** — `OnionExitPolicy::is_closed()` / `allows_target()` added with closed-by-default semantics and docs. See new finding #2: the checker isn't actually on the exit data path yet.

### New findings (5fa31edb)

**1. `crates/node/src/extension/transport/engine/mod.rs:206` — `relay_tcp_stream` evicts the pending stream before `bind_accepted` can consume it; the native onion HTTP proxy relays nothing (correctness, high)**

```rust
let Some(token) = self.stash_pending(Pending::Tcp(stream)) else { return; };
inject_accepted(&scope, token, peer, service).await;
self.evict_pending(token);   // ← removes the stash immediately
```

`inject_accepted` only enqueues an `Accepted` command (`scope.inject(..).await`); the `Accepted → OpenAccepted → bind_accepted` round-trip through the core runs asynchronously afterwards. `bind_accepted` is what takes the stashed stream (`pending.remove(&token)`) and spawns `tcp::relay_tcp`. But `relay_tcp_stream` calls `evict_pending(token)` synchronously right after the inject returns — before that round-trip completes — so the stream is removed first and `bind_accepted` hits its `// token already consumed` early-return and drops the `TcpStream`. The listener paths (`listen_tcp`/`listen_udp`) never evict on the success path for exactly this reason; `evict_pending` is documented as a failure-path safeguard ("A no-op once `bind_accepted` has consumed the token"). Net effect: `handle_connect` writes `200 Connection Established`, then no bytes are ever relayed — every tunnel is silently dead, i.e. the whole new ingress is non-functional. Drop the unconditional `evict_pending` here and let `bind_accepted` consume the token, matching the listener paths.

**2. `crates/node/src/native/onion_http_proxy.rs:62` — the native exit ignores `denied_targets`; the deny-list is never enforced (security)**

`register_native_onion_exit_targets` iterates `policy.allowed_targets` and registers a relay TCP service for each, never consulting `denied_targets` or calling `OnionExitPolicy::allows_target`. But `allows_target` documents (and unit-tests) "Deny entries override allows." A target listed in both allow and deny is still registered and served by the exit, contradicting the policy's own contract. Relatedly, `allows_target` / `is_closed` are currently exercised only by their unit tests — no production path calls them — so the deny-aware checker added in cabaaa5a is effectively dead relative to the exit that actually serves traffic. Filter `allowed_targets` through `allows_target` (or otherwise subtract `denied_targets`) at registration.

### Minor

- `onion_http_proxy.rs:read_http_header` reads one byte per `stream.read` syscall up to 8 KiB — a bounded but avoidable syscall-per-byte loop; a small buffered read would do.
- `onion_http_proxy.rs:resolve_target` resolves each target once at startup and pins `.next()` (first address only) — no re-resolution on DNS change and no failover to later addresses. Acceptable for a first cut, worth a note.

Finding #1 is the blocker — as written the native HTTP CONNECT proxy establishes tunnels that carry no data.

RyanKung:
## Re-review (commit 01c2bb19 "Fix onion proxy relay stream binding")

### Correction — my previous finding #1 was a false positive

I need to retract the "high severity / release blocker" claim I made last round about `relay_tcp_stream` evicting the pending stream before `bind_accepted` consumes it. **The original code was correct; the native onion HTTP proxy would have relayed data.** Here is what I missed:

`inject_accepted(..).await` calls `Scope::inject` → `Core::inject` → `Core::dispatch`, and `dispatch` drives the fixpoint loop **synchronously and awaited** to completion. Within it, `Runner::handle` runs the pure `step`, then `for effect in effects { self.interpret.run(&scope, effect).await? }` — so the `OpenAccepted` effect, and therefore `bind_accepted` (which does `pending.remove(&token)` and spawns `relay_tcp`), executes **inside** the `inject_accepted().await`. By the time that await returns, the token is already consumed, so the subsequent `evict_pending(token)` was a **no-op** on the success path. This is the same pattern the pre-existing `listen_tcp` / `listen_udp` accept loops have always used (and which `tcp_relay_round_trip` covers) — I incorrectly claimed those paths didn't evict on success; they did, and it was fine for the same reason. Apologies for the false alarm.

So commit 01c2bb19's change to `relay_tcp_stream` (and the matching changes to `tcp.rs` / `udp.rs`) is a reasonable robustness/clarity improvement — `inject_accepted` now returns `Result` and eviction happens only when the inject itself fails — but it was not fixing a functional defect. The added `tcp_stream_relay_round_trip` example test is a good end-to-end guard for this path either way.

One small note on the new shape (non-blocking): eviction now runs only when `inject_accepted` returns `Err`. If `dispatch` ever returned `Ok` without binding the token (e.g. a decode-reject drop, which returns `Ok(Vec::new())`), the stashed stream would leak, whereas the previous unconditional evict cleaned it up. For the relay's own self-injected `Accepted` command this drop path isn't reachable, so it's fine — just worth being aware that "evict only on Err" leans on the inject always either binding or erroring.

### Previous finding #2 — resolved

The deny-list gap is properly fixed: `register_native_onion_exit_targets` now goes through `native_onion_exit_target_authorities`, which filters `allowed_targets` through `policy.allows_target(..)`, so `denied_targets` is honored (and the empty-allow "closed" default still registers nothing). The `native_exit_target_registration_honors_deny_list` test covers it, including host normalization. Good.

### Net

With this commit, all items I had open are addressed. I don't see new issues in 01c2bb19. The substantive prior concerns (weighted-random path selection, relay opt-in, browser CSPRNG, browser-transport mismatch, exit policy enforcement, the shared descriptor envelope) are all resolved; remaining differences are stylistic/forward-looking (e.g. DNS re-resolution at the exit, byte-at-a-time CONNECT header read) rather than correctness.

RyanKung:
## Re-review (5eac68c5 … 92d02371 — target-agnostic config, browser HTTPS exit + native circuit data plane, boundary hardening, abstraction cleanup)

This round adds the actual data plane: `onion/circuit.rs` (routed frames), `onion/tcp.rs` (native TCP exit/client), `onion_https.rs` (browser HTTPS exit), plus the `onion.rs` → `onion/` module split. I read the new data-plane files in full. Two things I want to flag prominently, then what checks out.

### 1. The circuit relay forwards for every node regardless of `advertise_onion_relay`, with no hop cap and no per-relay rate limit (consent bypass + DoS)

`bin/rings.rs:686` installs the circuit protocol unconditionally:

```rust
let onion = NativeOnionCircuitHandle::install(
    &provider.extensions(),
    advertise_onion_exit.then_some(onion_exit_policy.clone()),
)?;
```

Only the **exit** role is gated (`exit_policy` is `Some` only when `advertise_onion_exit`). The **relay/forward** path is not gated at all: `OnionCircuitProtocol::step` (circuit.rs) is a stateless reducer that, for any `Forward` frame with non-empty `remaining_hops`, emits `Send { to: next_hop }` — for anyone, to any DID, over the overlay. So:

- A node that never set `advertise_onion_relay` still forwards onion circuit frames for arbitrary senders. The opt-in we added in earlier rounds only governs route *selection* (who gets picked); the forwarding data plane ignores it, so the consent it represents is not actually enforced.
- `remaining_hops` is caller-controlled with no maximum, and the relay path has no limiter (only the exit has `TcpExitLimiter` / the HTTPS limiter). One crafted frame with a long `remaining_hops` list turns any set of nodes into unmetered relays — a free-relaying / circuit-flooding primitive.

Suggest gating the forward reducer (or the `install`) on `advertise_onion_relay`, enforcing a max hop count on receipt, and giving relays a per-sender/per-circuit rate limit comparable to the exit's.

### 2. This is not onion encryption — relays and the exit see everything in clear text

`circuit.rs` is explicit and honest about it:

> Security model: this wire format is a routed circuit frame, not layered onion encryption. Relays receive the remaining route in clear text … A cryptographic onion layer needs a descriptor-published relay encryption public key and per-hop decrypt/encrypt logic before this protocol can provide onion confidentiality.

Concretely, every relay sees the full `remaining_hops`, the full `return_path`, and the application `payload` (`TcpOpen{target}`, raw `TcpData`, the HTTPS request/response), and the exit sees the plaintext target and bytes. So the current feature provides *routing* but essentially none of the confidentiality/anonymity that "onion" implies — a single on-path relay deanonymizes the whole circuit and reads the traffic. I'd raise this to the API/config/README surface (and ideally the method/flag names), not leave it only in a module comment, so operators don't enable it believing it gives onion properties. Whether to ship the routing layer ahead of the crypto layer is a maintainer call, but it should be a loud, deliberate one.

### 3. Exit data frames don't pin the adjacent relay (minor, hardening)

`onion/tcp.rs` `handle_exit` ignores `_from`: after `TcpOpen`, subsequent `TcpData`/`TcpShutdown`/`TcpClose` are keyed only by `(return_path_client, stream_id)` and injected into the exit stream without checking the immediate sender. The client side does the symmetric check (`client_inbound_sender` rejects a `from != expected_return_peer`). It's mitigated by the 64-bit random `stream_id`, but the exit could pin the adjacent relay at `TcpOpen` and reject mismatches for parity.

### What checks out

- **HTTPS exit SSRF is properly gated.** `execute_exit_fetch` rejects `!policy.allows_target(&authority)`, builds the fetch URL from the *same* checked `authority` with a hardcoded `https://` scheme, `normalize_path` forces a leading `/` or `?` (so the path can't relocate the host), and body size is bounded (content-length pre-check + streamed cap).
- **Native TCP exit** enforces `allows_target`, per-circuit stream/circuit limits (`admit_exit_stream` + `TcpExitLease` Drop accounting), and `max_bytes_per_minute` on both read and write paths.
- **Client ingress authenticates the return peer** (`expected_return_peer == from`), and `validate_exit_return_path` binds the exit's first return hop to the authenticated sender.
- Module split (`onion/{mod,circuit,route,target,tcp,tests}`) and the `admit`/`lease` accounting are clean and well-tested.

Findings 1 and 2 are the ones worth a decision before this lands; 3 is polish. (No repeat of last round's mistake here — the forwarding gap is verified against the unconditional `install` and the stateless forward reducer, not inferred.)

RyanKung:
## Re-review (b4538d5f "Encrypt onion circuits with ElGamal AEAD", 769a13d2, c5f6b5cd)

### Previous findings #1 and #2 — resolved, and thoroughly

- **#2 (no layered encryption)** is now real onion crypto. `circuit.rs` is `OnionCircuitSecurity::LayeredAead`: `build_forward_layers` wraps exit→entry, each layer AEAD-sealed (ElGamal-KEM over secp256k1 wrapping ChaCha20-Poly1305, new `crates/core/.../elgamal/impls/secp256k1.rs`) to that hop's `session_public_key`, now published in the online/exit descriptors. A relay decrypts exactly one layer and learns only the immediate `next_hop` + an opaque `inner`; the backward payload is AEAD-encrypted to the client's session key and relays forward it via local `relay_returns` state, so the plaintext route/return-path/payload exposure from last round is gone. The AEAD `aad` binds `namespace + direction + circuit_id`, which correctly prevents cross-circuit layer replay and forward/backward confusion. The KEM/DEM construction itself looks sound (fresh random content key + nonce per seal, wrapped key bound into the AAD, tamper/AAD/empty-key rejection tests present).
- **#1 (relay consent + DoS)** is addressed: forwarding is gated by `allow_relay` (`validate_relay_forward` returns `NoPermission` when off), wired from `advertise_onion_relay` in both the native (`bin/rings.rs`) and browser (`install_onion_https_protocol`) install paths; `MAX_ONION_CIRCUIT_HOPS = 8` caps hop count on both build and receipt; `MAX_ONION_RELAY_CIRCUITS = 1024` bounds relay state.

Nice work — this is a big step up.

### New finding 1 — layer decryption runs inside the pure `step`, under the per-protocol `std::Mutex`, before the `allow_relay` gate (architecture + DoS)

`OnionCircuitProtocol::step` → `advance_forward`/`advance_backward` calls `decrypt_forward_layer` / `decrypt_client_payload` (an ElGamal scalar-mult + AEAD open) directly in the reducer. In `registry.rs::handle`, `step` runs while holding `self.state` (`std::sync::Mutex`, documented as a "brief synchronous critical section"). Two problems:

- **Entanglement / altitude.** A secret-key decryption is a keyed effect; putting it in the pure `step` (which also now carries `session_sk`) mixes an effect into the functional core. `decode` runs *outside* the state lock and already has `&self` (so it can reach `session_sk`) — decrypting there, or in the interpreter, keeps `step` pure and unlocks the critical section.
- **CPU-exhaustion / head-of-line DoS.** Any peer can force a node to perform an asymmetric decryption by sending a forward frame sealed to that node's *published* `session_public_key`, and the decrypt happens **before** the `allow_relay` check — so even relay-disabled nodes pay the cost. Because it runs under the single per-protocol `std::Mutex`, a flood of cheap-to-send frames serializes all circuit processing and pins a core. Backward frames with an unknown `(circuit_id, from)` fall through to a client-payload decrypt attempt too, widening the same surface. Consider a pre-decrypt cheap check / per-sender rate limit, and moving the crypto out of the locked region.

### New finding 2 — relay return table has a hard cap but no expiry (state-exhaustion DoS)

`remember_return_hop` bounds `relay_returns` at 1024 and only removes an entry on a `terminal` backward frame. An attacker (against a relay-enabled node) can send 1024 forward frames with distinct attacker-chosen `(circuit_id, next_hop)` and never send a terminal frame; the table fills permanently and every subsequent real circuit is rejected with "onion relay circuit table is full". There's no TTL or LRU eviction, so relay service is denied network-wide until restart. Tor-style circuit timeouts (evict stale `relay_returns` after an idle window) or a per-sender quota would close this.

### Notes / minor

- No replay protection on forward frames: an on-path party can re-inject a forward frame, and the relay will decrypt+forward it again (duplicate `TcpData`/`TcpOpen` at the exit). Mitigated if the overlay transport dedups, but the circuit layer itself has no per-frame nonce/sequence.
- `OnionCircuitProtocol` derives `Debug` while holding `SessionSk`; make sure `SessionSk`'s `Debug` redacts the secret so a `{:?}` on the protocol can't log key material.

New finding 1 is the one I'd prioritize. (Lock scope verified against `registry.rs::handle` — `step` runs under `self.state.lock()`, `decode` does not.)

RyanKung:
## Re-review (64030b67 … 4e995032 — boundary refactor, ElGamal-AEAD hardening, stream identity, replay, rustfmt)

Big refactor: `circuit.rs` split into `circuit/{protocol,reducer,shell,codec,crypto,limiter}` plus new `onion/{replay,failure,exit_accounting,directory}` and `tcp/{exit,duplex,inbound,target}`. I re-read the security-critical paths.

### Previous findings #1 and #2 — resolved, and well

- **Decryption out of the locked `step`.** The reducer is now genuinely pure: it emits `DecryptForward` / `DecryptClient` / `TimestampBackward` *effects*, and the crypto runs in `OnionCircuitShell` (`Interpret::run`), which executes outside the `std::Mutex` state critical section. The former "heavy ECC under the state lock" is gone.
- **CPU-exhaustion gate.** `OnionCryptoGate` / `OnionCryptoLimiter` rate-limits decrypts per authenticated `from` (windowed, self-purging), and `OnionCircuitCapabilities::accepts_forward_layers()` makes client-only nodes reject forward frames *before* any decrypt. Together these close the unauthenticated decrypt-flood.
- **Local re-injection is safe.** `decode_event` only accepts the `OnionLocalMessage` channel (`ForwardReady` carrying a plaintext layer) when `wire.from == wire.me`; a remote peer's frame always goes through `decode_wire_message` (encrypted `Forward`/`Backward` only), so an attacker can't forge a pre-decrypted layer to skip crypto or spoof routing.
- **Relay-table TTL.** `RelayReturnEntry.expires_at_ms` + `purge_expired_return_hops` (before capacity checks and before backward lookup) + refresh-on-use fix the state-exhaustion DoS, and `ReturnEdgeConflict` prevents rewriting a live edge.
- Bonus hardening I noticed: per-edge `next_circuit_id` (unlinkable across hops), a bounded `replay.rs` nonce cache with forward/backward keys, Exit-layer `expires_at_ms` (`ForwardPayloadExpired`), and an authenticated backward path (`BackwardSignerMismatch` / `InvalidBackwardSignature` against the selected exit's keys). This is a strong revision.

### New — `crates/core/src/message/protocols/verify.rs` is a system-wide auth-semantics change bundled into the onion PR

The refactor to `is_live_at` / `verify_unexpired` also **changes acceptance for every signed message in rings**, not just onion:

- Old `is_expired` returned `false` (i.e. "not expired", so `verify` accepted) when `ttl_ms > MAX_TTL_MS` *or* when the timestamp was in the future beyond tolerance. New `is_live_at` requires `ttl_ms <= MAX_TTL_MS` **and** `ts_ms - TS_OFFSET_TOLERANCE_MS <= now`, so both cases are now rejected.

The over-`MAX_TTL_MS` case is a genuine latent hole being closed (previously a valid signature with an absurd TTL passed the freshness gate), so the direction is right. But two things:

1. **Scope / blast radius.** This alters core message verification for all protocols and belongs in its own change (or at least an explicit call-out), not silently inside the onion feature — it's exactly the kind of behavior-boundary change worth isolating and sweeping docs for.
2. **Interop headroom.** `TS_OFFSET_TOLERANCE_MS` is only 3000 ms. Now that future timestamps beyond tolerance are strictly rejected (they were accepted before), any peer whose wall clock runs >3 s fast will have *all* its signed messages dropped. 3 s is tight for arbitrary internet peers with imperfect NTP; worth confirming that's acceptable, or widening the tolerance, before this rides out system-wide.

Please confirm nothing legitimately emits `ttl_ms > MAX_TTL_MS` and that the 3 s future skew is intentional under strict enforcement.

### Net

Both prior blockers are resolved and the crypto/boundary work is high quality. The one thing I'd act on is the bundled `verify.rs` semantics change — split it out or document it, and sanity-check the 3 s skew tolerance. No other new correctness issues surfaced in this pass. (Verified: shell decrypt runs outside `registry.rs`'s state lock; local-message channel gated on `from == me`.)

RyanKung:
## Re-review — onion routing correctness & completeness (d4252646 … c2663793)

This round is mostly reorg (module/test splits) plus two functional fixes (`OnionServiceName` canonicalization, native TCP exit service binding). I focused on end-to-end correctness/completeness of the routing itself.

### Correctness — the core path checks out

I traced the whole selection → layering → return path:

- **Selection.** `build_onion_proxy_route` (the traffic path) filters exits by service *name* → transport (`offers_service_transport`, `NoExitWithTransport`) → policy (`allows_target`, `NoExitAllowsTarget`), then weighted-samples exit + relays. `validate_route_hops` enforces `hops == encryption_hops.map(did)`, no duplicate DID, last hop == exit descriptor (DID **and** session key), and `exit.offers_service`.
- **Layering.** `build_forward_layers_with_ids` nests correctly: the AAD `circuit_id` each layer is sealed under equals the wire `circuit_id` that hop will receive (`circuit_ids[index]`), and the relay's remembered return key `(next_circuit_id, next_hop)` matches what the backward frame carries — so forward routing, per-edge unlinkable ids, and reverse routing are all consistent.
- **Backward authentication.** `OnionAuthenticatedPayload::into_verified_payload` binds the response to the selected exit with five checks (return_id, exit account DID, account pubkey, session DID, and a signature over `return_id‖nonce‖exit_session_pubkey‖payload`), so a relay can't forge exit responses. Forward `expires_at_ms` + forward/backward replay nonces are all wired.

Service consistency is enforced at both ends (`validate_route_payload_service` on encode, `PayloadServiceMismatch` at the exit). This is a complete, coherent implementation.

### Completeness gap 1 — `allow_short_paths` can degrade to a 0-relay circuit (no anonymity)

`select_onion_route_from_candidates` sets `relay_hops_needed = target_hop_count - 1` and, when `allow_short_paths` is set and relays are scarce (or when `hop_count == 1`), happily returns `encryption_hops = [exit]`. `validate_route_hops` accepts any length in `1..=MAX`, so a single-hop "circuit" is valid: the client connects straight to the exit, which then sees `from = client`. That silently removes the entire onion property — the exit learns the client. For an anonymity system I'd enforce a floor of at least one relay (≥2 hops) even under `allow_short_paths`, or surface the degradation rather than returning a route that looks like a circuit but isn't one.

### Completeness gap 2 — the authenticated backward path is fragile to >3 s clock skew

`into_verified_payload` accepts a response only if `authentication.verify_unexpired(...)` passes, and after the `verify.rs` tightening that requires `ts_ms - TS_OFFSET_TOLERANCE_MS <= now` with `TS_OFFSET_TOLERANCE_MS = 3000`. The exit signs each backward payload with its own wall clock. So if the exit's clock leads the client's by more than 3 s, **every** response fails `InvalidBackwardSignature` and the circuit delivers nothing — a silent, total failure with no fallback, between two otherwise-healthy peers. 3 s is tight for per-frame freshness on the hot data path of a globally distributed overlay (this is the concrete data-path consequence of the `verify.rs` note from last round). Consider a wider tolerance for the onion backward transcript, or decoupling its freshness window from the 3 s message-skew constant.

### Net

The onion routing is now correct and largely complete end-to-end. The two items above are completeness/robustness gaps rather than crypto or routing bugs: gap 1 lets an "onion" silently collapse to a direct client→exit link, and gap 2 makes the data path brittle under modest clock skew. Neither blocks the crypto/routing core. (Verified: layer/edge `circuit_id` consistency and the relay return-key match by tracing `build_forward_layers_with_ids` against `reducer::advance_forward`/`advance_backward`.)

PR #660, opened 2026-07-06
Implement storage virtual nodes

RyanKung:
## Summary

- add storage-only Chord virtual positions derived from `network_id`, owner DID, and virtual index
- route storage lookup, operate, repair, and sync ownership through virtual storage owners when enabled
- expose `dht_virtual_nodes` through swarm, processor, and native config, defaulting to disabled
- derive the active virtual-owner set from the local DHT topology view instead of a separate mutable registry

## Notes

This implements the ownership-only virtual-node path discussed in #659. Virtual positions are not signing identities; storage and transport still use the physical owner DID.

## Validation

- `git diff --check`
- `CARGO_TARGET_DIR=/private/tmp/rings-virtual-nodes-659-target cargo test -p rings-core dht::virtual_node --lib`
- `CARGO_TARGET_DIR=/private/tmp/rings-virtual-nodes-659-target cargo test -p rings-core dht::chord::storage_tests --lib`
- `CARGO_TARGET_DIR=/private/tmp/rings-virtual-nodes-659-target cargo check -p rings-node --no-default-features --features node`


crates/core/src/message/handlers/storage/tests/test_storage_sync_report.rs
```diff
@@ -0,0 +1,268 @@
+use std::sync::Arc;
```
RyanKung, line 1: this test file should not place here

PR #664, opened 2026-07-08
Add self-authenticating .rings names

### LINKED ISSUE #662: RFC: Add .rings self-authenticating service names over Chord
## Context

We want Rings to support a Chord-backed name system under `.rings`, with semantics closer to `.onion` than public DNS. The `.rings` suffix should be a Rings overlay pseudo-TLD intercepted by Rings resolvers, not a dependency on the public DNS root.

The repository already has the substrate pieces:

- DHT storage can store and fetch topic-addressed `EntryKind::Data` values.
- `registerService` / `lookupService` currently provide a weak `name -> Vec<DID string>` mapping.
- `online_nodes` and `onion_exits` already demonstrate signed descriptor registries with heartbeat and expiry.
- Onion route construction can select live relays and exits, but it does not yet resolve a hidden service name into an authenticated service descriptor.

The missing layer is an authenticated Rings name protocol.

## Goal

Add an RFC and implementation plan for `.rings` self-authenticating service names over Chord.

The MVP should treat `.rings` as an overlay namespace:

```text
<self-authenticating-label>.rings
  -> DHT fetch signed RingsNameRecord
  -> verify owner/signature/expiry/sequence/network
  -> resolve target service descriptor
  -> build route/connect through the Rings onion/overlay data plane
```

## Proposed MVP scope

- Define `.rings` as a pseudo-TLD handled only by Rings resolver code.
- Define canonicalization and validation for `.rings` names.
- Prefer self-authenticating labels for the first version, derived from the service owner key or service key material.
- Add a signed `RingsNameRecordV1` schema with at least:
  - schema version
  - canonical `.rings` name
  - owner verification public key
  - target DID and/or hidden-service introduction points
  - session public key or service encryption key material needed by route setup
  - service name and transport metadata
  - `network_id`
  - monotonic `seq`
  - `ttl` / `expires_at_ms`
  - signature over the canonical record body
- Store records under a domain-separated DHT key, for example:

```text
rings-name:v1:<network_id>:<canonical .rings name>
```

- Avoid global registry scans for name resolution. A resolver should fetch the DHT key for the requested name directly.
- Add resolver APIs for publish and resolve, separate from the current weak `registerService` / `lookupService` interface.
- Add cache and negative-cache behavior bounded by record TTL.
- Integrate `.rings` resolution with onion route construction so a client can connect to the resolved service without treating it as a public Internet exit target.
- Add browser/native API surfaces so `*.rings` can be intercepted by the local resolver or proxy path.

## Security and correctness requirements

- A DHT storage owner must not be trusted as the authority for a name.
- A resolver must reject records with invalid signatures, wrong `network_id`, expired timestamps, stale sequence numbers, or names that do not bind to the advertised owner key.
- Record updates must be monotonic per owner/name via `seq` or an equivalent conflict rule.
- Tombstone/revocation semantics must be explicit.
- Multiple replicas returned from Chord storage must converge to a deterministic accepted record set.
- The design should document privacy limitations of direct DHT name lookup and leave room for blinded descriptor keys or onion-routed lookup later.

## Non-goals for the first issue

- Public DNS integration or ICANN TLD registration.
- Human-readable name allocation such as `alice.rings`.
- Auctions, registrars, transfer markets, or global uniqueness policy for human names.
- Reusing `registerService` / `lookupService` as-is for trusted name resolution.
- Requiring a full Tor-compatible hidden-service protocol in the first patch.

Human-readable aliases can be a later layer, for example:

```text
alice.rings -> <self-authenticating-label>.rings
```

That requires separate ownership, conflict, renewal, and recovery rules and should not block the self-authenticating MVP.

## Acceptance criteria

- There is a concrete `.rings` RFC/design document or issue-linked design section covering name grammar, record schema, DHT keying, verification, updates, expiry, revocation, resolver API, and onion integration.
- The implementation plan identifies the repo surfaces that need changes, including DHT storage API usage, node processor APIs, RPC/browser/native interfaces, and onion route construction.
- The MVP explicitly distinguishes self-authenticating `.rings` names from future human-readable aliases.
- Tests are planned for canonicalization, signature verification, expired/stale records, wrong-network records, conflicting replicas, and resolver cache behavior.

RyanKung:
## Summary

- add `.rings` self-authenticating service name records backed by Chord storage
- expose publish/resolve APIs through processor, JSON-RPC, native CLI, and browser provider
- reject human-readable aliases in v1 and validate owner-bound signatures, network id, expiry, and sequence selection

Closes #662.

## Validation

- `cargo check -p rings-node --no-default-features --features node`
- `cargo check -p rings-node --target wasm32-unknown-unknown --no-default-features --features browser`
- `cargo test -p rings-node --no-default-features --features node rings_name`
- `git diff --check`

### DIFF
--- a/crates/node/bin/rings.rs
+++ b/crates/node/bin/rings.rs
@@ -589,6 +589,7 @@ enum ServiceCommand {
     Lookup(ServiceLookupCommand),
     PublishRingsName(PublishRingsNameCommand),
     ResolveRingsName(ResolveRingsNameCommand),
+    BuildRingsNameRoute(BuildRingsNameRouteCommand),
 }
 
 #[derive(Args, Debug)]
@@ -641,6 +642,24 @@ struct ResolveRingsNameCommand {
 
     #[arg(long, default_value_t = false)]
     include_expired: bool,
+}
+
+#[derive(Args, Debug)]
+struct BuildRingsNameRouteCommand {
+    #[command(flatten)]
+    client_args: ClientArgs,
+
+    name: String,
+
+    #[arg(
+        long,
+        default_value_t = 0,
+        help = "Desired hop count including the .rings target; 0 uses node default"
+    )]
+    hop_count: u32,
+
+    #[arg(long, default_value_t = false)]
+    allow_short_paths: bool,
 }
 
 #[derive(Args, Debug)]
@@ -968,6 +987,15 @@ async fn run(cli: Cli) -> anyhow::Result<()> {
                 .new_client()
                 .await?
                 .resolve_rings_name(args.name.as_str(), args.include_expired)
+                .await?
+                .display();
+            Ok(())
+        }
+        Command::Service(ServiceCommand::BuildRingsNameRoute(args)) => {
+            args.client_args
+                .new_client()
+                .await?
+                .build_rings_name_route(args.name.as_str(), args.hop_count, args.allow_short_paths)
                 .await?
                 .display();
             Ok(())
--- a/crates/node/bin/rings.rs
+++ b/crates/node/bin/rings.rs
@@ -102,18 +102,8 @@ fn parse_onion_exit_service(raw: &str) -> Result<OnionExitService, String> {
     if name.is_empty() {
         return Err("onion exit service name must not be empty".to_string());
     }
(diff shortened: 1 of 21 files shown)


RyanKung:
Not plan

PR #666, opened 2026-07-20
Add gateway WebView and pending connection admission

RyanKung:
## Summary

This PR introduces a WebView gateway and the associated browser/runtime safety boundaries.

- **Gateway and browser runtime:** shared request policy, CORS and double-keyed virtual-cookie handling, first-render HTML rewriting and runtime DOM rewriting, typed gateway failures, bounded `srcdoc` recursion, and browser-service-worker integration.
- **Node and connection lifecycle:** target-aware native/Wasm runtime predicates, Provider listener lifecycle ownership, pending-handshake admission, generation-aware retirement, and serialized extension protocol transitions/effects.
- **Onion transport:** route-aware HTTPS/onion proxying, encrypted fixed-bucket hop cells, fixed four-cell link batches with authenticated cover cells, injectable send pacing, quantized forward expiry/open timing, and activity-reset idle reclamation.
- **Resource isolation:** global and per-peer crypto operation/visible-byte budgets with LRU peer windows, per-return-peer exit circuit quotas, LRU replay partitions with retained replay witnesses, count-and-byte-bounded send queues, and Service Worker admission before request-body materialization.
- **Platform and functional boundaries:** canonical `rings_native`/`rings_browser` cfgs, target-aware detached tasks, native `Send + Sync` gateway transports, and pure decision functions for cookie, header, replay, DNS, queue, and exit-accounting transitions.

## Safety, privacy, and liveness properties

- Onion observers learn only a small public cell size class; direction, exact payload length, message discriminant, circuit frame, and application bytes are hop-encrypted. Each link emits fixed four-cell batches, filling unused slots with authenticated one-hop cover cells.
- Relay layers do not expose an exact remaining-hop countdown. Malicious work is bounded by identity-independent crypto operation and visible-byte budgets plus return-table budgets; LRU eviction prevents a full peer window table from permanently excluding a new peer.
- A rotated peer identity cannot reopen a consumed forward nonce, and rejection at the replay-owner bound does not mutate the live peer state.
- One return peer cannot pin the global exit-circuit budget, one minted DID set cannot bypass the global crypto budget, and oversized public cell buckets consume their full visible-byte cost.
- Gateway request metadata uses a closed locale-free allowlist, cookies are partitioned by schemeful top-level site, bodyless requests bypass body admission, and retained request bodies are bounded before serialization without an unbounded waiter queue.
- A pending peer cannot dispatch application/DHT messages or become routable before its data channel opens and its current generation is admitted.
- For one extension protocol, committed transitions and externally visible effects have one order; terminal relay-control and onion sends execute outside the transition gate.
- Idle onion pumps and native TCP/UDP relay sessions reclaim their resources after an activity-reset deadline, including generation-safe removal of client-side UDP flow entries.

## Validation run on final head `2a284db6`

- `cargo test --release --all`
- `cargo test --release -p rings-webview` — 97 passed, 0 failed
- `cargo test -p rings-node --release --target=wasm32-unknown-unknown --features browser_default --no-default-features` with a Chrome-151-compatible ChromeDriver — 12 passed, 0 failed
- `npm run build:frontend-extension-scripts`
- `npm run test:frontend-extension-webview` — packaged MV3 extension, Onion WebView Chromium fixture, and focused regressions passed
- `npm run test:frontend-extension-wallet` — packaged extension wallet bridge passed
- `cargo clippy -p rings-node --no-default-features --features node --lib -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `cargo clippy -p rings-node --features browser --no-deps --no-default-features --target=wasm32-unknown-unknown --lib -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing`
- `RUSTDOCFLAGS='-D warnings' cargo doc --all --no-deps`
- `taplo format --check`
- `git diff --check`

GitHub Actions QACI and CodeQL passed on final head, including the packaged-extension Chromium tests and the full Wasm/browser matrix.


frontend/rings-webview-service-worker.js
```diff
@@ -0,0 +1,477 @@
...
+  }
+  try {
+    return decodeURIComponent(encoded);
+  } catch (_error) {
+    return undefined;
+  }
+}
+
+function requestKind(request) {
+  const taggedKind = request.headers.get("x-rings-webview-kind");
```
RyanKung: P1: `X-Rings-Webview-Kind` is page-writable, but this value controls whether the host runs the virtual CORS checks. The bootstrap first adds `xhr`, then page code can call `xhr.setRequestHeader("X-Rings-Webview-Kind", "subresource")`; browsers combine duplicate request headers, yielding `xhr, subresource`. That misses the exact matches below and falls back to `subresource`, so `WebviewHostRequest::subresource` skips virtual CORS even though the XHR caller can read the response. Please derive the request kind from trusted host state (or reject malformed/multiple runtime tags) rather than treating an unrecognized tag as a subresource.

RyanKung (commented):
One blocking CORS-isolation finding.

RyanKung (commented):
Rustacean re-review: the pushed head still has required WASM/clippy failures.

frontend/src/webview/onion_browser_tests.rs
```diff
@@ -0,0 +1,478 @@
...
+            .await
+            .map_err(|error| WebviewError::Transport(format!("open idb storage: {error:?}")))?,
+    );
+    let processor = ProcessorBuilder::from_config(&config)
+        .map_err(|error| WebviewError::Transport(format!("build processor config: {error:?}")))?
+        .storage(storage)
+        .dht_finger_table_size(TEST_DHT_FINGER_TABLE_SIZE)
+        .build()
+        .map_err(|error| WebviewError::Transport(format!("build processor: {error:?}")))?;
+    let provider = Arc::new(Provider::from_processor(Arc::new(processor)));
```
RyanKung: P1: this makes the required WASM/clippy job fail. `Provider` and `Processor` are explicitly neither `Send` nor `Sync`, so wrapping them in `Arc` triggers `clippy::arc_with_non_send_sync` (the CI log points here). This browser-only, single-threaded ownership needs `Rc` throughout its local test boundary, or a justified Send+Sync boundary. The same job also reports the `value.as_ref()` call below as useless. Please make this test compile cleanly under the repository's `-D warnings` policy.

frontend/src/webview.rs
```diff
@@ -0,0 +1,840 @@
...
+
+  const script = document.createElement("script");
+  script.src = "/assets/webview-overlay.js";
+  script.async = false;
+  script.dataset.ringsWebviewOverlayLoader = "";
+  (document.head || document.documentElement).append(script);
+})();
+"#;
+
+thread_local! {
```
RyanKung: P1: the required clippy job also rejects this new `thread_local!` initializer with `clippy::missing_const_for_thread_local` under `-D warnings`. Use a `const { ... }` initializer (or otherwise make the intentional runtime initialization explicit) so the PR's required lint job is green.

crates/core/src/swarm/transport/pending.rs
```diff
@@ -0,0 +1,292 @@
...
+
+#[derive(Debug)]
+pub(super) struct PendingPeer {
+    generation: u64,
+    admitted_at_ms: i64,
+}
+
+/// Bounded, non-routable handshakes owned by the swarm lifecycle.
+///
+/// The pool deliberately has no DHT reference: a peer is visible to Chord
```
RyanKung: P2: the pending-admission invariant is documented, but it has no executable state-transition model. `Pending -> Active` is split across this pool, `InnerSwarmCallback`, and DHT join/leave effects; the new Stateright-style model covers storage CRDT convergence, not this lifecycle. Add a finite model (or a property/state-machine test) over `Absent/Pending(generation)/Active` with `open`, `close`, `failed`, `timeout`, replacement, and late-callback actions. It should witness: a pending peer is never routable; only the matching generation may promote; and every terminal/expiry transition removes the DHT membership and transport slot. The current example tests do not cover those event orderings.

RyanKung (commented):
The lifecycle model still needs a formal witness.

frontend/scripts/test-webview-service-worker.mts
```diff
@@ -0,0 +1,114 @@
...
+  Response,
+  performance,
+  setTimeout,
+  clearTimeout,
+  self: {
+    location: new URL("http://127.0.0.1:8080/"),
+    addEventListener() {},
+  },
+};
+context.globalThis = context;
```
RyanKung: P1: this currently breaks the required WASM job before the frontend checks run. `npm run build:frontend-extension-scripts` reports TS4111 here because `ServiceWorkerTestContext` has an index signature: access this as `context["globalThis"] = context` (or model `globalThis` as an explicit property). The local Node execution passes, but the generated-script TypeScript check in CI does not.

crates/core/src/swarm/transport/pending.rs
```diff
@@ -0,0 +1,490 @@
...
+                capacity: MAX_PENDING,
+            });
+        }
+
+        self.next_generation = self.next_generation.wrapping_add(1);
+        let attempt = PendingConnectionAttempt {
+            peer,
+            generation: self.next_generation,
+        };
+        self.peers.insert(
```
RyanKung: P1: `cargo +nightly fmt --all -- --check` still reports this block (and the analogous assertion formatting in `transport/tests.rs`) as unformatted. QACI runs that exact nightly command, so the required rustfmt job will fail. Please format with the repository's nightly toolchain before pushing.

crates/core/src/swarm/transport/pending.rs
```diff
@@ -0,0 +1,490 @@
...
+        if self.peers.contains_key(&peer) {
+            return Err(Error::AlreadyConnected);
+        }
+        if self.peers.len() >= MAX_PENDING {
+            return Err(Error::PendingConnectionCapacityExceeded {
+                capacity: MAX_PENDING,
+            });
+        }
+
+        self.next_generation = self.next_generation.wrapping_add(1);
```
RyanKung: P2: the new model is not a refinement of the implementation at the generation-exhaustion boundary. The production counter wraps here, whereas `LifecycleModel::replace_pending` uses `saturating_add`; consequently the model never represents the ABA case where an old callback generation becomes equal to a new live generation after wraparound. Either make production generation allocation fallible/non-reusing and model that error, or give the model the same wrap semantics and explicitly state the accepted bound. The claimed generation-safety proof is otherwise incomplete.

RyanKung (commented):
Fresh review of the latest feedback commit: required CI still has reproducible blockers.

crates/node/src/registration.rs
```diff
@@ -343,3 +362,88 @@ impl RegistrationTask for OnlineNodeRegistration {
...
+
+    fn encoded(value: &str) -> Encoded {
+        value.into()
+    }
+
+    fn encoded_subset(mask: u8) -> BTreeSet<Encoded> {
+        ["a", "b", "c"]
+            .into_iter()
+            .enumerate()
+            .filter_map(|(bit, value)| (mask & (1 << bit) != 0).then(|| encoded(value)))
```
RyanKung: P1: This new property test makes the wasm CI job fail. The workflow runs clippy with `-D warnings`, and this `filter_map`+`bool::then` is rejected by `clippy::filter_map_bool_then`, so `Build and test for wasm` cannot complete. Please rewrite it as `filter(...).map(...)` (or otherwise avoid the lint) and rerun the wasm job.

RyanKung (commented):
Node/wasm follow-up: the wasm CI job is currently blocked by the new registration test.

crates/transport/src/connections/native_webrtc/mod.rs
```diff
@@ -66,6 +68,101 @@ const DELIVERY_POLL_INTERVAL: Duration = Duration::from_millis(300);
...
+        .collect()
+}
+
+fn duplicate_host_candidate_line(line: &str, extra_addresses: &[String]) -> Vec<String> {
+    if !line.starts_with("a=candidate:") {
+        return Vec::new();
+    }
+
+    let fields = line.split_whitespace().collect::<Vec<_>>();
+    if fields.len() < 8 || fields[6] != "typ" || fields[7] != "host" {
```
RyanKung: P1: The native node CI cannot pass with the repository's panic-free clippy policy. Even though the preceding `len() < 8` guard makes these indices logically safe, `cargo clippy -p rings-node --features ffi` treats `clippy::indexing_slicing` as an error here (and at the related accesses below), so both native examples and the lint workflow fail. Destructure the required fields with checked access/an iterator before using them.

RyanKung (commented):
Native follow-up: the native examples/lint workflows are also blocked by the new candidate rewrite.

RyanKung (commented):
P1 — The runtime abstraction is not consistent across native and wasm.

The core changes now model browser semantics with `all(feature = "wasm", target_family = "wasm")`: a native build that enables the wasm feature remains Send-capable and uses native time/storage implementations. `rings-node` still defines the corresponding runtime distinction with `feature = "browser"` alone:

- `extension/ext/mod.rs:58-66` makes `MaybeSend` empty whenever the browser feature is enabled;
- `registration.rs:80-95` and `processor/mod.rs:98-111` select `window_sleep` from that feature alone;
- `RegistrationTask` and its `async_trait` expansion follow the same feature-only predicate.

Thus the same native target has two incompatible models when both capability features are selected: core requires native Send semantics, while node erases them and attempts JS-only adapters. This is not merely an unsupported feature combination: the PR specifically made the core predicate target-aware, but did not refine the node layer to it. `cargo check -p rings-node --all-features` demonstrates the broken composition (missing `js_utils`/IndexedDB plus Send and wasm-bindgen errors).

Please define one target-aware runtime predicate (or use `cfg(all(feature = "browser", target_family = "wasm"))` consistently) for node's MaybeSend, async-trait, timers, browser modules, and exports. Add a native “browser feature enabled” compile check so this refinement cannot drift again.

RyanKung (commented):
P2 — The lifecycle capability is implemented below the public browser boundary, so native and wasm expose different abstractions.

Native `Provider` exposes `listen_with(StopToken)` in `crates/node/src/provider/mod.rs:293-299`. The wasm-exported `Provider::listen` in `provider/browser/provider.rs:619-625` still always calls `Processor::listen()`, which constructs `StopToken::never()`; JavaScript callers therefore cannot request the lifecycle that the new model introduces. The frontend works around this by retaining a separate `Processor` and spawning `Processor::listen_with` directly in `frontend/src/node.rs:140-156`.

The Provider facade is no longer the runtime-neutral boundary: native callers control the loop through Provider, while browser callers must bypass Provider and know its implementation. Expose an equivalent browser lifecycle handle/stop operation at the Provider boundary (or explicitly make the lifecycle owner a separate cross-runtime type) and add a parity test for start → stop → settled listener in both adapters.

RyanKung (commented):
P2 — The public browser boundary is now present, but the webview frontend still bypasses it.

`Provider::listen()` now returns `ProviderListener`, yet `frontend/src/node.rs:140-163` retains an `Arc<Processor>`, creates a separate `StopSource`, and starts `Processor::listen_with` directly. Consequently the same browser node has two lifecycle abstractions: external callers use `ProviderListener`, while the in-repo webview owner uses an unrelated source/token pair below the Provider boundary.

The boundary is not yet the single model. Store `ProviderListener` in `DemoNode` and implement `DemoNode::stop` through it (or make a single runtime-neutral listener type used by both adapters). Add a frontend-level test that proves the owned listener task settles after `DemoNode::stop()`; the current wasm test only observes `StopSource::is_stop_requested`, not listener completion.

PR #681, opened 2026-08-15
Handle FFI ownership, CI gates, online-node GC, and threat model

RyanKung:
Closes #667.
Closes #670.
Closes #651.
Closes #671.

Supersedes #677, #678, #679, and #680.

Summary:
- Sync package metadata to `0.17.0` across the Rust workspace and npm/browser package so published artifacts report the same version.
- Remove obsolete protoc setup from QACI/release workflows; RPC DTOs are plain serde structs now and node/core builds no longer require protobuf generation.
- Remove PR/release-path `Swatinem/rust-cache` steps so CI no longer fails before execution when GitHub codeload rate-limits external action downloads.
- Pin GitHub Actions to immutable commit SHAs. The inline `# v3`/`# v4` comments record which upstream tag the SHA came from; the workflow now executes the reviewed commit, not a mutable tag ref.
- Add CI security gates: cargo-deny dependency policy, focused Miri coverage, and Linux FFI ASan/LSan coverage.
- Miri is Rust's interpreter for catching undefined behavior in unsafe-sensitive code paths. This PR runs it with strict provenance on core modules that do not require native sockets, wall-clock interpretation, or expensive crypto searches.
- Scope sanitizer rustflags to the Linux test target so host/proc-macro dependencies are not compiled with ASan flags.
- Replace FFI raw `Arc` reconstruction with an opaque destructible provider handle, add explicit provider/string free APIs, and update the Python FFI wrapper/tests.
- Break the FFI provider callback reference cycle during destroy and add a `Weak` regression assertion so provider ownership is actually released.
- Resolve the dependency-policy gate without vendored serializer code: remove advisory ignores, exclude SNARK extension/example crates at cargo-deny graph construction, and replace the former bincode/wincode wire wrapper with `rings-codec` backed by `postcard`.
- The codec change is intentional because this project does not require pre-release wire compatibility. Keeping bincode/wincode would either preserve a protocol we are willing to change or require carrying local serializer patches for gaps such as 128-bit integer handling. The maintained path is now `rings-codec`; old bincode crates remain only where unrelated transitive dependencies still pull them in.
- Delete `third_party/serde-wincode` and root/standalone path patches; refresh root, frontend, dweb, and proof-demo locks so Node/core/browser paths use `rings-codec` without vendored serializer code.
- Add codec regression coverage for stable postcard bytes, i128/u128 roundtrips, serialized-size accounting, and trailing-byte rejection.
[...]

PR #695, opened 2026-08-28
Extract measurement into rings-measure and add aMule-style local credits

### LINKED ISSUE #689: Extract measurement into rings-measure and implement aMule-style local credits
## Summary

Extract the current peer-measurement subsystem from `rings-core` and `rings-node` into a standalone `crates/measure` workspace crate, published as `rings-measure`.

The new crate must implement two explicitly separate local models:

1. an aMule/eDonkey-style byte-credit ledger for reciprocal resource priority; and
2. a recent transport-reliability window for connection-attempt ordering.

These models are local observations. They are not global reputation, and neither model may become a hidden Chord membership, ownership, or storage-placement rule.

Related: #639 introduced advisory peer-quality ordering. This issue extends and restructures that implementation; it is not a replacement request for the already completed #639 scope.

## Why this is needed

The current code is split across:

- `crates/core/src/measure/**`: traits, counters, quality types, and ordering;
- `crates/node/src/measure.rs`: periodic counter state and persistence;
- `crates/core/src/swarm/transport/outbound/measurement.rs` and transport callbacks: event recording;
- DHT, onion routing, provider, and RPC call sites: policy and query consumers;
- `rings-derive::MeasureBehaviour`: a measurement-specific forwarding macro.

It measures connection and message outcome counts, but it does not implement the aMule/eDonkey local credit relation based on useful bytes exchanged with an authenticated peer.

The extraction must also fix, rather than preserve, these problems:

- the persisted value is only a bare counter, so the active window, epoch boundary, and age are not restored faithfully;
- cache misses and hits both consult storage on the counter hot path;
- storage errors are hidden behind `Measure` methods that cannot return an error;
- the current outbound recorder may drop queued observations on capacity pressure or shutdown;
- chunked receive paths can account at chunk and reassembled-message boundaries, so one logical transfer can be counted more than once;
- bulk queries enumerate connected peers rather than the complete measured ledger;
- stale records have no complete `last_seen` lifecycle and the in-memory set is unbounded;
- the documented advisory quality relation is also consumed by `LocalFailureLimit` topology removal, which turns local evidence into a Chord eviction rule.

## Crate boundary

Add this workspace dependency:

```toml
rings-measure = { path = "crates/measure", version = "0.18.0" }
```

`rings-measure` must not depend on `rings-core`. The domain is generic over a stable `PeerId`; `rings-core` and `rings-node` instantiate it with `Did`. This avoids a `rings-core -> rings-measure -> rings-core` cycle.

Suggested modules:

- `credit`: byte ledger, policy, and opaque score type;
- `reliability`: aligned recent windows, evidence, policy, and `Healthy | Unknown | Degraded`;
- `event`: authenticated, logical measurement events;
- `ledger`: pure update/query/prune transitions keyed by `PeerId`;
- `snapshot`: versioned serializable state and restore validation;
- `error`: typed domain and snapshot errors.

The crate owns the model and state transitions. Runtime effects remain adapters:

- `rings-core` emits authenticated transport events and consumes typed ordering inputs;
- `rings-node` owns synchronization, native/IndexedDB storage, batching, lifecycle, and RPC mapping.

Remove `MeasureBehaviour` from `rings-derive` if it has no remaining non-measurement consumer.

## Model A: local byte credits

For each authenticated peer, persist:

- `bytes_sent_to_peer`: useful payload bytes delivered by this node;
- `bytes_received_from_peer`: useful payload bytes delivered by the peer;
- `last_seen`.

Use the aMule score relation, with direction named from the local node perspective:

```text
if bytes_received_from_peer < 1_000_000:
    score = 1.0
else:
    ratio = 10.0 if bytes_sent_to_peer == 0
            else 2.0 * bytes_received_from_peer / bytes_sent_to_peer
    volume_cap = sqrt(bytes_received_from_peer / 1_048_576 + 2.0)
    score = clamp(min(ratio, volume_cap), 1.0, 10.0)
```

Requirements:

- expose an opaque finite `CreditScore` constrained to `[1.0, 10.0]`;
- accrue credit only after the remote `Did` has been authenticated;
- count useful logical payload bytes, not framing, retransmission, or duplicate chunk overhead;
- use checked arithmetic and typed overflow errors;
- preserve totals across restart;
- expire records after 150 days without observation by default, with the retention policy explicit and configurable;
- expose the score as a typed resource-priority input, but do not feed it into DHT topology decisions.

The defaults deliberately match the aMule threshold, ratio/cap formula, score range, authenticated-identity gate, and record-retention behavior.

## Model B: recent transport reliability

Keep transport reliability separate from byte credit.

A reliability record must retain aligned, timestamped buckets for logical outcomes:

- connection succeeded / failed or disconnected;
- logical message sent / failed;
- logical message received and verified / failed verification.

Requirements:

- a configurable `ReliabilityPolicy` owns window duration, minimum evidence, and classification thresholds;
- `Unknown` represents insufficient evidence, not failure;
- all success and failure observations for a logical operation land in one epoch;
- rollover is based on explicit timestamps and handles clock rollback without underflow or extending stale data indefinitely;
- snapshot/restore preserves the same live-window judgement;
- the DHT consumer performs only a stable `Healthy -> Unknown -> Degraded` permutation;
- no candidate is inserted, removed, or made ineligible by this score;
- remove score-driven `LocalFailureLimit` topology eviction. Topology removal must remain based on the explicit DHT/transport transition that owns membership and liveness.

## Runtime and persistence design

Use a memory-first state transition:

1. apply an event to the in-memory ledger synchronously at the named transport boundary;
2. mark the peer or snapshot dirty;
3. coalesce and flush dirty state asynchronously;
4. perform a final bounded flush during graceful shutdown.

This keeps network I/O independent from storage latency without dropping domain events through a bounded recorder queue.

Persistence requirements:

- versioned snapshot schema;
- credit totals, complete reliability buckets and epoch timestamps, and `last_seen` stored together;
- atomic replace or equivalent backend-safe commit;
- explicit load/save/corruption/version errors surfaced to the caller and metrics/logging;
- no storage read on every cached update or query;
- no synchronous filesystem I/O on an async network hot path;
- bounded memory through expiry/pruning;
- native and IndexedDB adapters with equivalent restore semantics.

## Integration and migration plan

1. **Scaffold and dependency direction**
   - add `crates/measure`, workspace metadata, features, lints, and public documentation;
   - make domain APIs generic over `PeerId`;
   - add direct `rings-measure` dependencies to the consumers that need it.

2. **Implement the pure model**
   - implement byte credits, score calculation, reliability transitions, classification, pruning, snapshots, and typed errors;
   - keep time explicit in transition inputs so tests do not depend on wall-clock sleeps.

3. **Replace the node implementation**
   - replace `PeriodicMeasure` internals with the new ledger/runtime adapter;
   - migrate existing stored counters when possible; otherwise use a documented schema-version fallback that never interprets legacy values as byte credits;
   - remove the measurement-specific derive macro and legacy traits after call sites migrate.

4. **Correct event boundaries**
   - emit one success or one failure per logical send/receive;
   - record useful bytes only after delivery/verification;
   - ensure chunking and reassembly do not double count;
   - bind records to the authenticated stable `Did`;
   - make persistence dirty notifications coalescing, observable, and lossless for already-applied in-memory events.

5. **Migrate consumers**
   - keep DHT candidate sorting on reliability only and preserve the candidate set;
   - remove measurement-driven topology eviction;
   - let onion routing query every route candidate, including disconnected peers with retained history;
   - make provider/RPC queries read any measured peer and list the complete retained ledger;
   - expose byte totals, `CreditScore`, reliability evidence, class, timestamps, and schema-safe DTOs without implying global trust.

6. **Delete legacy code**
   - remove `crates/core/src/measure/**`, superseded node counter logic, dead adapters, and obsolete re-exports only after all native and browser callers migrate.

### Pure model

- score is always finite and within `[1.0, 10.0]`;
- below 1,000,000 received bytes the score is exactly `1.0`;
- zero sent bytes, ratio lower bound, square-root cap, and maximum cap match the formula;
- score is monotone with additional received bytes for a fixed sent total and cannot improve when only sent bytes increase;
- unauthenticated events do not accrue credits;
- overflow returns a typed error without partially mutating state;
- reliability classification covers no evidence, insufficient evidence, healthy evidence, and threshold crossings;
- epoch rollover and clock rollback preserve the documented window relation;
- snapshot round-trip, restart restoration, unsupported version, corruption, and 150-day expiry are covered;
- pruning is idempotent.

### Integration

- a single logical payload produces one outcome event and the exact useful-byte delta;
- the same payload sent through chunking produces the same ledger delta as an unchunked payload;
- invalid, duplicate, and conflicting chunks never count both success and failure;
- storage latency or failure does not block transport and is observable;
- graceful shutdown restores all applied ledger updates after restart;
- listing includes retained disconnected peers;
- DHT quality ordering is a stable permutation and no quality class directly evicts a peer;
- onion routing uses retained quality for all candidates while preserving fallback;
- native and `wasm32-unknown-unknown` builds expose equivalent public semantics.

Use model-based or property tests for transition sequences, in addition to example tests. Test assertions should state the invariant being witnessed.

## Validation

At minimum:

- `cargo check -p rings-measure --all-features`
- `cargo test -p rings-measure --all-features`
- `cargo test -p rings-node --lib`
- `cargo check -p rings-node --no-default-features --features browser --target wasm32-unknown-unknown --lib`

Also run the focused transport, DHT stabilization, onion route, provider/RPC, persistence-restart, and chunked-delivery suites added or changed by this issue.

## Non-goals

- global, signed, or replicated reputation;
- Sybil-resistance claims;
- using credit or reliability to decide Chord ownership, membership, successor responsibility, or storage placement;
- gossiping local scores over the wire;
- silently changing transport fairness or relay billing. A future resource scheduler may consume `CreditScore` through an explicit policy.

## References

- aMule credit counters, authenticated-identity gate, and score formula: https://github.com/amule-project/amule/blob/master/src/ClientCredits.cpp
- aMule persistence, 150-day expiry, and periodic save behavior: https://github.com/amule-project/amule/blob/master/src/ClientCreditsList.cpp

RyanKung:

Feature and measurement-correctness refactor.

Peer measurement is split between core and node counter implementations. It stores bare recent counters, does not implement aMule/eDonkey byte credits, may read storage on hot paths, and lets local failure evidence influence DHT topology removal. Logical chunked transfers also lack an explicit exact useful-byte accounting model.

See #689.

- Adds `rings-measure` as a pure, generic algorithm crate with explicit time, typed errors, deterministic snapshots, and no runtime or storage dependencies.
- Implements authenticated aMule-style useful-byte credits, finite scores, exact 150-day retention, and safe migration of obsolete short-term reliability windows without losing long-term credit.
- Separates byte credit from recent reliability evidence and uses peer quality only for stable advisory ordering.
- Models attribution explicitly: authenticated ingress can establish peer state, while a locally selected DID can update only `FailedToSend` reliability for an already-retained peer. Local-only failures cannot create records, refresh authenticated `last_seen`, or wash out credit through identity rotation.
- Keeps retained peer state bounded at 16,384 records with deterministic stalest-record eviction, observable eviction reports/warnings, and bounded paginated RPC access.
- Records one bounded terminal outcome and exact useful-byte deltas per verified logical transfer, including chunked delivery, expiry, invalid input, replay, local-capacity rejection, and terminal-history saturation/drain.
- Replaces node measurement state with a memory-first ledger adapter, coalesced cancellation-safe persistence, retry/final-flush handling, and explicit FFI flush before runtime shutdown even when `listen` was never started.
- Removes the measurement-specific derive macro, the dead boolean judgement API, and measurement-driven DHT eviction.

Yes, for direct consumers of measurement and RPC APIs:

- `PeriodicMeasure` construction is async, returns a typed result, and requires a live Tokio runtime on native targets.
- `Measure` adds attribution-aware `record` / `record_batch`; the provided batch fallback is counter-only and non-atomic, while ledger-backed implementations override it atomically.
- `MeasurementLedger::apply` / `apply_batch` return `ApplyReport`, exposing both the semantic outcome and any deterministically evicted peer.
- `BehaviourJudgement::good` and the obsolete `MeasureBehaviour` derive are removed.
- `ListPeerMeasurements` now uses a bounded `limit` plus exclusive cursor and returns a next cursor.
- Legacy bare counter records are intentionally not interpreted as byte credits; the versioned snapshot starts independently because old values cannot prove byte direction, timestamp, or identity attribution.

Validation completed locally on `26afa0ed`:

- `cargo test -p rings-measure --all-features`: 37 passed
- `cargo test -p rings-core --lib --features dummy`: 616 passed, 1 ignored performance probe
- `cargo test -p rings-node --lib --features ffi`: 308 passed
- workspace doc tests: 1 passed, 2 ignored examples
- workspace Clippy with warnings denied
- strict production Clippy for measure/core/native node plus core/node WASM configurations
- locked frontend WASM check and Clippy
- `cargo +nightly fmt --all -- --check` and `git diff --check`

The final patch was self-reviewed after all fixes. That review found and closed a two-stage terminal-history saturation/drain counterexample; no actionable P0-P3 finding remains locally.

Closes #689


RyanKung:
Full read of the diff (4,876 lines) plus cross-checking the touched files on the branch. The overall direction is sound: the pure algorithm crate boundary, transactional transitions (a failed transition preserves the prior record), snapshot validation, removal of measurement-driven DHT eviction, an aMule formula that matches the eMule original (including the 1e6 activation threshold vs. MiB volume-cap quirk), and the constant-space lossless outbound aggregator are all good. But there are several correctness/robustness problems and a batch of abstraction issues.

## Correctness / robustness

**1. A clock regression makes the node fail to start — `crates/node/src/measure.rs:154`**

`new_with_clock` propagates `ledger.prune(now, …)?`. `is_expired` computes `now - last_seen` via `checked_sub`, so if any record in the snapshot has `last_seen` ahead of the current wall clock (NTP step-back, VM resume, clock correction before a restart), the constructor returns `Err` and `daemon_run` / `Provider::new` abort. With 150-day retention the exposure window is long: one snapshot written under a fast clock means the node refuses to start until the wall clock catches up. The startup path should fail open (skip or clamp regressed records) and treat `ClockRegression` as an observable degradation, not a fatal error.

**2. The reassembly path loses `FailedToReceive`, and introduces dead code plus a double deserialization — `crates/core/src/swarm/callback.rs:623`**

To compute `is_chunk`, `decode_verified_message` now performs a full `data::<Message>()?` on the reassembled payload, and the error propagates via `?` without calling `record_receive_failure`. On master this failure was scored by the match arm in `crates/core/src/swarm/callback/inbound/reassembly.rs:62-68` — that arm is now unreachable (the second decode of the same bytes cannot fail once the first succeeded). So a peer that delivers a valid, signature-verified envelope with an undecodable inner `Message` is no longer scored `FailedToReceive`, contradicting the PR's own model of "one outcome per verified logical transfer". At the same time, every reassembled message (which is by definition large) is deserialized twice back to back on the hot path — right next to a test literally named `test_prepared_chunk_is_moved_without_redecoding_transaction_data`. The root cause is that the chunk-ness decision sits at the wrong boundary: the single decode point (`advance_chunk_event`) should decide whether to record the measurement; `decode_verified_message` should not decode at all.

**3. Persistence write storm — `crates/node/src/measure.rs:276`**

Every `record()` sets `dirty` and wakes the worker; `persist_pending` loops writing the **complete** snapshot until `!dirty`. Under sustained traffic the worker becomes a 100%-duty-cycle serialize-the-entire-150-day-ledger-and-write loop, O(retained peers) per write. The old implementation wrote one `u64` per counter per hour. This needs a minimum write interval / debounce.

**4. One bad record poisons every projection — `crates/measure/src/ledger.rs:216`**

`measurements()` is a collect-of-`Result`s: the first `ClockRegression` aborts the whole projection, and the transport layer degrades to an empty list. `prune` is likewise all-or-nothing. A single peer whose `last_seen` is in the future empties the entire RPC list and does not self-heal. Prefer per-peer degradation (skip and log) over collective failure.

## Model / abstraction

**5. `u64::MAX` sentinel thresholds — `crates/core/src/measure/quality.rs:52`**

The counter-only projection classifies with MAX thresholds, so it can never report `Degraded` — except, ironically, when a counter equals exactly `u64::MAX`, because the comparison is `>=`. Meanwhile the same implementation's `BehaviourJudgement::quality` may say `Degraded`, so the RPC `reliability` field and `quality()` disagree. The sentinel is encoding "no policy"; model that explicitly instead.

**6. `record_as_logical_message: bool` is naked boolean plumbing.** Both call sites re-derive the proposition "a chunk is not a logical message" independently (`inbound.rs:426` via `matches!`, `callback.rs` via an extra decode), even though `MessageKind::is_chunk` already exists. Name the predicate and decide it at a single point — same root cause as #2.

**7. The event↔counter mapping is written four times**: the `record` default in `behaviour.rs`, `transport/tests/mod.rs`, `test_chunk_e2e.rs` (three verbatim forward copies), and the inverse in `node/src/measure.rs::incr`. Three occurrences is the abstraction threshold — give `MeasurementEvent` a named conversion in core.

**8. `crates/core/src/swarm/transport/outbound/measurement.rs:72,91`** — the `Sent` arm bodies of `take_next`/`take_other` are five verbatim-duplicated lines, and `OutboundMeasurement` doubles as both "event" and "kind": `other()` fabricates a `Sent { useful_bytes: 0 }` dummy. Split out a payload-free kind enum and write take as a `find_map` over `[first, first.other()]` (the old code already had that shape). Also note the coalescing semantics attach the entire byte total to the last `Sent` event of a batch (the rest carry 0) — count and sum are preserved, but per-event bytes are distorted; worth documenting at the event definition.

**9. `Authentication` has exactly one live value at runtime**: the adapter passes `Authenticated` unconditionally and `Unauthenticated` appears only in tests. If the argument is "the swarm only records verified events", then the proof lives in the caller and this type axis is currently speculative. Either derive it from a real fact in the adapter, or state why the axis is kept.

## Layering / exposure

**10. Mixed import provenance**: `rings_core::measure` re-exports `CreditPolicy`/`CreditRecord`/`CreditScore`, yet `node/src/measure.rs` imports half its types through `rings_core::measure` (`ApplyOutcome`, `MeasureError`, `UnixTime`) and the other half directly from `rings_measure` (`CreditPolicy`, `Authentication`, `MeasurementLedger`). One type, two paths — and core re-exports credit types its own API never mentions (the trait signatures only need Event/Error/Outcome/UnixTime). Pick one boundary.

**11. `ListPeerMeasurements` now returns the complete 150-day ledger** with no cap and no pagination — the RPC response is unbounded, and it enumerates every peer DID the node has seen. At minimum this needs a limit.

## Minor

- `crates/node/src/processor/mod.rs:72`: `MEASUREMENT_FLUSH_TIMEOUT` sits in the middle of the `use` block.
- `crates/node/src/processor/tests/test_network.rs:223`: the `sleep(100ms)` is a probabilistic band-aid; the assertions are `>`-relative to baselines, so the sleep is neither necessary nor sufficient — remove it or replace it with a deterministic wait.
- `crates/node/src/measure.rs:40`: `RELIABILITY_WINDOW`'s `None => NonZeroU64::MIN` silently maps 0 to 1; in const context a misconfiguration can simply fail to compile.
- Semantics change to confirm: evidence is zeroed at every aligned-epoch rollover (documented by the test), so a `Degraded` peer resets to `Unknown` at the top of each hour; the old implementation's previous-window fallback was smoother. Fine if intentional, but worth confirming.
- `ledger.measurements()` performs a second `get` per key; iterate the entries directly.

Docs sweep is clean: no stale `MeasureBehaviour`/`ConnectBehaviour`/eviction references remain. The `ProcessorBuilder::measure` setter already took a concrete `PeriodicMeasure` on master, so the field-type change is not a new API narrowing.

The heaviest items are #1 (startup failure) and #2 (lost failure scoring + double decode); both are correctness problems that need fixing.

RyanKung:
## Review — a79c3456 ("close scoring and persistence invariants")

Fresh full read of the new commit (rebased onto #696 master). All 13 checks are green on this head. The quality of this round is high: `MeasurementBatch` removes the synthetic zero-byte events cleanly, `Authentication` is now genuinely derived at the ingress boundary (`peer_authentication()` from the pending attempt + admission state) instead of hardcoded, the reassembly terminal-outcome bookkeeping (one logical id → at most one terminal outcome; local capacity loss never attributed to the peer; early close cannot erase an incomplete-message outcome) is carefully reasoned and witnessed by the law-named adversarial suite, and the window-mismatch, retained-peer bound, and exact retention boundary all close real invariants. The findings below remain.

### 1. Missing-connection `FailedToSend` evidence is silently neutralized — this needs an explicit decision

`payload_send.rs` now passes `Authentication::Unauthenticated` at both missing-connection sites (the detached path and `connection_for_send`). Every shipping implementation — `PeriodicMeasure::record_batch`, the trait default `record`, and every test impl — returns `IgnoredUnauthenticated` for that value. Consequently:

- The entire `records_missing_connection_failure` mechanism (`MessageKind::records_missing_connection_failure()` in `message/types.rs`, the flag computation, both guarded call sites) is now a **no-op against every real `Measure`**. The previous commit recorded these failures into the ledger; this commit makes them vanish, while the flag, its doc, and the call sites still claim recording happens.
- Neither direction is locked by a test: `FailedSendMeasure`'s only use asserts the unrelated proposition `count == 0` for a local delivery timeout.

Please answer these directly rather than patching around them:

1. **Was this an intentional semantic decision** ("no admitted connection ⇒ no peer evidence") or an accidental consequence of threading `Authentication` through the send path? The commit message does not mention it.
2. If intentional — what is the justification for discarding this evidence class? A missing-connection send failure is a *locally originated* observation about a DID *we chose to address*: there is no spoofing vector (nobody else can induce it), and it is exactly the Model-B "logical message failed" evidence that issue #689 keeps for connection-attempt ordering. Classifying it as "unauthenticated" conflates *identity proof of received data* (what the aMule credit gate is for) with *eligibility of locally generated failure evidence* (what reliability ordering consumes). If the position is that failure evidence also requires a live authenticated connection, that position should be stated in the `Authentication` docs and the flag machinery deleted — dead call sites guarded by a flag whose name promises recording are worse than no call sites.
3. If unintentional — the fix is not merely flipping the constant. Decide what authentication means for locally originated failure evidence, document it on `MeasurementEvent::FailedToSend` / `Authentication`, and add a test that pins the chosen behavior (a missing-connection send either does or does not reach the ledger — today nothing witnesses either).

Either resolution is acceptable; an inert mechanism with recording-shaped names is not.

### 2. `Expired` reclassified Invalid → Replay: the evidence consequence is undocumented

For a retained pending entry, expiry is now charged exactly once via the expiry count — self-consistent. But a chunk that arrives *already expired* for an id with **no retained state** now produces no failure evidence at all (previously it was `Invalid` → one `FailedToReceive`). Fail-open for clock skew is defensible; it also means purely-expired spam is now reputation-free for an adversary. `Rejected::rejection`'s regrouping carries no comment about this evidence-facing trade-off. Please confirm intent and document it where the grouping is defined.

### 3. `record_expired_reassembly_failures` hardcodes `Authentication::Authenticated`

This is correct today only because of two invariants living in other files: the expiry count includes only `peer_attributable` entries (authenticated at ingress, `chunk/reassembly.rs`), and the reassembler is per-connection so all entries belong to the `pending_attempt` peer. Neither invariant is stated at the call site. Derive the value from the fact recorded at ingress, or write the precondition (`Pre: count covers only authenticated-ingress entries; this reassembler serves one connection`) next to the constant so the coupling cannot drift silently.

### 4. Smaller new items

- `PeriodicMeasure::count` still maps projection errors silently to 0 (`let Ok(Some(..)) else return 0`); the maintain-path error now logs, this path does not.
- `from_snapshot`'s `NonZeroUsize::new(DEFAULT_MAX_RETAINED_PEERS).unwrap_or(NonZeroUsize::MIN)` is a dead fallback masking a const invariant — define `DEFAULT_MAX_RETAINED_PEERS` as `NonZeroUsize` and the fallback disappears.

### Leftover from the previous round, untouched in this commit

- **`BehaviourJudgement::good` is still dead code with a self-contradicting doc**: the trait doc says mapping the three-valued quality to a boolean would create a hidden gating rule, and the node impl is exactly `!matches!(quality, Degraded)`. Zero production callers. Issue #689 says legacy traits go once call sites migrate — they have.
- **`Processor::peer_qualities` (and the browser provider) still re-classify with the legacy `ReliabilityEvidence::classify`** instead of using the `quality` the projection already carries. Two classification paths that agree only while `minimum_positive_observations == 1`; they will drift silently the day the policy changes.
- `rpc_impl.rs`: the `NonZeroUsize::new(...).ok_or_else(...)` after the `limit == 0 || limit > MAX` guard is an unreachable error path — parse once into a NonZero.
- FFI destroy-without-listen: the worker's final flush (now with retries — good) still races `Runtime::shutdown_background()` in `ProviderHandle::shutdown` when a consumer destroys a provider that never called `rings_node_listen`.
- `PeriodicMeasure::new` spawns via `tokio::spawn` on native; the ambient-runtime precondition is still not in the constructor doc, even though the FFI layer had to be restructured around exactly this requirement.
- `ReliabilityPolicy` proves `NonZeroU64` and then stores a bare `u64`; `ensure_policy` now mitigates the consequence at runtime, but the type-level proof for the `%` divisor is still discarded.

### Confirmed fixed from the previous round

File-size split (703 lines + three test modules), zero-byte synthetic events (replaced by atomic batches), hardcoded authentication (now derived), shutdown flush ordering (sibling stop + bounded final retries, with a test), plus the self-driven invariant work called out at the top.

Item 1 is the one that blocks: it is either an undeclared behavior regression or a mechanism that should be deleted — it cannot stay as shipped.

RyanKung:
Re-reviewed the two new commits (`1fe3d10d`, `a79c3456`) with a fresh full read (~10k diff lines).

## Status of the previous round

All 16 items from the previous review are substantively addressed, and mostly by restructuring at the right boundary rather than patching the flagged lines:

- Startup clock regression → an explicit `reconcile_clock` recovery transition at startup and on live-query regression detection (via the `last_clock` sentinel, with clock sampling serialized under the runtime lock).
- Lost `FailedToReceive` / double decode → the chunk-ness decision now lives at the single decode point; `decode_verified_payload` no longer decodes the inner `Message`, and `accept_verified_logical_message` runs after the one decode in `advance_chunk_event`. On top of that, the reassembler grew a full terminal-outcome model (one peer-attributable failure per logical transmission, local capacity loss never charged to the peer, close-time retention until TTL) with serious adversarial test coverage.
- Persistence write storm → 60s debounce, retry with interval, bounded final flush on close, and the `persisting`/`mutated_while_persisting` flags also fix the dirty-clearing race the old code had.
- Poisoned bulk projections → `MeasurementProjection`/`MeasurementPage`/`PruneReport` isolate per-peer failures.
- The `u64::MAX` sentinel, the boolean plumbing, the 4× event↔counter mapping, the `take_next`/`take_other` duplication (now Kind + `find_map` + `MeasurementBatch`, which also fixes the per-event byte distortion), the speculative `Authentication` axis (now derived from `pending_attempt` + connection admission, with a retained-peer bound), the mixed import provenance, the unbounded RPC listing (limit/cursor with default 100 / max 1000), the const placement, the test sleep, the `NonZeroU64::MIN` fallback, the epoch-reset semantics (now documented as deliberate), and the double lookup — all resolved.

## Remaining / new issues

**1. `ReliabilityWindowMismatch` has no migration path and recreates a fail-closed startup — `crates/node/src/measure.rs:177`**

Snapshots now store `window_seconds` per record and `ensure_policy` returns a typed error on mismatch — but the adapter never migrates. `new_with_clock` propagates `ledger.reconcile_clock(now, reliability_policy())?`, so any future release that changes `DURATION` (the reliability window) makes every node with an old snapshot **fail to start**; and even past startup, every `record()` for such a peer fails forever via `observe_batch → ensure_policy`. The window holds one hour of advisory evidence — resetting it on mismatch is a trivially safe migration. This is the same class of problem this round just fixed for clock regression: "needs migration" encoded as "refuses service".

**2. `record_batch` as a required method forces identical boilerplate on every implementation**

Five test implementations each hand-write the same "ignore if `Unauthenticated`, loop `occurrences` times over `record`" body (`transport/tests/mod.rs`, `test_chunk_e2e.rs`, `test_outbound_scheduler` ×2, `test_stabilization`). The rationale for making it required is presumably atomicity, but in practice every non-ledger implementation chose the non-atomic loop — the abstraction is fighting its users. A provided default (documented as non-atomic for counter-only implementations) removes ~100 duplicated lines. Related: `incr` is documented as "Increment a legacy counter for an already-authenticated peer" yet now takes an `Authentication` parameter — the doc and the signature contradict each other. If `incr` is legacy, keep it narrow and let `record` do the gating.

**3. Expired-on-arrival chunks now produce no outcome at all — `crates/core/src/chunk/reassembly.rs:731`**

`Rejected::Expired` moved from the Invalid group to the Replay group: a peer that streams already-expired chunks is never charged `FailedToReceive` (no pending is created, so no TTL-expiry charge either). This may be consistent with the one-outcome-per-transmission model, but it is now the only adversarial input that consumes bandwidth while leaving zero evidence. It deserves a comment or a test making the choice explicit rather than a silent regrouping in the classification table.

**4. The ledger-full policy is implicit — `crates/measure/src/ledger.rs`**

`RetainedPeerLimitExceeded` is reject-new / retain-old rather than evict-stalest-by-`last_seen`. With 150-day retention and a 16,384 bound, a long-lived hub node can realistically hit the cap; from then on every event for every new peer logs at error level in `record_measurement` (log storm), and new peers can never accumulate credit until old records age out. At minimum this should be warn-once; evicting the record with the earliest `last_seen` is the better policy. Either way, state the decision.

## Minor

- `PeriodicMeasure::new` now implicitly requires a live async runtime (it spawns the persistence worker at construction) — the FFI change exists precisely because this was violated. The precondition belongs in the constructor's docs.
- `persist_final_with_retries` makes its three retries back-to-back with no spacing, which only helps for immediate races. If that is deliberate (bounded shutdown must not sleep), a one-line comment would prevent the next reader from "fixing" it.
- The very first event after startup also waits the full 60s debounce before its first write (the interval test pins this behavior). Losing up to a minute on crash is a reasonable trade — just confirming it is intentional.
- Scope note: the `try_sleep`/`InboundTimerUnavailable` hardening and the registration sibling-stop are collateral infrastructure pulled in by the close-time retention loop and the shutdown hang — justified, but the PR description still describes round one. The breaking `Measure` trait changes (`Authentication` parameter on `incr`/`record`, the new required `record_batch`) and the paginated `ListPeerMeasurements` request shape belong in the breaking-changes section.

The heaviest item is #1 — same fail-closed-startup class as before, and it should be fixed. #2 is an abstraction-quality issue. #3 and #4 need an explicit design decision, even if the decision is the status quo, written down.

RyanKung:
## Review — db9c8e71 ("resolve attribution and migration review")

Fresh full read of the new commit. All checks are green on this head.

### The interrogation was answered properly

`Authentication::LocallyAddressed` is the right shape of answer — a typed decision, not a patched constant. `permits()` puts the attribution rule into the model itself (`Authenticated` permits everything; `LocallyAddressed` permits only `FailedToSend` and is documented as never authenticating received bytes or credit; `Unauthenticated` permits nothing), all three missing-connection sites use it, `ApplyOutcome::IgnoredUnattributable` names the new semantics, and `test_missing_connection_records_locally_addressed_failure` pins the behavior end to end. This is exactly what was asked for.

I also verified the remaining resolutions in the implementation rather than the reply: `good()` removed with its contradictory doc; both `peer_qualities` sites now consume the projected `measurement.quality`; the RPC limit parses once into `NonZeroUsize`; FFI destroy flushes on a dedicated thread before `shutdown_background()` with a durable-storage witness for the destroy-without-listen path; `RuntimeUnavailable` typed error with a captured runtime handle and constructor doc; `ReliabilityPolicy` keeps its `NonZeroU64` proof; expired-on-arrival is now Invalid-once with `LogicalTransmission (uuid, ts, ttl)` terminal identity — better than either prior state, since exact-once scoring no longer blocks legitimate UUID reuse after TTL; the hardcoded `Authenticated` expiry site carries its precondition; `count()` logs projection errors; `DEFAULT_MAX_RETAINED_PEERS` is a `NonZeroUsize` const; the window migration resets reliability while preserving credit, and the full-ledger LRU eviction has N/N+1 coverage.

### New finding: LRU eviction × LocallyAddressed record creation erodes the 150-day retention semantics

This is the one substantive item this round. Three consequences chain together:

1. A `LocallyAddressed` `FailedToSend` **creates a ledger record for a DID that was never observed**, and `transition_batch` calls `credit.touch(at)` for failure events — so a dead DID we keep retrying **never expires**: our own failed attempts refresh `last_seen` forever, and `prune` can never reach it. That inverts the intent of "expire after 150 days without observation", where observation was meant to be evidence of the *peer's* behavior.
2. At the 16,384 bound, a successful new-peer transition evicts by earliest `last_seen` **with no regard for credit value**. Local addressing churn — a burst of sends to dead DHT-derived targets — can evict a 149-day-old record carrying a 10.0 credit score, which is precisely the record the aMule retention window exists to protect. The eviction doc says it "keeps active peers measurable"; the evicted victim is by definition the least-recently-seen, i.e. the retention window's beneficiary.
3. `CreditRecord::last_seen`'s doc still reads "Most recent **authenticated** local observation" (`credit.rs:119`) — now contradicted by `LocallyAddressed` updating it.

Possible resolutions, in roughly decreasing preference:
- `LocallyAddressed` events do not touch `credit.last_seen` (reliability keeps its own epoch timeline; retention stays a statement about peer-evidenced observation);
- or a purely local failure does not create a record for an unknown peer (it only updates existing records — its ordering value is for peers we already know as DHT candidates);
- or eviction prefers credit-free records first (`min by (has_credit, last_seen)`);
- at minimum, document the trade-off on `DEFAULT_MAX_RETAINED_PEERS` and fix the `last_seen` doc.

This is not a bug in what shipped; it is the follow-on decision that introducing `LocallyAddressed` makes necessary — what does `last_seen` mean now, and what does the bound protect?

### Smaller items

- Terminal-witness saturation edge: when a pending entry expires, its failure is counted and its `LogicalTransmission` witness is inserted — unless the bounded set is full, in which case the insert is dropped. If the set later drains and a late chunk for that same transmission arrives, it takes the Expired → Invalid path and is charged a second failure. Bounded, adversarial-timing-only, and it damages only the adversary's own record — but the comment "fails closed to Replay, preserving exact-once attribution" is not strictly true across a saturate-then-drain sequence; worth one clarifying sentence.
- The `Arc::try_unwrap` failure branch in `ProviderHandle::shutdown` still leaks via `mem::forget` without a comment stating why no other owner can remain at that point (carried over from the previous round; minor).

Item 1 is a design decision to make explicitly, not a defect to hotfix — either direction is defensible once stated and witnessed.

RyanKung:
Re-reviewed `db9c8e71` with a fresh full read (~2,100 diff lines) plus cross-checks on the branch.

## Status of the previous round

All four of my round-two items, and the eight items from the parallel review, are substantively addressed:

- **Window migration**: `reconcile_runtime` is now infallible; a window mismatch resets only the short-term reliability evidence while fully preserving byte credit, and the startup test witnesses both the migration and the persisted dirty state. The fail-closed startup path is gone.
- **`record_batch` default**: the non-atomic counter-only default is provided with its contract documented, the five duplicated test bodies are deleted, `incr` is narrow again (`did, counter`), and the gating lives in `record` — the doc/signature contradiction is resolved.
- **Expired-on-arrival**: now `Invalid` once, then tombstoned to `Replay` keyed by `LogicalTransmission` (id + ts + ttl) — no longer evidence-free, without blocking legitimate UUID reuse after TTL; `remove_expired_at` also records expired pendings as terminal witnesses so a late chunk cannot be double-charged. The classification move now carries its rationale.
- **Ledger-full policy**: `RetainedPeerLimitExceeded` is replaced by deterministic LRU eviction (`last_seen`, peer-key tie-break) with N/N+1 coverage.
- From the parallel review: `Authentication::LocallyAddressed` with the `permits()` predicate, the dead `good()` API removed (no callers remain — verified by grep), route readers consuming the projected `quality`, the single `NonZeroUsize` validation chain in `rpc_impl`, the FFI destroy-without-listen flush with a durable-storage witness, the captured Tokio handle with a typed `RuntimeUnavailable`, and `count()` logging projection errors. The documentation minors (debounce, shutdown retries, runtime precondition) are all in.

## New findings

**1. `LocallyAddressed` × LRU eviction composes into a credit-ledger washout channel**

Two individually reasonable decisions combine badly. `LocallyAddressed` `FailedToSend` **creates a new ledger record** for a DID with no attributable history whatsoever (`locally_addressed_failure_is_attributable_without_remote_authentication` explicitly witnesses creation), and `apply_batch` at capacity unconditionally evicts the record with the earliest `last_seen`. So 16,384 doomed sends to throwaway DIDs wash out the entire 150-day credit ledger: every garbage record carries `last_seen = now` and is therefore the newest, while the evicted records are exactly the ones holding real credit. The property proven by `authentication_tests` — an unattributable identity flood cannot consume retained-peer capacity — does not hold for a `LocallyAddressed` flood.

Reachability: trivially local (`send_direct_message` to an arbitrary DID — the new delivery test is itself the reproduction). Remotely it is conditional but real: the missing-connection failure target is `payload.relay.next_hop` (`payload_send.rs:559`), `records_missing_connection_failure` covers every non-storage-route Detached kind, and finger-table entries without live connections are influenced by network input.

Fix, either of: (a) `LocallyAddressed` may only **update an existing record, never create one** — its semantics (charging a failure to a peer we already know) survive intact, and a failure-only record for an unknown DID carries no credit information anyway, it is pure pollution; or (b) eviction prefers zero-credit records. (a) is the cleaner invariant. Related: **eviction is currently silent** — a 150-day record is dropped and the adapter cannot even log it. Return the evicted key from `apply_batch` (in the outcome or alongside it) so the adapter can emit a warning.

**2. The `NonZero*::MIN.saturating_add(N-1)` idiom trades readability for lint surface**

`DEFAULT_MAX_RETAINED_PEERS = NonZeroUsize::MIN.saturating_add(16_383)`, `RELIABILITY_WINDOW = NonZeroU64::MIN.saturating_add(3_599)`, and a series of `MIN.saturating_add(1)` in tests meaning "2". The reader must do arithmetic to learn the actual values 16 384 / 3 600 / 2. The previous `match NonZeroU64::new(3600) { Some(w) => w, None => panic!(...) }` form is a compile-time check in const context and was fine; Rust 1.83+ also has const `unwrap`. Please spell the numbers.

**3. Minor**

- `MeasureState.runtime_handle` is held for the adapter's lifetime: if the owning runtime stops before the measure is dropped, a later `flush()` panics in `handle.spawn`. The FFI shutdown ordering already flushes first, but at the library level this is an implicit precondition — one line in the `flush` docs.
- `record_locally_addressed_send_failure_for_test` bypasses the transport to poke measurement state directly. Acceptable as a test-only hook, but note that adopting fix (a) above changes what this fixture must do, and the FFI destroy-flush test with it.

#1 needs a fix — it is a behavior introduced by this round's combination of changes. #2 is a readability regression.

RyanKung:
Re-reviewed at `26afa0ed` — this round as a full read of the entire PR diff (base..head, ~12,900 lines across 72 files), reviewing the composed final state rather than the commit delta.

## Previous round, verified in the final state

1. **The `LocallyAddressed` × LRU washout channel is closed, and closed at the type level.** `Authentication` gained two named predicates — `establishes_peer()` / `refreshes_peer_observation()`, both true only for `Authenticated` — and `apply_batch` returns `IgnoredUnknownPeer` for a new peer without an establishing proof. `LocallyAddressed` can neither create a record, nor trigger eviction, nor refresh `credit.last_seen` (the `(false, _)` arm only validates ordering, so a peer kept "alive" solely by local failures still prunes on schedule — witnessed by a test). Eviction is no longer silent: `ApplyReport` carries the evicted key and the adapter logs it. The `u8::MAX`-scale double flood test (Unauthenticated + LocallyAddressed) proves neither capacity nor existing credit is affected.
2. **The NonZero constants** are now `NonZeroUsize::new(16_384).unwrap()` with a narrowly-scoped `#[allow(clippy::unwrap_used, reason = ...)]`. To be explicit: I consider this allow acceptable. The unwrap is const-evaluated — failure is a compile error — so the lint is a false positive here, and smallest-scope-plus-documented-reason is exactly the right treatment. It is strictly better than the previous `MIN.saturating_add(N-1)` arithmetic.
3. The Tokio-runtime precondition is documented on `new`/`flush`/`flush_with_timeout`.
4. The FFI fixture follows the semantic change (Authenticated `Connected`), and the destroy-without-listen flush witness is preserved.

Cross-round compositions I re-checked in the whole-PR pass and found sound: the expired-on-arrival → Invalid-once → `LogicalTransmission` tombstone → saturation-horizon state machine is self-consistent (the fail-open at saturation cannot be used to escape scoring — the first `max_completed_ids` failures were each charged, which already crosses the Degraded threshold); `incr` has exactly one production caller left (the trait default bridge); per-frame envelope-expiry scoring for chunk frames matches pre-PR semantics, so it is not a regression.

## Remaining / new findings

**1. `SECURITY.md` does not cover the new credit/measurement subsystem** — the one substantive item this round. The threat-model document from #681 explicitly lists "reputation" among non-goals, and this PR introduces exactly an aMule credit + reliability subsystem. Two things belong in it:

- (a) A boundary statement: credit and reliability are **purely local advisory** values and never feed Chord membership, routing, ownership, or storage placement — this is precisely why this PR removed measurement-driven DHT eviction, and the threat model is where that boundary should be pinned down.
- (b) The residual Sybil surface: under LRU, completing 16,384 **real authenticated handshakes** (one per fresh DID) can still wash the entire credit ledger. The cost is significant, and it is consistent with the document's existing "identities are not scarce" assumption — but it should be stated as a known residual risk (the 1 MB activation threshold being credit's own Sybil mitigation).

**2. `ApplyReport` is in the wrong module** (`crates/measure/src/event.rs:141`). It is the result type of `apply`/`apply_batch`, its `pub(crate)` constructors are called only from `ledger.rs`, and its siblings — `MeasurementPage`, `PruneReport`, `LedgerReconciliation` — all live in `ledger.rs`. Move it there. (Minor.)

**3. The `Measure::record` default doc** says "compatibility bridge for counter-only **test** implementations" — but the default equally binds counter-only production implementations by external library users, and for those, a `LocallyAddressed` failure for an unknown DID **is** counted (the default has no retained-peer set; the following sentence already explains this). Drop the word "test" so readers do not assume production implementations are outside this contract. (Minor.)

Overall: the composed final state is coherent — the authentication axis, terminal-outcome accounting, LRU, clock/policy reconciliation, and persistence debounce each have their invariants witnessed by tests, and the whole-PR pass found no further behavioral defect in how they compose. What remains is the threat-model documentation gap (#1, should be added) and the two minors.

PR #697, opened 2026-08-28
Add native TUN gateway for transparent TCP over onion circuits

### LINKED ISSUE #684: Add a native TUN gateway for transparent TCP over Rings onion circuits
## Summary

Add a native IPv4/TCP gateway that forwards **explicitly selected traffic** through Rings Onion circuits. The gateway is an ingress adapter with an explicit route policy; a TUN/OpenVPN-style design does not inherently mean full-tunnel routing.

Traffic reaches the shared Onion TCP egress through one of these operator-selected mechanisms:

1. an exact destination address or prefix routed to the TUN/utun/Wintun interface;
2. an external router or policy-routing rule that directs selected traffic to the gateway; or
3. an explicit application proxy such as SOCKS5 or HTTP CONNECT.

The gateway must never install a default route, split a default route into `/1` routes, accept any prefix set whose union captures the entire IPv4 address space, capture IPv6, or mutate system DNS. OpenVPN-style route management is not the problem: OpenVPN supports both selected routes and an explicit `redirect-gateway def1` full-tunnel policy. Rings intentionally has no `redirect-gateway` or full-tunnel mode. Full capture would also capture ordinary host-network transport carrying existing Rings neighbor links and would require dynamic bypass state or guaranteed relay. The full-coverage rejection is an intent stop-line, not a recursion proof: near-cover sets such as all IPv4 except one `/32` remain valid and use the same fail-closed behavior after packets enter the gateway. TURN may be configured, but no gateway or server capability may require it; STUN remains ordinary discovery. Rings also does not wrap systemd, launchd, Windows Service Control Manager, or other service managers.

This issue is broader than #665. That issue covers a controlled WebView/HTTP gateway; this issue owns the native packet ingress, shared Onion TCP connector, explicit route ownership, cleanup, platform bindings, and gateway health.

## Product decisions

These decisions supersede the earlier daemon-service proposal and closed PR #685, and correct the earlier full-host-capture design in this issue:

- `rings run` remains the foreground, composable process primitive.
- Rings does not install or generate service-manager configuration.
- All gateway core code, platform bindings, Unix configuration, and cleanup leases live in `crates/gateway`.
- The gateway has no full-tunnel or kill-switch mode; it captures only operator-selected destinations.
- STUN remains ordinary ICE discovery. TURN is optional, and no gateway or server capability may depend on it.
- Linux, macOS, and Windows native targets provide client + server + gateway capability.
- WASM remains client-only and must not compile, link, or export gateway/server code.
- iOS NetworkExtension and Android VpnService integration are out of scope for this iteration.
- Swift and Kotlin/JNA desktop FFI examples remain in scope and align with the Python C FFI example.

## Traffic-selection contract

Let `C` be the normalized set of destination prefixes explicitly supplied by the operator.

```text
installed_destination_capture_routes = C
```

The platform binding must not add any other capture route. In particular:

- no `0.0.0.0/0` and no pair or larger set of routes whose union covers all IPv4, including the two-`/1` `def1` form;
- no IPv6 capture routes;
- no capture route derived from DNS, TURN, ICE, SDP, bootstrap endpoints, or observed peers;
- no runtime RPC or remote message may grant route authority;
- interface addresses must be IPv4 `/32` host addresses so they cannot create an implicit connected capture prefix;
- an already existing exact or more-specific route inside a capture prefix is a conflict because longest-prefix routing would shadow gateway capture; and
- an empty capture set is valid for externally steered packet operation.

The fail-closed guarantee is scoped to traffic that has already entered the gateway: a selected flow either opens its immutable target through a valid Onion route and compatible exit, or fails without direct fallback. The gateway does not claim to prevent unselected host traffic from using the host's ordinary network path.

At establishment, one consistent route-table snapshot rejects exact and more-specific routes that would shadow a configured capture prefix. Rings deliberately does not subscribe to system route changes. A route added after the snapshot, including while establishment finishes, can therefore shadow capture and send that sub-range through the host route; operators must prevent or monitor such mutation.

Rings signaling and WebRTC underlay keep their normal routing behavior. SDP offers and answers continue to travel through Rings message routing. STUN remains ordinary discovery, and TURN remains optional: no gateway or server capability may require it. Selective capture does not require global relay-only ICE, candidate-derived host routes, HTTP bootstrap pinning, or a dynamic underlay-bypass authority.

## Ingress and shared egress

All ingress forms converge on one target and one Onion connector:

```text
explicit route + packet target ----\
SOCKS5 target authority ------------> immutable target -> Onion TCP stream -> advertised TCP exit
HTTP CONNECT authority -------------/
optional synthetic DNS mapping -----/
```

- **TUN route ingress:** preserve the original destination IP and port from captured packets. Install only configured prefixes.
- **External-router ingress:** the operator owns router/policy-routing configuration. Rings must not assume authority over that external system.
- **SOCKS5 / HTTP CONNECT ingress:** the application supplies the target hostname/IP and port directly, so no system route or DNS mutation is required.
- **Optional DNS integration:** pointing multiple hostnames at one gateway address loses the original generic TCP target. A DNS ingress therefore needs a dedicated synthetic prefix and an explicit `synthetic IP -> original target` mapping. It is an adapter, not system DNS ownership.

The existing native HTTP CONNECT ingress and the packet gateway must reuse the same Onion target validation, route construction, stream opening, transfer, half-close, timeout, and accounting behavior. A SOCKS5 frontend should add only protocol parsing and authentication policy, not a second egress implementation.

### One native gateway crate

The shared gateway data plane and platform effects remain under `crates/gateway`:

```text
crates/gateway/
├── bin/gateway-config-unix.rs
└── src/
    ├── config.rs                 validated explicit capture plan
    ├── packet.rs                 PacketIo
    ├── flow.rs / flow_table.rs   bounded per-flow state
    ├── tcp/                      shared smoltcp IPv4/TCP data plane
    ├── bridge.rs                 bounded bidirectional byte bridge
    ├── stream.rs                 OnionStreamConnector boundary
    ├── runtime.rs / server.rs    orchestration and lifecycle
    ├── status.rs                 process-independent health
    └── bindings/                 Linux, macOS, Windows effect boundaries
```

Platform code owns only explicit interface, route, and cleanup effects. The shared server, flow, TCP, admission, backpressure, and Onion mapping logic contains no Linux/macOS/Windows system API calls.

### Core contracts

- `PacketIo`: receive and inject complete IP packets.
- `GatewayPlan`: declarative addresses, MTU, and exact capture prefixes.
- `TunnelControl`: reconcile stale state, establish the packet device and exact capture routes, and consume a linear cleanup lease on teardown.
- `OnionStreamConnector`: bind one admitted immutable target to one Onion TCP stream without exposing Rings internals to the gateway crate.

### Lifecycle and safety invariants

1. A flow admitted by the gateway never opens a direct public connection from the client node.
2. If no valid Onion route or compatible exit exists, the selected flow fails without direct fallback.
3. A flow target is immutable for the lifetime of its Onion stream.
4. Installed destination capture routes equal the normalized operator-declared capture set; interface addresses are `/32` and cannot widen it.
5. Rings control, SDP, bootstrap, and ICE traffic receive no special gateway route authority.
6. Gateway failure remains observable independently from the `rings run` process.
7. Stop, restart, startup failure, and stale-state reconciliation are idempotent and clean only resources owned by the gateway lease.
8. UDP and IPv6 forwarding remain explicit non-capabilities for this milestone; unselected UDP/IPv6 traffic is outside gateway policy.
9. Existing exit target policy remains authoritative at both gateway admission and exit connection.
10. If runtime processing and shutdown both fail, diagnostics retain both failures and cleanup still attempts every owned resource.

### Unix privilege boundary

- `gateway-config-unix` is a separately launched foreground resource helper; it never invokes privilege elevation or a service manager.
- The helper authenticates one node UID, transfers the packet descriptor with `SCM_RIGHTS`, and owns only the explicitly requested interface/routes/lease.
- A normal teardown cleans the lease. Disconnect or malformed control traffic retains the packet descriptor and selected routes fail-closed; the same plan may reconnect and resume the lease.

### Linux, macOS, and Windows

- Linux uses `/dev/net/tun`; macOS uses the built-in `utun`; Windows uses Wintun.
- Each platform installs only explicit capture prefixes and records them in a recoverable lease.
- Platform runtime behavior must be proven by native smoke tests; cross-compilation alone is not acceptance evidence.
- External router configuration, system DNS mutation, service installation, and supervisor policy remain outside the helper.

## Revised implementation roadmap

Status snapshot: 2026-09-01. PR #697 is open from the isolated sibling worktree on branch `codex/tun-gateway-684`. The selective-ingress implementation landed in `2f3f1ec2`, the OpenVPN mechanism-versus-policy terminology was corrected in `7042a145`, equivalent IPv4-covering prefix sets are rejected in `51bcdf06`, and route-shadow review fixes landed in `7f8af29b`. Hosted CI on the current head completed with 15 successful checks and one conditionally skipped matrix check.

- [x] Create the runtime-neutral gateway, flow model, bounded TCP data plane, Onion connector, lifecycle, and status model.
- [x] Implement Linux TUN, macOS utun, and Windows Wintun bindings with recoverable cleanup leases.
- [x] Integrate the optional foreground gateway with `rings run` without service-manager wrappers.
- [x] Keep WASM client-only.
- [x] Add Swift and Kotlin/JNA desktop FFI examples aligned with Python.
- [x] Remove default-route/full-tunnel mode and reject every IPv4 covering prefix set; TURN remains optional and never a gateway prerequisite.
- [x] Make exact operator-declared prefixes the complete destination-capture authority; require `/32` interface addresses and allow an empty set for external steering.
- [x] Reject pre-existing exact and more-specific route conflicts from one establishment snapshot; document post-snapshot shadowing as an operator-monitored hazard.
- [x] Remove mandatory TURN, relay-only ICE, dynamic candidate admission, TURN refresh, and gateway-specific HTTP bootstrap pinning.
- [x] Scope fail-closed status, documentation, and tests to explicitly selected traffic.
- [x] Re-run corrected-model unit tests, the six A-B-C routed-SDP tests, native strict lint, WASM dependency isolation, Linux/Windows gateway cross-checks, and the Swift FFI smoke test locally.
- [x] Complete privileged TUN/utun/Wintun and public Onion integration jobs in hosted CI on the corrected commit.

## Acceptance criteria

- Installing a plan produces exactly its normalized explicit capture prefixes and no hidden default or IPv6 routes.
- Establishment rejects an exact or more-specific existing host route that would shadow any configured capture prefix.
- A real explicitly captured IPv4/TCP flow reaches a public test server only through a multi-hop Rings Onion route and TCP exit.
- Removing the exit or breaking Onion route construction fails that selected flow without direct egress.
- Rings overlay/bootstrap/SDP/ICE continues to use ordinary host routing without a gateway-specific global transport policy.
- Application-proxy-only operation starts no packet gateway. Existing HTTP CONNECT uses the same Onion connector and exit policy as packet ingress; SOCKS5 is a future ingress frontend.
- TCP data, EOF, half-close, reset/error, backpressure, retransmission, timeout, MTU, and connection limits remain covered.
- Linux TUN, macOS utun, and Windows Wintun each pass native runtime smoke tests for explicit capture and cleanup.
- RPC/inspection distinguishes process health from interface state, explicit capture set, exit availability, flow count, and gateway lifecycle.
- WASM remains client-only and contains no gateway/server dependency or exports.
- Python, Swift, and Kotlin examples exercise equivalent C ABI ownership and provider lifecycle behavior.
- Documentation and benchmarks distinguish route mechanism from policy: the Rings gateway contract is selective ingress, has no full-tunnel or kill-switch configuration, and never requires TURN.

## Explicit non-goals for this iteration

- iOS NetworkExtension integration.
- Android VpnService integration.
- System DNS mutation or ownership.
- External router configuration.
- systemd, launchd, Windows SCM, or other service-manager wrappers.
- Full-tunnel, `redirect-gateway`, or kill-switch semantics.
- UDP over Onion circuits.
- IPv6 forwarding over Onion circuits.
- Private/LAN target access through public exits.
- Raw source-IP or end-to-end TCP packet identity preservation.

RyanKung:

Feature, native networking, platform bindings, CI, documentation, and desktop FFI examples.

Native Rings nodes expose Onion TCP streams and an HTTP CONNECT proxy, but do not provide one platform-neutral packet gateway for explicitly selected host IPv4/TCP destinations. Interface ownership, exact capture routes, stale cleanup, and gateway health are not modeled as one lifecycle.

Issue #684 defines the corrected scope. Closed PR #685 is intentionally not reused: Rings remains a foreground composable process and does not wrap systemd, launchd, or Windows service managers.

- Adds `rings-gateway`, one shared native IPv4/TCP data plane backed by smoltcp.
- Captures only destination prefixes explicitly selected by the operator. It rejects `/0` and any equivalent prefix set covering the entire IPv4 space; Rings has no full-tunnel or host-wide kill-switch mode.
- Requires IPv4 `/32` interface addresses so interface setup cannot create an implicit connected capture prefix.
- Reads one route-table snapshot before installing capture routes and rejects every pre-existing exact or more-specific route that would shadow a configured prefix under longest-prefix routing.
- Adds Linux TUN, macOS utun, and Windows Wintun bindings behind narrow platform effect boundaries.
- Adds a foreground `gateway-config-unix` privilege helper with peer authentication, `SCM_RIGHTS` descriptor transfer, durable exact-route leases, stale reconciliation, and disconnected-lease retention/recovery.
- Integrates an optional gateway task and status RPC into `rings run` without daemon lifecycle commands.
- Keeps Rings bootstrap, SDP signaling, and ICE on their ordinary routing path. STUN remains ordinary discovery; TURN is optional, and no gateway or server capability may require it. The gateway does not force relay-only ICE or derive route authority from peers or runtime RPC input.
- Keeps WASM client-only and rejects gateway dependencies from the browser graph.
- Adds Swift and Kotlin/JNA desktop C-ABI examples aligned with the existing Python ownership and provider lifecycle.
- Adds native, WASM, FFI, privileged Unix, and Windows CI coverage.

The packet core is platform-neutral. OS interface/route mutation remains under `crates/gateway/src/bindings`, and installed destination capture routes equal the normalized explicit configuration. Rings deliberately has no route-change subscription, so a host route added after the establishment snapshot can shadow a capture prefix; operators must prevent or monitor that mutation. Applications may instead use the existing HTTP CONNECT ingress or external router policy without granting Rings default-route authority; SOCKS5 remains a future ingress frontend to the same Onion egress.

This follows the OpenVPN mechanism-versus-policy distinction rather than rejecting its route machinery: selected `route` entries and `redirect-gateway def1` full-tunnel redirection are different configurations over the same TUN mechanism. Rings intentionally implements only selected-prefix routing and has no `redirect-gateway` or full-tunnel mode. Full capture would also capture ordinary host-network transport carrying existing Rings neighbor links and would require dynamic bypass state or guaranteed relay. The full-coverage rejection is an intent stop-line rather than a recursion proof: near-cover prefix sets remain valid and rely on fail-closed behavior after traffic enters the gateway. TURN may be configured, but it is never a prerequisite.

No existing gateway behavior is replaced. The native gateway is opt-in. Operators must explicitly configure every destination prefix they want the packet gateway to capture; a `/0`, an equivalent IPv4-covering prefix set, and every IPv6 route are rejected. WASM remains client-only. iOS NetworkExtension and Android VpnService integration are outside this iteration.

The earlier full-host-capture model has been removed. `GatewayPlan` now contains only `/32` interface addresses, exact `included_routes`, and MTU; `0.0.0.0/0`, any equivalent prefix set covering all IPv4, IPv6, non-`/32` interface prefixes, removed plan route-authority fields, and pre-existing exact or more-specific shadowing routes fail closed. The unused server-request timeout classification was removed; real client response timeouts remain enforced by the configured Unix control-socket deadline. The implementation retains the shared TCP flow engine, Onion connector, TUN/utun/Wintun bindings, foreground Unix helper, cleanup ledger, status model, WASM boundary, and Python/Swift/Kotlin FFI coverage.

Local corrected-model evidence:

- `cargo test -p rings-gateway --all-targets`: 79 library tests and the helper CLI test passed; two privilege/public-network tests remained explicitly ignored locally.
- `cargo test -p rings-node -p rings-transport --lib`: 309 Node tests and 93 Transport tests passed; the public three-node gateway integration remained explicitly ignored locally.
- Six Core three-node connection tests passed and observed SDP offers and answers routed through the intermediate Rings node.
- Strict native Clippy passed for Gateway, Node, Core, and Transport; strict WASM browser Clippy passed and the browser dependency graph contains no gateway/TUN/route-manager dependency.
- Gateway cross-checks passed for Linux and Windows targets. The local host lacks the C cross sysroots needed to finish full Node cross-checks, which remain covered by hosted CI.
- Swift FFI smoke and Node FFI Clippy passed. Kotlin/JNA remains covered by the existing hosted Linux job because Gradle is unavailable on this host.
- Nightly formatting and `git diff --check` passed.

Privileged TUN/utun/Wintun and public Onion integration are not claimed from the local run. Hosted CI completed on `7f8af29b83d8e2ce5a1770cc16e53972d46074c7` with 15 successful checks and one conditionally skipped matrix check, including Linux TUN/helper/kernel-TCP integration, macOS utun, Windows Wintun, public two-hop Onion egress, WASM, Miri, FFI sanitizers, dependency policy, and release builds.

Closes #684


RyanKung:
I read the full `rings-gateway` crate (core, runtime, bindings, helper, Windows route binding), the node/transport/core integration diffs, the CI matrix, and the tests. The layering is genuinely good: the gateway crate has zero rings imports, the `OnionStreamConnector` / `TunnelControl` / `UnderlayPolicy` effect boundaries are well cut, and the helper's privilege model (peer-uid auth, mode-0600 socket, ancestor-chain path validation, write-ahead route ledger) is careful work. The pure lifecycle models are the cleanest part of the PR — and notably, none of the defects below live in them. They all live where the model stops: the imperative shell of the runtime loop and the capability shapes around it. I've grouped the findings by structural root cause rather than by file, because several of them are the same mistake wearing different clothes.

## 1. Blocker: any single bad packet kills the gateway — and then the node

`process_input` propagates every error from `ingest_packet` with `?` (`crates/gateway/src/runtime.rs:305-315`), the `run()` loop breaks on any `Err` → `GatewayEvent::Fail` → all flows aborted → gateway task exits; then `rings.rs`'s `select!` sees the gateway task finish and calls `stop.request_stop()` — **the whole node shuts down**. The following per-packet events all take that path, and all of them are routine in the wild:

- **A fragmented IPv4 packet**: `classify_ipv4_packet` returns `Err(FragmentedIpv4)` (`crates/gateway/src/packet.rs:76-78`). With a captured default route, any application emitting a >MTU UDP datagram (legacy EDNS0 DNS, NFS, etc.) produces one. The README says "fragmented IPv4 fail closed", but the actual semantics is "the entire node fails".
- **A non-SYN packet for an unknown flow** → `MissingInitialSyn` (`crates/gateway/src/tcp/mod.rs:250-252`). This is guaranteed during normal operation: in-flight client segments arriving after we RST a flow; a client retransmitting its FIN because our final ACK was lost; any stray packet toward a released flow.
- **The SYN that exceeds `max_flows`** → `TcpCapacityExhausted`. Capacity exhaustion should refuse *that flow*; instead it kills the node.
- Garbage with a version nibble other than 4/6 → `Err(NonIpv4)`, equally fatal.

This is a trivially exploitable DoS (any host on the same segment — or any local process — can trigger it) and also a self-destruct path under perfectly normal traffic. None of the runtime tests (`crates/gateway/src/runtime/tests.rs`) cover stray packets, capacity exhaustion, or fragments, which is why it never surfaced.

**Root cause: partiality is used to encode a total classification.** "Drop this packet" is split across both sides of a `Result`: `Ok(DropUdp)` on one side, `Err(FragmentedIpv4 | Malformed… )` on the other. But `Err` is not a data category — it belongs to the caller's short-circuit monad, and `?`'s bind semantics is *abort*. Putting a packet-level fact into `Err` automatically promotes it to a computation-level termination. The classifier's codomain should be total:

```
PacketDisposition ::= Tcp(segment) | Drop(DropReason)
DropReason       ::= Udp | Fragmented | Malformed | NonIpv4 | …
```

With no `Err` branch, `?` cannot exist syntactically and the escalation becomes unrepresentable. `MissingInitialSyn` / `TcpCapacityExhausted` / `ListenRejected` are likewise ordinary values of the flow-admission codomain ("reject, emit RST"), not errors.

## 2. The error algebra has no grading, so `?` acts as an implicit scope-escalation operator

`GatewayError` is one flat sum covering four distinct blast radii — packet-scoped, flow-scoped, gateway-scoped, node-scoped — so `?` silently upcasts a small-radius failure into a large-radius one at every layer. Finding 1 is the worst instance; another is:

- **A single flow failure rewrites global health**: `BridgeEvent::Failed` (e.g. the exit refusing one target's port — a purely per-target event) sets `exit_availability = Unknown` with that flow's error string (`crates/gateway/src/runtime.rs:405-412`), flipping the status endpoint to Degraded until the next refresh. A per-flow event and an exit-wide event are different scopes stored in one global scalar. At minimum, only `OnionUnavailable` (route construction failure) should be allowed to touch availability, not any `StreamIo`.

Suggested shape: make the loop's type `Result<PacketOutcome, GatewayFatal>` where `PacketOutcome ::= Consumed | Dropped(DropReason) | FlowRejected(FlowId, RejectReason)`. Packet- and flow-scoped outcomes live on the `Ok` side and structurally cannot break the loop; only device-level IO failures and internal invariant violations inhabit `GatewayFatal`. Findings 1 and this one then become type errors rather than review findings.

## 3. Liveness was never discharged: a Listen socket can leak its slot forever

The `FlowState`/endpoint models are verified for *safety* (transition legality) but nobody proved *liveness* — "every non-terminal state eventually leaves the table". That theorem needs a named exit-event source per state: Established has the idle timeout, SynReceived has smoltcp's SYN-ACK retransmission exhaustion, but **Listening has nothing**. A SYN whose checksum smoltcp rejects creates the socket in `ensure_flow` and then parks it in `Listen` permanently; `reconcile` does nothing for states that are neither `Established` nor `Closed`, and `flow_idle_timeout` never fires for Listen. Each such packet permanently consumes one of `max_flows` slots until restart. Low probability (needs raw-socket injection), but the invariant break is structural.

Fix: turn the obligation into data — `fn exit_clock(state: TcpEndpointState) -> Option<Duration>` — and have the Tick reaper enforce it for every state, Listening included. That moves "every state has a way out" from a comment into a tested function.

## 4. `StopToken` models an event as a predicate, and the shutdown latency falls out of that shape

`refresh_underlay_and_exit` only observes `stop` / `runtime_done` *after* `ticker.tick().await` (`crates/node/src/native/gateway.rs:339-343`), and `StopToken` offers only `should_stop() -> bool`, so polling at the ticker's cadence is the only option. `tokio::join!(run, update)` therefore cannot return until the next tick: invisible at the 2 s default, but `underlay_refresh_secs` is an arbitrary `u64` — configure it above 30 and the 30-second timeout in `rings.rs` calls `gateway_task.abort()`, tearing down the cleanup path mid-await and reporting a misleading "gateway did not finish route cleanup" (on Windows the routes then really do dangle until the next reconcile).

Stop is a one-shot event; modeled as a state observation, every consumer is forced to bring its own clock. Add an awaitable form (`async fn stopped(&self)` backed by `Notify`) so consumers `select!` instead of poll — this bug class disappears wholesale.

## 5. `BridgeHandle`'s linearity is a convention, not a type — and the orphan task busy-spins

If the commands sender is dropped without `abort()`, `commands.recv()` returns `None` immediately, the match arm only sets `write_open = false`, and while `read_open` holds the `select!` completes instantly every iteration — a 100 % CPU spin (`crates/gateway/src/bridge.rs:141-151`). All normal paths call `abort()`, but finding 4's `gateway_task.abort()` manufactures exactly the "runtime dropped without `stop()`" scenario. The codebase already commits to linear-capability style elsewhere (`teardown(lease)` consumes by value); `BridgeHandle` applies the same discipline inconsistently. `impl Drop for BridgeHandle { fn drop(&mut self) { self.task.abort() } }` (or treating `None` as terminal) closes it.

## 6. ICE resolution: the check is not homomorphic with the aggregation

`resolve_ice_server_ips` decides "this URL has no IPv4" via `addresses.len() == before` (`crates/node/src/native/gateway.rs:403-410`). "Every URL resolves to ≥1 IPv4 address" is a per-element property, but it's tested *after* an idempotent set union — a lossy fold that swallows multiplicity. Two URLs for the same host (`stun:` + `turn:` on one server, a completely legitimate config) make the second `extend` a no-op → spurious "ICE server has no IPv4 address" → the gateway cannot start. Test whether *this URL's* resolution yielded any IPv4 before merging, not the aggregate's cardinality delta.

## 7. Smaller items

- **`#[allow(clippy::too_many_arguments)]`** on `refresh_underlay_and_exit` (`crates/node/src/native/gateway.rs:328`): new code; per project policy this should be a parameter struct, not an allow.
- **`dns_policy: block` silently drops TCP:53** (`crates/gateway/src/runtime.rs:341-345`): clients hang to timeout instead of getting an RST. Fail-closed, but an RST matches "reject ordinary DNS" better.
- **Hot-path efficiency**: every packet is classified twice (`runtime.rs:341` and `tcp/mod.rs:122` — `TcpStack::ingest_packet` should accept the already-classified segment); `socket()`/`socket_mut()` do an O(n) scan of the `SocketSet` per TCP operation (`tcp/mod.rs:274-302`), on every read/write/state query; `commit_application_read` allocates a throwaway buffer to discard bytes (`tcp/mod.rs:184-195`) where smoltcp's closure-form `recv` consumes without copying; `drive_client_to_onion` allocates a fresh 16 KB buffer per flow per 10 ms reconcile and copies each chunk three times (peek → `to_vec` → commit-discard) (`runtime.rs:470-492`); `publish_status()` clones a String-bearing snapshot under a lock every loop iteration (`runtime.rs:268`) — publish on change.
- **`PacketIo::read_packet` has an implicit cancel-safety contract**: it sits inside `select!` (`runtime.rs:247`); an implementation that awaits internally more than once will silently lose packets. The trait docs (`crates/gateway/src/packet.rs:19-26`) should state the requirement.
- **Exposure: `/gateway/status` is also mounted on the external API** with permissive CORS (`crates/node/src/native/endpoint/mod.rs`), revealing interface name, DNS policy, active flow count, and raw error-reason strings to anything that can reach `external_api_addr`. Gateway health is host-operations data; internal-only seems right.
- **Model/implementation mismatches worth folding together**: two parallel capacity ledgers (`FlowTable` and `TcpStack.handles`, same `max_flows`, kept consistent by hand in the runtime); and `BuildRoute` → `Open` fired back-to-back (`runtime.rs:441-442`) so `RouteBuilding` is never observable — either merge the states or let the transition actually happen where route construction happens.

## Summary

The functional core is in good shape; the defects concentrate in the imperative shell, whose central semantic decision — *which errors terminate which enclosing scope* — currently exists only as a side effect of `?` placement, unnamed and untested. Suggested order: totalize `classify`'s codomain → introduce the `PacketOutcome` / `GatewayFatal` grading in the runtime loop → per-state exit clocks + reaper → awaitable `StopToken` → `Drop` for `BridgeHandle` → items under 7. The first two make the blocker class unrepresentable; 3–6 (findings 2–6) I'd also consider pre-merge; the rest are judgment calls.

RyanKung:
Re-reviewed 617b5135 as a fresh read of the rewritten files (`packet.rs`, `runtime.rs`, `tcp/mod.rs`, `bridge.rs`, `stop.rs`, `native/gateway.rs`, `endpoint/mod.rs`, all new tests), not just a closure check of the previous round. Verification of the fixes first, then three new findings from this commit.

## Fixes verified

1. **Total classification + graded loop** — `classify_ipv4_packet` now returns `PacketDisposition::{Tcp, Drop(PacketDropReason)}` with no error channel; the loop's type is `Result<ReconcileScope, GatewayFatal>` with packet drops and `FlowRejected` on the `Ok` side. I traced every path from `process_input`: stray non-SYN, capacity exhaustion, listen rejection, and blocked TCP/53 all end in `reject_segment` (the segment is fed to smoltcp with no matching listener, which emits the RST — a nice reuse) and cannot break the loop. The regression tests cover exactly the previously fatal cases, including the live-loop test that drops a malformed packet and proves a following SYN is still served.
2. **Health scoping** — only `OnionUnavailable` may touch `exit_availability`; `stream_io_failure_is_flow_scoped_not_exit_scoped` pins it.
3. **Liveness** — `TcpEndpointState::exit_clock` names the delegation explicitly (Listening/Handshaking → runtime deadline; Established/Closing → smoltcp timers; Closed → reconciliation), the Tick reaper enforces it, and the lazy deadline-clear in `expired_pending_flows` is correct (a deadline surviving into Established is cleared on first expiry check, and states cannot regress to Listening). Checksum validation moving before endpoint admission also closes the bad-checksum-SYN variant; IP-header-checksum failures (which `classify` does not verify) now land in the reaper instead of leaking — the two layers compose.
4. **Awaitable stop** — `StopToken::stopped()` uses the register-then-recheck pattern over `event_listener`, so a racing `request_stop` cannot be missed; the refresh loop selects over stop / runtime-done / tick, and `UnderlayRefresh` replaces the `too_many_arguments` allow.
5. **Bridge linearity** — `Drop` aborts the task, and a closed command channel is now terminal (shutdown + return) instead of a spin state. I also checked the new `ClientBuffer` round-trip for the tail-of-stream race: `drive_client_to_onion` peeks before testing `client_read_open`, and `CloseWrite` can only enter the capacity-1 command queue once the receive buffer has drained, so the defensive `Data`-after-close-discard arm is unreachable and no tail data is lost on client half-close.
6. **ICE resolution** — per-URL IPv4 check before the union; the shared-IP STUN+TURN case is regression-tested.
7. The rest as described: external API no longer mounts `/gateway/status` (README updated accordingly), `RouteBuilding` removed, capacity owned solely by `FlowTable`, single classification per packet, `commit_application_read` consumes via smoltcp's closure `recv`, status published on change, cancel-safety documented on `PacketIo`. The retained checked `SocketSet` scan with the in-code rationale (panic-free over smoltcp's panicking `get`) is a coherent tradeoff and it's now documented where the next reader will look.

## New findings in this commit

1. **Doc drift on TCP/53** (`crates/gateway/README.md:70`): "captured UDP plus TCP port 53 are dropped" — under `dns_policy: block`, TCP/53 is now *refused with a TCP reset* (`FlowRejectReason::DnsBlocked` → `reject_segment`), which is the better behavior, but the sentence still describes the old silent drop. UDP remains a drop, so the sentence needs to split the two.
2. **The grading blurs at one seam** (`crates/gateway/src/runtime.rs:283-284`): `reconciled.map_err(GatewayFatal::gateway)` immediately followed by `error.into_gateway()` is a wrap/unwrap no-op. Since reconcile-path errors are gateway-fatal by construction, either type `reconcile_flow`/`reconcile_all` as returning `GatewayFatal` so the grade is carried by the signature, or break with the plain `GatewayError` — the current round trip re-introduces, at exactly the boundary the grading exists to mark, the convention-over-type style the rest of the commit removed.
3. **TCP checksums are now verified twice per ingested packet**: once in `classify_tcp::verify_checksum` and again by smoltcp's interface poll, because `PacketQueueDevice::capabilities()` leaves `DeviceCapabilities::default()` checksum handling (verify-on-RX) in place. Setting `capabilities.checksum.tcp = Checksum::Tx` on the queue device would keep transmit-side checksum generation while skipping the redundant receive-side verification that `classify` has already performed — the IPv4 header checksum capability should stay as-is, since `classify` does not cover it.

CI is green on this head, including the privileged Linux TUN/helper/namespace jobs, the macOS utun job, and the Windows Wintun run. Findings 1–3 are small; nothing blocker-class surfaced in this pass.

RyanKung:
Reviewed `9991aa28`. All three follow-up findings are addressed correctly; no new findings in this diff.

1. **README** — the `dns_policy: block` sentence now states the two distinct outcomes (captured UDP dropped, TCP/53 refused with a reset), matching the implemented behavior.
2. **Grade carried by signature** — `reconcile_flow` / `reconcile_all` return `Result<(), GatewayFatal>` and the loop consumes that result directly; the wrap/unwrap seam is gone. The sub-helpers below `reconcile_flow` still return `GatewayError` and are wrapped per-call, but those wrap sites now sit immediately inside the function whose signature declares the grade, and I re-traced all four helpers to confirm every remaining error path there is invariant-class (map desync, commit mismatch, illegal transition) — fatal by construction. That's a reasonable place for the typed boundary to stop.
3. **Checksum delegation** — `Checksum::Tx` for TCP keeps transmit generation while skipping the receive verification `classify_tcp` already performed, and IPv4 header checksums stay delegated to smoltcp, which `classify` deliberately does not cover. I also confirmed the precondition holds on every ingress path: `reject_segment` is only ever fed segments that passed classification (DnsBlocked, MissingInitialSyn, rejected admission), so no unverified TCP bytes reach smoltcp with RX verification off. The capability regression test pins all four facts.

CI is green on this exact head across all 14 runs, including the privileged Linux, macOS, and Windows gateway jobs.

RyanKung:
Full re-review of the entire PR at `9991aa28` — every touched file, including the areas earlier rounds under-weighted (the Swift/Kotlin/C FFI examples and fixtures, `gateway-bench` in full, README/BENCHMARKS in full, the vendored crate's provenance, test-support code, Gradle build), plus cross-cutting passes (repetition, placement, exposure, doc consistency) over the whole diff rather than per-file reads. The blocker-class material from earlier rounds stays fixed; this pass surfaced a set of quality findings, none blocker-class.

## A. Duplication the crate should fold (in-crate, exact or near-exact)

1. **`first_ipv4_address` exists twice** — `crates/gateway/src/tcp/mod.rs:396` and `crates/gateway/src/bindings/native.rs:544` are the same projection of the plan. It is plan vocabulary and belongs as a method on `GatewayPlan` (fixing placement and the duplication at once).
2. **`platform_error` is defined three times** — identical free functions in `bindings/native.rs:814`, `bindings/unix/transport.rs:197`, `bindings/unix/helper.rs:361`. One `GatewayError::platform(operation, error)` constructor in `error.rs` replaces all three.
3. **Both `UnderlayPolicy` impls copy-paste the trait's contract enforcement** — the IPv6-reject + `BTreeSet` normalize preamble is verbatim-identical in `bindings/native.rs:419-431` and `bindings/unix/client.rs:215-227`, down to the error message. Contract enforcement that every impl must repeat is a shape smell; extract `normalize_underlay_targets(&[IpAddr]) -> Result<Vec<IpAddr>, GatewayError>` next to the trait in `bindings/mod.rs`.
4. **Unnamed repeated expressions in `runtime.rs`** — `self.config.tcp_buffer_bytes.min(BRIDGE_IO_CHUNK_BYTES)` appears 4×, and the flows-map lookup + `.ok_or(TcpStackError::UnknownFlow(flow))` pattern 6×. A `fn chunk_bytes(&self) -> usize` and a `fn runtime_flow_mut(&mut self, flow) -> Result<&mut RuntimeFlow, _>` would name both.
5. **Test-support duplication**: `ChannelPacketIo`, `TcpObservation`, and the hand-rolled TCP packet emitters are re-implemented three times (`runtime/tests.rs`, `gateway-bench.rs`, node `test_gateway.rs`), and `probe_bypass_http` / `ipv4_destination` / capture-loop twice (`bindings/native/tests.rs`, `tests/unix_helper.rs`). The cross-crate copies are defensible; the in-crate ones would fold into a `#[cfg(test)] pub(crate) mod test_support`.

## B. Exposure: the public API is much wider than its consumers

Every external consumer in the repo (node, the binary, examples, integration tests) imports ~26 items — runtime, config/plan/status types, connector traits, bindings. Yet `lib.rs` additionally exports the entire internal engine: `TcpStack` (its full mutation API), `FlowTable`, `GatewayServer`, `GatewayState`/`GatewayEvent`, `FlowState`/`FlowEvent`, `classify_ipv4_packet`, `PacketOutcome`/`PacketDropReason`/`FlowRejectReason`, `TcpFlowAdmission`, `TcpEndpointState`, and all transition-error types. None are used outside the crate. Two concerns: (1) for a published crate this is a semver commitment on internals; (2) `TcpStack` specifically is an API whose safe use depends on an invariant the type does not own — the endpoints↔flows synchronization lives in `GatewayRuntime`. Either make the engine types `pub(crate)`, or document explicitly that the pure models are deliberately public and state `TcpStack`'s invariant-ownership at its doc.

## C. The failed-teardown policy in `NativeGatewayRunner::run` is significant and undocumented

When `underlay.teardown(lease)` fails, the candidate-admission gate deliberately stays installed (`if cleanup_result.is_ok()` guards the clear). Because the helper control connection is by then typically dead, every subsequent `admit` errs — meaning **the node can no longer form any new WebRTC connection until restart**. I traced this and it is coherent fail-closed behavior (capture routes may genuinely still be installed, so admitting direct traffic would leak into them), but a policy with node-wide consequences is currently encoded in one unannotated conditional. It deserves (1) a comment stating the intent at both occurrences (the `activate`-failure branch has the same asymmetry), and (2) ideally a surfaced signal, since the operator will otherwise see only mysterious connect failures.

## D. FFI example wrappers: provider-creation failure discards the signer's real error

Both wrappers record the underlying signer failure but lose it on the construction path: Swift's `init` calls `signerRegistry.release(slot)` (which clears the recorded failure) before throwing generic `.providerCreation`; Kotlin never consults `signerFailure` when `new_provider_with_callback` returns NULL. The request path in both correctly surfaces the recorded cause — the creation path should too. (Verified separately that use-after-close is safe: Kotlin guards via `requireHandle()`, Swift passes NULL which `ProviderHandle::from_ptr` null-checks on the Rust side.)

## E. Verified clean in this pass

- FFI fixture (`fake_rings.c`), `test.sh`, Gradle build, and both smoke tests are internally consistent with the 65-byte signer contract and the ownership rules the README claims; the Kotlin JNA callback is retained as a provider field (the classic JNA GC pitfall is avoided).
- `BENCHMARKS.md` is honest about its measurement boundary and records the scoped-reconcile fix with numbers; the bench validates payloads and flow counts rather than just timing.
- Vendored `netlink-packet-core`: `RINGS-PATCH.md` documents provenance, `Cargo.toml.orig` confirms the only manifest delta is `paste` → `pastey`, sources contain no build script or IO/process/env access. I could not byte-compare against the crates.io archive from this environment — worth one maintainer-side `cargo vendor` diff as a final supply-chain check.
- Re-derived at final head: the `drive_client_to_onion` ordering argument (CloseWrite cannot overtake buffered client data through the capacity-1 command queue), the reaper's lazy deadline clearing, and the `reject_segment` precondition (only classified segments reach smoltcp with RX checksum verification off) all still hold.

## Follow-ups already discussed off-PR, recorded here for tracking

Test-coverage gaps that merit issues rather than blocking: (1) `NativeGatewayRunner::run` orchestration has no test at any level — making it generic over `TunnelControl` would allow a fake-control test of the admission-install/establish/activate/teardown sequencing and its failure branches; (2) no property/fuzz test asserts `classify_ipv4_packet`'s totality over arbitrary bytes, which is the actual input domain of a TUN; (3) the Established/Closing → smoltcp liveness delegation is untested for the FIN-WAIT-2-with-silent-peer case; (4) `windows/route.rs` has no unit tests — its sockaddr/LUID/prefix-mask conversions are covered only by the single Wintun CI smoke test.

RyanKung:
Re-reviewed the OS glue layer — the virtual-interface registration, the route transaction, and above all what the routing layer does **when the data plane is absent**. That question surfaces two blocker-class holes. Both are fail-open failures of the one invariant a transparent onion gateway exists to hold: *traffic that policy routes through the onion must never travel directly while the operator believes protection is active*. Everything inside the packet path fails closed; the routing boundary around it fails open.

## 1. IPv6 is fail-open: dual-stack hosts leak most traffic around the gateway

Capture is implemented exclusively as IPv4 routes — `capture_routes` emits the v4 `0.0.0.0/1` + `128.0.0.0/1` pair and v4 host routes; nothing anywhere in `bindings` touches the host's IPv6 routing state (verified: no `::/1`, no blackhole/unreachable route, no v6 handling at all). The host's IPv6 default route remains live. Plan validation *rejects* IPv6 configuration, and the README claims this is safe: "IPv6 configuration is rejected rather than leaking outside an undefined policy" — but rejecting **configuration** does nothing about **traffic**. On a dual-stack host (the common case), Happy Eyeballs prefers AAAA: connections to any dual-stack destination go out the physical interface over IPv6, in the clear, while `/gateway/status` reports Active. The gateway's `classify` drops v6 packets that reach the TUN — but v6 traffic never reaches the TUN, because no v6 route points there.

`dns_policy: block` makes this actively worse: it blocks IPv4 DNS (TCP/53 reset, UDP dropped), while IPv6 resolvers cannot even be listed (`Ipv6DnsUnsupported`) and are not blocked — so a host with RA/DHCPv6 resolvers fails over to DNS-over-IPv6, resolution succeeds, and the resulting connections go direct over v6. The strictest documented policy steers traffic onto the leak path.

**Fix**: whenever capture routes are installed, install the IPv6 counterpart as *reject/blackhole* routes (`::/1` + `8000::/1` → unreachable) through the same write-ahead ledger and lease, so v6 fails closed by default; alternatively refuse to establish on a host with a global IPv6 default route absent an explicit `allow_ipv6_leak`-style opt-in. Either way the README sentence must stop claiming that config rejection prevents leaking.

## 2. Process death is fail-open: there is no kill-switch, by construction

The interface's lifetime **is** the node process's single file descriptor. The helper creates the TUN, transfers the fd via `SCM_RIGHTS`, and retains nothing (verified — no `dup`; `into_owned_fd` moves the only copy). So on any node crash — OOM-kill, panic outside the runtime's containment, or this PR's own `gateway_task.abort()` after the 30-second cleanup timeout — the kernel destroys the interface and purges every route bound to it. All captured traffic instantly reverts to the physical default route; applications transparently reconnect in the clear; no surviving component records that protection lapsed (the helper just observes a disconnect and "cleans up" routes that are already gone).

The code demonstrably knows this mechanism — `route_is_absent` tolerates ENXIO/ENODEV/ESRCH precisely because "a route bound to a TUN/utun interface may disappear when the last packet descriptor is closed" — but only as a cleanup convenience. The entire durable-ledger / disconnect-cleanup design is oriented toward *never leaving stale capture routes behind*, i.e. availability of direct networking. For an anonymity gateway that is exactly the inverted default: leftover capture routes make the host lose connectivity (annoying, visible, recoverable); silently restored direct routing deanonymizes the user (invisible, unrecoverable). This is the same reason production VPNs pair `def1` with a kill-switch, and Tor transparent proxies use firewall redirection: bare `def1` has always had this crash behavior. The trade may be legitimate for some deployments — but right now it is not stated in the README or threat description, not configurable, and not even logged.

**Fix**: the helper should `dup()` the TUN descriptor before the `SCM_RIGHTS` transfer and hold its copy. The interface and its routes then survive node death, and captured traffic *blackholes* (fail closed: nobody reads the queue) instead of reverting. What happens next becomes an explicit, documented policy in the helper's disconnect path — hold-and-blackhole by default, with restore-direct as an operator opt-in or after a timeout — instead of an accident of kernel fd semantics. This also closes the orderly-shutdown window (`release_packet_device(device)` currently drops the interface *before* `teardown(lease)` runs, reverting traffic to direct mid-teardown) and gives the Windows path, which has no helper, a reason to document its own weaker guarantee.

## Why the (genuinely extensive) test suite could not see either hole

Every test in the PR asserts the fate of **captured** packets — that they are terminated, bridged, reset, or dropped correctly. No test asserts the fate of traffic the gateway does **not** capture, and no test asserts the routing state **after the process dies**. Two concrete additions, both fitting the existing namespace harness: (a) a namespace with a v6 default route and a dual-stack target, asserting the v6 path is unreachable while the gateway is active; (b) `kill -9` the gateway process inside the namespace and assert the captured target remains unreachable afterward. Today (a) would fail against intent and (b) would demonstrate the silent revert.

I'll flag my own part in this: earlier rounds reviewed this glue as lifecycle correctness — write-ahead ordering, reconciliation, errno tolerance — and it is well built as such. But reviewed as the anonymity boundary, the lifecycle machinery guarantees the wrong postcondition. Both holes need either a fix or a loud, deliberate threat-model statement before this ships as a "transparent gateway".

RyanKung:
Two follow-ups from a deeper pass on the OS-glue layer: a system-effect (no-shell-out) audit, which comes back clean, and a correction to the scope of the IPv6 finding above.

## System-effect audit: clean, including the dependency chain

Confirmed that no code in this PR mutates host network state by shelling out. Every OS effect goes through a programmatic API:

- Routes: `route_manager` → netlink (Linux) / `PF_ROUTE` routing socket (macOS); `windows-sys` IP Helper (`CreateIpForwardEntry2`/`DeleteIpForwardEntry2`) on Windows.
- Interface: `tun-rs` → `/dev/net/tun` `TUNSETIFF` ioctl (Linux) / utun control socket (macOS) / `libloading` + Wintun (Windows).
- Descriptor/credential/ownership: `nix` (`sendmsg`/`recvmsg` + `SCM_RIGHTS`, `getsockopt(PeerCredentials)`/`getpeereid`, `chown`).

The only `Command::spawn` in the diff is `tests/unix_helper.rs` launching the PR's own `gateway-config-unix` binary — not a system tool, and it configures nothing via CLI. Even the Linux test harness avoids shell-out: namespace via `libc::unshare(CLONE_NEWNET)` (not `ip netns`), loopback via `ioctl(SIOCSIFFLAGS)` (not `ip link set lo up`). I also checked the transitive chain, since a clean top layer can still wrap a CLI underneath: `tun-rs`'s address/MTU configuration flows through `netconfig-rs` 0.1.6, whose dependency graph is netlink-sys / system-configuration-sys / `windows` with no `std::process` or CLI-wrapping dependency — netlink/SCFramework/Win32, not `ip`. (This is from the dependency graph, not a line-by-line read of that crate; worth one confirming glance from someone who can vendor it, but nothing suggests a shell-out.) No action needed here — recording it because "no bare-command syscalls" is a hard requirement for this layer and the audit should be on the record.

## Correction to the IPv6 finding (round above): the protocol layer *does* support IPv6

I need to tighten the IPv6 leak finding, because the protocol layer is not the limitation I implied. The onion stack is IPv6-capable: `is_public_exit_ip` has a full v6 branch (`target.rs`), `OnionProxyTarget::parse_authority` handles `[2001:db8::1]:8443` with a test, and `connect_exit_target` dials resolved v6 addresses via `TcpStream::connect`. The gateway feeds only v4 targets purely because `flow.target` comes from captured v4 packets — the protocol layer is not what blocks v6.

That splits the concern in two, and only one half is optional:

- **Forwarding v6** (capturing v6 flows into the onion) is a feature. Deferring it to a later milestone is fine — and cheap when it comes, since the exit side already dials v6; the work is data-plane only (smoltcp v6 + v6 capture routes).
- **Blocking v6** (preventing v6 from egressing in the clear around the gateway) is security, and is independent of whether v6 forwarding is implemented. "We don't do v6 yet" does not justify leaving the host's v6 default route live while `/gateway/status` reads Active.

So the round-above finding stands, but its fix is not "support v6" — it's "fail v6 closed while unsupported": install `::/1` + `8000::/1` unreachable routes through the same lease/ledger whenever capture is active, or refuse to establish on a host with a global v6 default route absent an explicit leak opt-in. The README line "IPv6 configuration is rejected rather than leaking" should be corrected too — rejecting configuration does not stop traffic.

## Where things stand

Net of all rounds, the pre-merge items I'd weight highest are the two fail-open holes (v6 egress and process-death revert) from the round above — both are threat-model, not lifecycle, and each needs either a fix or a deliberate, documented, ideally logged statement of the weaker guarantee. The round-4 quality items (in-crate duplication, over-wide public surface, undocumented teardown-failure policy, FFI creation-path error loss) and the four test-coverage gaps are quality/robustness, not correctness blockers. The packet-path containment from the earlier rounds remains solid; the gap is specifically at the routing boundary around it.

RyanKung:
## Review (base `678d169d` → head `7082a149`, full diff read)

Overall impression: the model completeness and fail-closed discipline here are among the best in this repository — pure dual state machines for flow/gateway lifecycle, a write-ahead route ledger, an SCM_RIGHTS privilege boundary with ancestor-chain path validation, a relay-only ICE activation barrier, and the deliberate `std::mem::forget(device)` on cleanup failure to keep capture blackholed. Proof obligations are documented where they matter. The findings below need fixes or an explicit author response.

### Major

**1. Per-flow socket lookup is O(n), so tick-driven full reconciliation is O(n²) every 10 ms — infeasible at the configured maximum.**
`socket()` / `socket_mut()` / `release_socket` (`crates/gateway/src/tcp/mod.rs:347-398`) linearly scan the whole `SocketSet` to avoid the panicking `SocketSet::get`. Every 10 ms tick (`crates/gateway/src/runtime.rs:373`) triggers `reconcile_all`, which performs at least four such lookups per flow (`endpoint_state`, peek, write, `finish_if_closed`), i.e. ≥ 4n² steps per tick; `expired_pending_flows` is O(n²) as well. At the default n = 1024 the benchmark already shows the cost; at `MAX_GATEWAY_FLOWS = 16_384` (`crates/gateway/src/config.rs:16`) the timer path alone is on the order of 10⁹ steps per tick. BENCHMARKS.md records the packet-path quadratic fix, but each step of the retained timer-path "full scan" still hides an inner linear scan. The fix does not require reintroducing a panic path: maintain the invariant "handle ∈ `endpoints` ⇒ handle valid in `SocketSet`" (they are already inserted and removed together), prove membership in O(1) via an owned `HashSet<SocketHandle>` before calling `SocketSet::get`, or bypass the iterator-based lookup entirely.

**2. `underlay_refresh_secs` does not do what its name promises: the refresh loop only polls exit availability and never refreshes the underlay bypass.**
`refresh_exit_availability` (`crates/node/src/native/gateway.rs:430`) only performs `lookup_onion_exits` + `set_exit_availability`. TURN hostnames are resolved once at startup and installed as bypass routes; if webrtc's own later DNS resolution returns a rotated IP (low TTL / multiple A records), relay traffic falls into the capture route and is dropped — the gateway degrades with no self-healing. The helper protocol has `ReplaceBypass` and the client implements `replace_bypass_targets`, but nothing calls them after establish. Either make this loop actually re-resolve TURN and invoke `ReplaceBypass` (the machinery exists), or rename the config field and `GatewayRefresh` and document that TURN IP drift is outside the guarantee. Right now the name promises behavior that does not happen.

**3. The `ConnectPeerViaHttp` pinning / IPv4-only / no-redirect policy applies unconditionally to all native nodes, not just gateway mode.**
Starting at `crates/node/src/rpc_impl.rs:45`, native builds always go through `admit_http_signaling_underlay` + `pinned_signaling_http_client`. The old path used `Client::new` with reqwest defaults (follows redirects, allows IPv6). Now, even with no gateway configured, an IPv6-only signaling endpoint fails with "the native gateway milestone is IPv4-only" and any HTTP redirect fails. `admit_underlay_targets` is a no-op without an installed policy, but the resolution constraints are not. This is a behavioral regression beyond gateway scope that the PR description does not declare. Suggestion: enable the strict path only when an admission policy is installed, and keep the original client otherwise; if the tightening is intentional and universal, state it explicitly and fix the error wording.

**4. IPv6 underlay bypass targets from the config are silently dropped.**
`GatewayUnderlayGate::new` (`crates/node/src/native/gateway.rs:72`) applies `filter(IpAddr::is_ipv4)` to `underlay_bypass_targets`. Everywhere else the gateway crate rejects IPv6 explicitly (`Ipv6RouteUnsupported` / `Ipv6DnsUnsupported`); only here it swallows silently — the operator believes a bypass exists while traffic to that target is captured and dropped. Validate and reject in `NativeGatewayRunner::new`, consistent with the rest of the crate.

**5. `TunnelControl::teardown` consumes the lease even on failure, so the helper can never retry.**
The trait takes `lease` by value (`crates/gateway/src/bindings/mod.rs:58`); `NativeTunnelControl::teardown` restores its internal `active` on error, but the lease value is already dropped. On the helper side (`crates/gateway/src/bindings/unix/helper.rs:184`) `lease.take()` followed by a failure enters the "cannot be retried" dead end: subsequent teardown and establish are both refused, and the only recovery is killing the helper process and relying on stale reconciliation. Fail-closed is right, but "one transient route-delete failure ⇒ mandatory privileged-process restart" is too heavy. A linear capability should be returned on failure: `teardown(lease) -> Result<(), (Lease, GatewayError)>` or equivalent, with the helper putting the lease back.

### Minor

**6. Dead / test-only API on the public surface.** `TcpStack::client_write_open` (`crates/gateway/src/tcp/mod.rs:294`) has zero callers repository-wide; `read_application_data` (`:220`) is used only by in-crate tests while production uses peek/commit. Additionally, `lib.rs` publicly re-exports the entire internal model (`FlowTable`, `TcpStack`, `GatewayServer`, `classify_ipv4_packet`, …) while the node consumes only the runtime + bindings + status + stream vocabulary. If this crate is headed for crates.io, all of that is long-term semver surface; please confirm whether exporting the model is deliberate or the surface should be narrowed.

**7. An unreachable arm in `drive_client_to_onion` would lose the buffer.** `Err(TrySendError::Full(BridgeCommand::CloseWrite))` (`crates/gateway/src/runtime.rs:607`) is unreachable in a match that sent `Data`, yet it is written as `return Ok(())` without restoring `client_to_onion_buffer`. Match `Err(Full(command))` and destructure afterwards so unreachability is guaranteed by structure rather than by reader inference.

**8. `FlowTable::new(0)` returns `CapacityExhausted { limit: 0 }`.** (`crates/gateway/src/flow_table.rs:21`) "Capacity 0 is exhausted" is imprecise; zero capacity is an illegal construction and deserves its own variant (or the `ZeroMaxFlows` semantics from config).

**9. The same inline python3 test-binary extraction script is copy-pasted five times in CI** (three Linux occurrences plus two in the macOS block of `qaci.yml`). Repetition should have one source: a small script under `scripts/` invoked by each job.

**10. The helper reads requests with no timeout.** The single-threaded helper blocks in `read_request` (`crates/gateway/src/bindings/unix/helper.rs:144`); an authenticated but stalled (not disconnected) node connection wedges the helper forever, including the disconnect-recovery path. Only the same UID can connect, so it is self-inflicted, but one `set_read_timeout` plus disconnect classification eliminates it.

**11. `replace_active_bypass` inherits routes from the establish-time baseline.** (`crates/gateway/src/bindings/native.rs:323`) If the host's physical default route migrates while the tunnel is up (Wi-Fi → wired), newly installed bypasses point at a defunct gateway. Using the baseline instead of the live table is justified (it avoids the self-installed /1 capture routes), but this known limitation belongs in the README's operational assumptions.

**12. Switching `send_outbox` from `scope.send` to `scope.send_direct` changes delivery semantics of existing onion links** (Chord routing → exact physical edge) and is not mentioned in the PR description. It is plausibly the correct fix for the cover-traffic adjacency assumption, but: (a) please confirm it is required by this PR rather than bundled; (b) the module header in `send_outbox.rs` still says "overlay sends" / "overlay backpressure" — the wording should track the behavior change.

The test coverage is serious work (deterministic total-domain fuzz of the classifier, out-of-order/retransmit exactly-once, late-event idempotence, privileged Linux/macOS/Windows real-TUN + route-cleanup + disconnect-retention in CI). Items 1–3 are the ones I consider must-address before merge; the rest per your judgment.

RyanKung:
## Review round 2 (`7082a149` → `51bcdf06`, 4 new commits, full re-read of the resulting state)

### The new model boundary: sound, and considerably cleaner than the old one

The old model was "capture the default route + relay-only ICE + mandatory TURN + dynamic underlay admission" — an entire machinery whose only purpose was to keep the node's own WebRTC traffic from looping into the tunnel. The new commits remove the condition that created that problem in the first place:

- **Capture = explicit split routes, nothing else.** `GatewayPlan` drops `routing_mode` / `excluded_routes` / `dns_policy` / `dns_servers`, and `deny_unknown_fields` rejects the removed fields (with a regression test proving old full-capture configuration cannot be silently reinterpreted). Prefix-0 routes are rejected, `captures_entire_ipv4_space` rejects equivalent full-tunnel prefix unions via interval math, and interface addresses are forced to /32 so setup cannot create an implicit connected capture prefix.
- **Loop protection degrades from "dynamic mechanism" to "trivial fact":** the default route is untouched, underlay/ICE/signaling keep ordinary routing, and TURN becomes genuinely optional. The entire transport `UnderlayCandidateAdmission` surface, the three swarm methods, and the pinned HTTP client in `rpc_impl` are deleted — which resolves round-1 item 3 (the unconditional IPv6/redirect regression for non-gateway native nodes) by removing the mechanism as a whole; `ConnectPeerViaHttp` is back to its original behavior.
- The "operator captures their own bootstrap/ICE destination" case is answered honestly in the README: such a route behaves like any other operator route, fails closed (captured UDP dropped), and Rings does not silently punch a bypass. This fixes the problem at its boundary rather than patching symptoms.

### Round-1 findings: all 12 addressed on the merits, none papered over

1. **O(n²) reconciliation** → `owned_handles: HashSet<SocketHandle>` inserted/removed together with `endpoints` and the `SocketSet`; an O(1) membership proof now guards smoltcp's O(1) typed accessor, with the invariant documented. 2. **`underlay_refresh_secs` misnomer** → renamed `status_refresh_secs` with bounded validation (1..=30); the TURN-drift hazard disappeared with the model. 3. → resolved by deletion, see above. 4. **silent IPv6 bypass drop** → the bypass concept no longer exists. 5. **teardown consuming the lease** → new `TeardownFailure<L>` returns the linear capability on failure; the helper re-arms `tunnel.lease` and keeps serving the connection so the client can retry without restarting either side of the privilege boundary; both platform implementations updated. 6. **dead / test-only API** → `client_write_open` deleted, `read_application_data` demoted to `#[cfg(test)]`; the remaining public model surface is now an explicit embedding contract ("Library model API" in the README). 7. **unreachable arm losing the buffer** → restructured to destructure `Full(command)`; the CloseWrite case is now a typed invariant-violation error. 8. → `ZeroCapacity` variant plus test. 9. **5× copy-pasted CI extraction** → `scripts/find-cargo-test-binary.py`, with a uniqueness assertion added. 10. **helper read stall** → 60 s write timeout; the absence of a read deadline is now a deliberate, test-pinned decision (an idle lease connection is legitimate). 12. **`send_direct` doc drift** → module docs reworded to direct-edge and an adjacency/cover-shape rationale comment added at the call site. The Windows binding also shrank to exactly add/delete plus destination-only list; `GetBestRoute2` and friends are gone.

### New findings this round

**A (medium — the one substantive gap): a more-specific pre-existing route silently shadows capture.** `ensure_capture_destination_available` (`crates/gateway/src/bindings/native.rs:273`) rejects only an **exact** destination+prefix collision. If the host already has (or later gains) a route strictly **inside** a captured prefix — capture `1.0.0.0/24` while the table holds `1.0.0.128/25 via eth0` — longest-prefix match wins and traffic to that sub-range goes direct, with no warning. The README's contract line "installed destination capture routes = C" will lead operators to believe destinations in C are captured; "fail-closed applies only after traffic enters the gateway" covers this in wording, not in mechanism. `list()` is already in hand at establish time: change the equality check to a containment check (refuse, or at least surface shadowed sub-ranges in status/logs), and document in the README that longest-prefix shadowing — including post-establish route changes, since there is deliberately no route-change subscription — is an operational hazard. Related nit: `list()` is currently fetched once per capture route; hoisting it out of the loop also gives the containment check one consistent table snapshot.

**B (wording-level): `FullTunnelUnsupported` is an intent stop-line, not a soundness invariant.** A route set covering the entire IPv4 space minus a single /32 passes validation while carrying exactly the recursion hazard the README cites as the reason for rejecting full tunnels. That is fine — the real soundness comes from "self-capture fails closed and never leaks" — but the README currently attributes the rejection to recursion avoidance. One sentence acknowledging that near-cover sets are allowed and rely on the same fail-closed property would make the argument exact.

**C (minor): the timeout-classification machinery in `transport.rs` is production-dead.** The helper never sets a read timeout (a test asserts `read_timeout == None`), so `frame_read_error`'s TimedOut/WouldBlock branch can only fire in the test that manufactures the condition, and `is_request_timeout` is itself `#[cfg(test)]`. This reads like a leftover from an intermediate design: either wire a real read-deadline policy, or drop the classification so the code contains only paths with a production trigger.

Item A is worth addressing (or at least explicitly documenting the shadowing hazard) before merge; B and C at your discretion. The response quality this round is high — rather than patching each flagged line, the commits removed the root cause (route authority ownership) as a whole.

RyanKung:
## Review round 4 (head `feb99bd8`; the only new commit is the 0.20.0 version bump)

This round re-reads the full PR state under two dedicated lenses: (1) boundary / blast radius — can this disrupt host networking or crash, and (2) cross-platform component reuse. It includes source-level verification of `route_manager 0.2.13`'s delete semantics on each platform, since those decide whether stale cleanup can ever remove someone else's routes.

### 1. Boundary: can it wreck host networking?

Positive confirmations first, then two real gaps.

**The blast radius genuinely stays inside C (the explicit capture set).** A grep over the whole diff shows no sysctl / `ip_forward` / DNS / resolv / firewall / interface-flag mutation anywhere; no default route, no `/1` halves, no IPv6 routes are ever installed; interface addresses are forced to `/32` so no implicit connected prefix can appear; smoltcp's internal AnyIP default route is a purely userspace reply path (now documented as such). The worst cases — SIGKILL'd node, failed cleanup, helper disconnect — blackhole only prefixes inside C and leave all other traffic untouched. That is not just a claim: the privileged CI tests on all three platforms actually verify that the unselected `1.1.1.1` stays directly reachable, `1.0.0.1` arrives at the packet device, teardown removes the ledger, and stale recovery works. On the crash side: workspace lints deny all panic classes; the one panicky API in use (`SocketSet::get`) is guarded by an O(1) membership proof; the write-ahead ledger reduces every crash window to a harmless cleanup intent; all queues and buffers are bounded (flow table, buffer budget, bridge event channel, ack backpressure), and the loop admits one packet per select iteration.

**F1 (medium, all platforms): no mutual exclusion on the ledger — a second instance silently tears down live capture routes, i.e. fail-open.** `reconcile_stale` runs unconditionally inside every establish (`crates/gateway/src/bindings/native.rs:160,191`) and deletes whatever the ledger names, but the ledger has no lock and no owner/liveness marker. A second helper or node instance pointed at the same ledger will delete the **first instance's live routes** during its own establish. I verified the delete keys: Linux netlink includes Oif (instance A's interface is alive, so the delete succeeds), Windows `DeleteIpForwardEntry2` keys on LUID (also succeeds), macOS keys on dst+netmask. The consequence: prefixes A had captured silently revert to direct routing or migrate into B's tunnel — previously fail-closed prefixes silently fail open, which is exactly the network-disruption class this review targets. Windows is the easiest to hit: the default `route_ledger_path` is a fixed per-user path, so two `rings run --gateway` invocations collide, with no Unix-style socket contention in the way (and on Unix, `remove_stale_socket` does not probe liveness either — it will unlink a *running* helper's socket). Suggestion: hold an exclusive advisory lock (flock / `LockFileEx`) on the ledger for the lifetime of `NativeTunnelControl`, require the lock for reconcile, and probe the socket with a connect before removing it as stale.

**F2 (medium, macOS-specific): stale-ledger reconciliation can bypass the `CaptureRouteShadowed` guarantee and delete a third party's route.** macOS `RTM_DELETE` carries only dst+netmask (capture routes have no gateway, `if_scope` is unset, and `RTA_IFP` exists only on the NetBSD branch). So: a crash leaves a ledger naming prefix P; a third party (another VPN, an administrator) later installs a route for P; a same-plan re-establish first deletes that **foreign** route via reconcile, after which the shadow check sees no conflict and capture proceeds — while the identical host state *without* a stale ledger is correctly refused with `CaptureRouteShadowed`. The same host state producing opposite outcomes depending on crash residue is the boundary hole. Linux (stale oif → ENODEV, treated absent) and Windows (stale LUID → NOT_FOUND) are naturally immune. Suggestion: scope macOS deletes with RTF_IFSCOPE plus the interface, or have reconcile `list()` first and delete only entries whose current route matches the recorded interface (skip and report otherwise).

**F3 (minor):** the run loop's `tokio::time::interval` keeps the default `MissedTickBehavior::Burst` — after a data-plane stall (e.g. TUN write backpressure), the backlog of 10 ms ticks fires back-to-back, each triggering a full-table reconcile, producing a CPU spike. `Delay` or `Skip` fits a data plane better.

### 2. Cross-platform reuse

The current shape is close to what it should be — this is not three parallel implementations. The **data plane is 100% platform-neutral** (runtime / tcp / bridge / flow / server / status / packet contain no platform branch at all); `NativeTunnelControl` + ledger + write-ahead route transaction + shadow check are **single-source** across all three platforms (`native.rs`); device creation goes uniformly through tun-rs. What remains genuinely per-platform: ~15 lines of `configure()` each, Windows' own minimal route binding (already shrunk to add/delete/destination-list in round 3), and the Unix privilege helper, which is deliberate (Windows is in-process, with the rationale documented). The privileged test support module and the CI extraction script are shared as well.

Two remaining consolidation candidates (observations, not defects):

1. **The Route/RouteManager cross-platform contract is duck-typed via cfg re-export** (`crates/gateway/src/bindings/route.rs`) rather than a trait — the Windows binding aligns with `route_manager` by naming convention, so method drift is only caught by each platform's separate compile. A small `RouteTable` trait would make the contract compiler-checked, and would also let the `RouteRecord::from_route` / `into_route` conversions (12 fields × two directions of per-OS cfg — the heaviest remaining platform-branch block in `native.rs`) sink into each platform's implementation.
2. Four shape-similar "aggregate primary + cleanup failure" helpers live at four layers: `record_first_error` / `with_cleanup_error` in `runtime.rs`, `combine_results` in `helper.rs`, `combine_gateway_results` in `native/gateway.rs`, `combine_foreground_results` in `rings.rs`. Their shapes differ slightly (2-way vs 3-way, `GatewayError` vs `anyhow`), so cross-crate unification has moderate payoff.

F1 and F2 are the items I consider worth addressing before merge — the lock is a few dozen lines, and the macOS interface check on reconcile is small; F3 and the reuse notes are at your discretion.

PR #709, opened 2026-09-02
Harden secret key and signature handling

### LINKED ISSUE #705: Key hygiene: redact Debug for Ed25519SecretKey/SigningSecretKey and zeroize long-lived secret keys
## Summary

Two account-signing secret key types derive `Debug` and print raw key bytes, and long-lived secret keys are never zeroized. This is inconsistent with the deliberate redaction already applied to `SecretKey`.

## Details

`SecretKey`'s `Debug` is correctly redacted (`crates/core/src/ecc/mod.rs:60`):

```rust
impl std::fmt::Debug for SecretKey {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_tuple("SecretKey").field(&"<redacted>").finish()
    }
}
```

But `Ed25519SecretKey([u8; 32])` and `SigningSecretKey` both `derive(Debug)` (`crates/core/src/ecc/keys.rs:105` and `:109`), so any `tracing` / `format!("{:?}")` / panic message touching one prints the raw 32-byte seed. `SigningSecretKey` also derives `Serialize`.

Separately, long-lived secrets are not zeroized: `SecretKey([u8; 32])` (`crates/core/src/ecc/mod.rs:57`) is `Copy` with no `Zeroize`/`Drop`, and is held inside `SessionSk`, `SigningSecretKey`, `Ed25519SecretKey`. The AEAD path already uses `Zeroizing` for ephemeral DEM key material (`crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs:407`), so the pattern is known — it just isn't applied to the higher-value persistent keys. Residual bytes linger in freed memory / swap / wasm linear memory.

## Proposed fix

- Hand-write redacting `Debug` for `Ed25519SecretKey` and `SigningSecretKey` (mirroring `SecretKey`).
- Apply `Zeroize` (on drop) to the long-lived secret key types, reconciling with the current `Copy` derive (likely dropping `Copy` for these).

### LINKED ISSUE #706: Signature malleability: enforce low-s (secp256k1/r1), verify_strict (ed25519); guard eip191 recovery-id arithmetic
## Summary

Signature verification does not enforce canonical/low-s form, and the eip191 recovery-id arithmetic is unguarded. Impact is limited today because signatures are not used as identifiers/dedup keys — but if a replay/dedup cache keyed on the signature is added (a natural fix for message replay), malleability becomes a bypass. Worth fixing proactively.

## Details

- **secp256k1** message verify recovers a pubkey and compares addresses (`crates/core/src/ecc/signers/secp256k1/mod.rs:31`); given `(r, s, v)` an attacker can produce `(r, n-s, v^1)` that recovers the same key — a second valid signature over the same message. No low-s check.
- **secp256r1** verify uses `verify_prehash` with `Signature::from_slice` and does not reject high-s (`crates/core/src/ecc/signers/secp256r1/mod.rs`).
- **ed25519** verify uses the non-strict `verify` (accepts non-canonical S / mixed-order points) rather than `verify_strict` (`crates/core/src/ecc/signers/ed25519/mod.rs:41`).
- **eip191 recovery-id arithmetic is unguarded.** `recover` does `sig712[64] -= 27` on the attacker-controlled byte with no range check (`crates/core/src/ecc/signers/eip191/mod.rs:34`). A crafted `v < 27` underflows: debug builds panic (subtract-overflow), release wraps (then fails `RecoveryId::from_byte`, so release is safe). Compare `bip137::recover`, which correctly range-checks `27..=34` before subtracting.

## Proposed fix

- Enforce low-s on secp256k1 and secp256r1 verification.
- Use `verify_strict` for ed25519.
- Range-check the eip191 recovery id before subtracting `27`, matching the bip137 guard.

### LINKED ISSUE #707: Replace silent crypto degradation (sign_hash -> [0u8;65], secp256k1_scalar -> ONE) with explicit errors
## Summary

Two cryptographic primitives replace an unsatisfiable invariant with a silent wrong value to satisfy the panic-free lints, instead of propagating an error or asserting the invariant. Both are "should never happen" given construction-time validation, but a silent wrong value is the worst failure mode if one ever does.

## Details

1. **`sign_hash` returns an all-zero signature on failure** (`crates/core/src/ecc/mod.rs:339`):

   ```rust
   let signing_key = match K256SigningKey::from_slice(&self.0) {
       Ok(signing_key) => signing_key,
       Err(_) => return [0u8; 65],
   };
   let (signature, recover_id) = match signing_key.sign_prehash_recoverable(message_hash) {
       Ok(signature) => signature,
       Err(_) => return [0u8; 65],
   };
   ```

   A signing failure yields a silent `[0u8; 65]` that fails verification downstream — fail-closed, but it converts a broken invariant into a hard-to-diagnose downstream failure rather than an error at the source.

2. **`secp256k1_scalar()` falls back to `K256Scalar::ONE`** (`crates/core/src/ecc/mod.rs:311`):

   ```rust
   Option::<K256Scalar>::from(K256Scalar::from_repr(self.0.into())).unwrap_or(K256Scalar::ONE)
   ```

   Falling back to a *known* scalar (1) is more dangerous than the zero-sig case if this scalar ever feeds ElGamal/DH: it would be a predictable secret. It "cannot" happen because `SecretKey::from_bytes` validates at construction, which is exactly why the branch should be `unreachable!` with the proof stated, or a propagated `Result`, not a silent constant.

## Proposed fix

Change both to return a `Result` (propagating the underlying error), or — where the invariant is genuinely guaranteed by construction — an `unreachable!`/`expect` with the invariant documented, so a real failure surfaces at the source instead of degrading silently. Prefer propagation for `sign_hash`; for `secp256k1_scalar`, never fall back to a usable scalar value.

RyanKung:
## Summary

- redact `Debug` output and zeroize long-lived secp256k1/Ed25519 signing secret storage on drop; remove implicit `Copy` from secret-bearing types
- store secp256k1 `SecretKey` as a construction-time-validated k256 secret, return zeroizing serialized key copies, and remove silent fallback signatures/scalars
- reject high-s secp256r1 signatures, canonicalize WebCrypto P-256 signatures before session authorization, use strict Ed25519 verification, validate EIP-191/BIP-137 recovery IDs through one helper, and keep the secp256k1 high-s check as an explicit error-mapping invariant

## Compatibility

This intentionally changes the public signing surface: `SecretKey::{sign, sign_raw, sign_hash}` and the secp256k1/EIP-191 signing helpers now return `Result`, signing helpers borrow secret keys, `SecretKey::ser()` returns `Zeroizing<[u8; 32]>`, by-value secret conversion impls were removed, and secret key types are no longer `Copy`. Secp256r1 high-s verification and strict Ed25519 verification are called out in the changelog as breaking validation changes. Browser WebCrypto session signatures are normalized to low-S before verification so freshly created frontend P-256 accounts remain accepted under the stricter verifier.

## Validation

- `cargo +nightly-2026-07-02 fmt --all -- --check`
- `cargo check -p rings-node --features browser_default --no-default-features --target=wasm32-unknown-unknown`
- `cargo check --target=wasm32-unknown-unknown` in `frontend/`
- `cargo clippy --target=wasm32-unknown-unknown --all-targets -- -D warnings` in `frontend/`
- `cargo test -p rings-core --features dummy --quiet` (655 passed, 4 ignored; doc tests 1 passed, 2 ignored)
- `cargo test -p rings-core --features dummy ecc::signers::secp256r1 --quiet`
- `npm run test:frontend-extension-webview`
- `git diff --check`

Closes #705
Closes #706
Closes #707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Secret signing keys are no longer `Copy`; signing helpers now borrow keys and return explicit
   errors instead of silently substituting invalid signatures or scalars.
+- Secret key containers now zeroize their long-lived scalar or seed storage on drop.
+- Secp256r1 verification now rejects high-s signatures, so persisted high-s secp256r1 session
+  proofs from older builds no longer verify. Ed25519 verification now uses strict signature
+  validation.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 0.20.0
 
+### Breaking changes
+
+- Secret signing keys are no longer `Copy`; signing helpers now borrow keys and return explicit
+  errors instead of silently substituting invalid signatures or scalars.
+
 ### Added
 
 - Add the opt-in native IPv4/TCP gateway for explicitly selected destination prefixes, with
--- a/crates/core/src/dht/mod.rs
+++ b/crates/core/src/dht/mod.rs
@@ -60,11 +60,11 @@ pub use virtual_node::MAX_STORAGE_VIRTUAL_POSITIONS_PER_OWNER;
 pub mod tests {
     //! test
     use super::*;
-    use crate::ecc::tests::gen_ordered_keys;
+    use crate::ecc::tests::gen_ordered_key_vec;
 
     /// Test get ordered did list
     pub fn gen_ordered_dids(n: usize) -> Vec<Did> {
-        gen_ordered_keys(n)
+        gen_ordered_key_vec(n)
             .iter()
             .map(|x| x.address().into())
             .collect()
--- a/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs
+++ b/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs
@@ -377,14 +377,14 @@ pub fn encrypt_bytes_with_rng(
 }
 
 /// Decrypt ciphertext produced by the current secp256k1 compatibility adapter.
-pub fn decrypt(m: &[CiphertextBlock], k: SecretKey) -> Result<String> {
+pub fn decrypt(m: &[CiphertextBlock], k: &SecretKey) -> Result<String> {
     String::from_utf8(decrypt_bytes(m, k)?).map_err(Error::Utf8Encoding)
 }
 
 /// Decrypt arbitrary bytes produced by [`encrypt_bytes_with_rng`].
-pub fn decrypt_bytes(m: &[CiphertextBlock], k: SecretKey) -> Result<Vec<u8>> {
+pub fn decrypt_bytes(m: &[CiphertextBlock], k: &SecretKey) -> Result<Vec<u8>> {
     let secret_key =
-        ElGamalSecretKey::<Point<Secp256k1>>::from_scalar(GroupScalar::<Secp256k1>::from(k));
+        ElGamalSecretKey::<Point<Secp256k1>>::from_scalar(GroupScalar::<Secp256k1>::try_from(k)?);
     let ciphertext = m
         .iter()
         .map(|(c1, c2)| Ok(((*c1).try_into()?, (*c2).try_into()?)))
@@ -431,7 +431,7 @@ pub fn encrypt_aead_with_rng(
 pub fn decrypt_aead(
     sealed: &AeadCiphertext,
     aad: &[u8],
-    recipient_secret: SecretKey,
+    recipient_secret: &SecretKey,
 ) -> Result<Vec<u8>> {
     if sealed.version != AEAD_VERSION {
         return Err(Error::MessageDecryptionFailed(format!(
--- a/crates/core/src/ecc/elgamal/impls/secp256k1/test_secp256k1.rs
+++ b/crates/core/src/ecc/elgamal/impls/secp256k1/test_secp256k1.rs
@@ -117,7 +117,7 @@ fn test_algorithm() {
         31, 146, 117, 219, 175, 223, 186, 129, 148, 46, 179, 51, 11, 7, 243, 140, 190, 228, 235,
         184, 107, 220, 33, 116, 175, 150, 72, 213, 245, 80, 154, 84,
     ];
-    let r_sca = GroupScalar::<Secp256k1>::from(r).into_inner();
+    let r_sca = GroupScalar::<Secp256k1>::try_from(&r).unwrap().into_inner();
     let mut got_r = [0u8; 32];
     got_r.copy_from_slice(r_sca.to_bytes().as_slice());
     assert_eq!(got_r, r_v);
@@ -164,7 +164,10 @@ fn test_algorithm() {
     assert_eq!(got_c2_x, c2_x);
     assert_eq!(got_c2_y, c2_y);
 
-    let t = K256ProjectivePoint::from(a_c1) * GroupScalar::<Secp256k1>::from(key).into_inner();
+    let t = K256ProjectivePoint::from(a_c1)
+        * GroupScalar::<Secp256k1>::try_from(&key)
+            .unwrap()
+            .into_inner();
     let a_t = t.to_affine();
     let t_x = [
         218, 19, 55, 137, 15, 46, 160, 160, 208, 222, 206, 77, 46, 79, 32, 80, 64, 243, 93, 23,
@@ -189,7 +192,7 @@ fn test_encrypt_decrypt() {
             .unwrap();
     let pubkey = key.pubkey();
     let t: String = random(1024);
-    assert_eq!(decrypt(&encrypt(&t, pubkey).unwrap(), key).unwrap(), t)
+    assert_eq!(decrypt(&encrypt(&t, pubkey).unwrap(), &key).unwrap(), t)
 }
 
 #[test]
@@ -200,7 +203,7 @@ fn test_encrypt_decrypt_keeps_nul_bytes() {
     let pubkey = key.pubkey();
     let message = format!("\0{}{}", "a".repeat(FIELD_CHUNK_SIZE - 1), "\0tail");
     assert_eq!(
-        decrypt(&encrypt(&message, pubkey).unwrap(), key).unwrap(),
+        decrypt(&encrypt(&message, pubkey).unwrap(), &key).unwrap(),
         message
     );
 }
@@ -217,7 +220,7 @@ fn test_encrypt_decrypt_binary_bytes() {
 
     let ciphertext = encrypt_bytes_with_rng(&payload, pubkey, &mut rng).unwrap();
 
-    assert_eq!(decrypt_bytes(&ciphertext, key).unwrap(), payload);
+    assert_eq!(decrypt_bytes(&ciphertext, &key).unwrap(), payload);
 }
 
 #[test]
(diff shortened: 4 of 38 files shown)


RyanKung:
Review of `0f1c0ee` (full diff read, logic only; no local gates run — CI is green so far with the remaining jobs pending).

The direction is right: secp256r1 low-s normalization + rejection, `verify_strict` for Ed25519, the EIP-191 recovery-id guard, `Result` propagation, dropping `Copy`, and the redacted `Debug` impls are all what #705–#707 asked for. The points below are ordered by importance.

### 1. The secp256k1 `NonCanonicalSignature` check duplicates what k256 already enforces

I checked the vendored sources: k256 0.13.4's `VerifyPrimitive::verify_prehashed` (`k256/src/ecdsa.rs:203`) already returns an error when `sig.s().is_high()`, and ecdsa 0.16.9's `recover_from_prehash` ends with `vk.verify_prehash(prehash, signature)?`. So high-s secp256k1 signatures were already rejected before this PR, just as `Error::ECDSAError`; `recover_hash_rejects_high_s_signature` would pass on `master` with a different error variant. (That also means the secp256k1 claim in #706 was wrong — my mistake when filing it. The secp256r1 half is real: p256 0.13.2 uses the default `SignPrimitive`/`VerifyPrimitive`, which neither normalizes on sign nor rejects on verify, so roughly half of the secp256r1 signatures this crate produced before were high-s.)

Suggestion: either drop the check in `recover_hash` (trust the library) or keep it as an explicitly documented invariant that exists only for the error mapping, with a comment saying k256 enforces the same thing. Separately, all three sites (`recover_hash`, `secp256r1::sign`, `secp256r1::verify`) spell the "is low-s" predicate as `normalize_s().is_some()/is_none()`, which constructs a whole new signature to test one bit. `signature.s().is_high()` is the direct predicate (it is what k256 uses); it should be one named predicate rather than three inline domain checks.

### 2. `SecretKey::pubkey()` still silently substitutes `PublicKey([0u8; 33])`

This is exactly the pattern #707 targets, but the PR only changed `sign_hash` and `secp256k1_scalar`, so the same "impossible" state now has two different treatments. The root cause is that `SecretKey` stores raw bytes and re-parses them on every use. Storing the construction-time-validated `K256SecretKey` (or `NonZeroScalar`) instead makes the invalid state unrepresentable, and the `PrivateKeyBadFormat` branch in `sign_hash` plus `invalid_secret_key_state_returns_explicit_errors` (which constructs a value the type's own invariant forbids) go away. That is a larger refactor and optional here, but the three paths should at least agree.

### 3. Zeroization is partial

`Drop` clears the container only. The `[u8; 32]` returned by `ser()`, the `FieldBytes` copies, and the bare `K256Scalar` copied out of the (self-zeroizing) `K256SecretKey` in `secp256k1_scalar()` — which then lives long-term inside `GroupScalar`/`ElGamalSecretKey` — are not zeroized. Not a bug, but the changelog wording ("zeroize secrets") overstates coverage. `ser()` returning `Zeroizing<[u8; 32]>` would close the most obvious gap.

### 4. CHANGELOG is missing the verification-side tightening

secp256r1 verification now rejects high-s, and previous versions of this crate produced high-s secp256r1 signatures about half the time, so any persisted secp256r1 session proof from an older build stops verifying. Ed25519 moving to `verify_strict` is also an interop change. Both belong under Breaking changes.

### 5. Dead / speculative conversion impls

After this PR the by-value `TryFrom<SecretKey> for Scalar<Secp256k1>`, `From<SecretKey> for FieldBytes<NistP256>`, and `TryFrom<SecretKey> for Fr` have no callers. `TryFrom<&SecretKey> for Fr` (`signers/bls/mod.rs:67`) duplicates the body instead of delegating, unlike the `Scalar<Secp256k1>` pair. The new `From<&SecretKey> for PublicKey<33>` has zero callers. Since the PR already declares a breaking change, keep only the borrowed forms.

### 6. `gen_ordered_keys(n).try_into().unwrap()` + array pattern repeated ~20 times across 5 test files

Make `gen_ordered_keys` const-generic (`fn gen_ordered_keys<const N: usize>() -> [SecretKey; N]`) so call sites become `let [key1, key2, key3] = gen_ordered_keys();`.

### 7. EIP-191 and BIP-137 each hand-write the same `v → recovery id` mapping

`eip191::recover` now uses `(27..=30).contains`, `bip137::recover` uses two hand-written ranges. Extract one shared `recovery_id_from_v(v, base)` (or at least use the same form in both).

### 8. Tests

- `secret_key_redacts_debug_and_zeroizes_on_drop` / `signing_secrets_redact_debug_and_zeroize_on_drop` only assert `needs_drop`; they do not test zeroization, so the names overclaim.
- `test_secp256r1_rejects_high_s_signature` reaches the public key via `super::super::super::keys::SigningSecretKey` plus `let … else unreachable!`. `crate::ecc::SigningSecretKey` is already re-exported, or derive the verifying key directly with `p256::ecdsa::SigningKey`.

### 9. Minor

`*secret.to_nonzero_scalar().as_ref()` in `secp256k1_scalar`: `NonZeroScalar` derefs to `Scalar`, so `*secret.to_nonzero_scalar()` suffices (and the whole function disappears if item 2 is taken).

RyanKung:
Addressed the review comments in 96413a36.

Key changes:
- `SecretKey` now wraps the validated k256 secret type, so invalid secp256k1 key state is no longer representable through the wrapper.
- `SecretKey::ser()` now returns `Zeroizing<[u8; 32]>`; changelog wording now scopes zeroization to long-lived secret storage.
- Low-s checks use a named `ecdsa_signature_s_is_high` predicate; secp256k1 recovery keeps the explicit `NonCanonicalSignature` mapping with a comment noting k256 also enforces it.
- Removed dead by-value secret conversion impls and the unused `From<&SecretKey> for PublicKey<33>` conversion.
- `gen_ordered_keys` is const-generic for fixed-size tests; runtime-size tests use `gen_ordered_key_vec`.
- EIP-191 and BIP-137 recovery-id normalization now share `recovery_id_from_v`.
- Renamed overclaiming `zeroize_on_drop` tests to `needs_drop`, and removed the secp256r1 test's `let else unreachable!` path.

Local validation on the pushed head:
- `cargo +nightly-2026-07-02 fmt --all -- --check`
- `cargo test -p rings-core --features dummy --quiet` (654 passed, 4 ignored; doc tests 1 passed, 2 ignored)

RyanKung:
Follow-up pushed in `faa7d0ea` for the Extension WebView CI failure after the review-fix commit.

Root cause: after secp256r1 verification was tightened to reject high-S signatures, browser WebCrypto could still produce either S form for the P-256 account signature used by `SessionSkBuilder`. That made the offscreen extension node intermittently reject its freshly generated session proof with `Failed on verify message signature`.

Changes in this follow-up:

- added `secp256r1::normalize_signature` in core for raw 64-byte P-256 ECDSA signatures
- kept core verification strict: high-S signatures still do not verify
- normalized WebCrypto signatures before passing them into session authorization
- added regression coverage for high-S normalization in both core and frontend wallet code

Validation on current head:

- `cargo +nightly-2026-07-02 fmt --all`
- `cargo check -p rings-node --features browser_default --no-default-features --target=wasm32-unknown-unknown`
- `cargo check --target=wasm32-unknown-unknown` in `frontend/`
- `cargo clippy --target=wasm32-unknown-unknown --all-targets -- -D warnings` in `frontend/`
- `cargo test -p rings-core --features dummy --quiet` (655 passed, 4 ignored; doc tests 1 passed, 2 ignored)
- `cargo test -p rings-core --features dummy ecc::signers::secp256r1 --quiet`
- `npm run test:frontend-extension-webview`
- `git diff --check`

PR #715, opened 2026-09-02
Harden control API and webview egress

### LINKED ISSUE #700: Security: unauthenticated JSON-RPC + permissive CORS + SSRF on local control API
## Summary

The local JSON-RPC control surface has no authentication, is served with fully permissive CORS, and exposes SSRF primitives. On a shared or misconfigured host this turns the "localhost-only" bind into a remotely reachable full-node-control surface.

## Details

There is no bearer token, API key, or signature check anywhere on either RPC listener; the only access control is a per-listener method allowlist. Both listeners start unconditionally.

1. **No auth + permissive CORS + no CSRF (highest impact).**
   - Both listeners apply `CorsLayer::permissive()` (`crates/node/src/native/endpoint/mod.rs:73` and `:102`), i.e. `Access-Control-Allow-Origin: *`.
   - The handler accepts a raw `String` body with no content-type / CSRF check. A `text/plain` POST is a CORS "simple request" and skips preflight.
   - Consequence: any web page the operator opens in a same-host browser can cross-origin `POST http://127.0.0.1:50000/` and drive the **full internal method set** (drive-by / CSRF-to-localhost). Permissive CORS also lets the page read responses.

2. **`external_api_addr` accepts non-loopback with no auth.** Default is `127.0.0.1:50001` (`crates/node/src/native/config.rs:40`), but the value is a free-form string fed straight into `addr.parse::<SocketAddr>()` (`endpoint/mod.rs:86`); nothing rejects `0.0.0.0`. Settable via CLI/env/YAML, so `--external-api-addr 0.0.0.0:50001` silently publishes the unauthenticated external API to the network.

3. **SSRF.** `connectPeerViaHttp` / `connectWithSeed` (`crates/node/src/rpc_impl.rs:32`, `:82`) fetch caller-supplied URLs; the seed form fans out and amplifies. These do not go through the `is_public_exit_ip` guard that the onion egress path already implements (`crates/node/src/onion/target.rs:182`).

4. **Storage disclosure.** `nodeInfo` and GET `/status` on **both** listeners dump the entire local DHT persistence + cache stores, routing table, and connected DIDs (`crates/core/src/inspect/snapshot.rs:12`, `:110`; `endpoint/mod.rs:137`). No secret-key/`SessionSk` disclosure path was found over RPC.

## Proposed fix

- Add a local auth token (random at startup, written to a file next to the config; clients send it as a header) on both listeners.
- Replace `CorsLayer::permissive()` with an explicit origin allowlist and disable wildcard for credentialed requests; require content-type / a CSRF guard.
- Reject non-loopback `external_api_addr` unless auth is explicitly configured (or document that exposing it is opt-in-with-auth).
- Route `connectPeerViaHttp` targets through `is_public_exit_ip`.
- Consider gating `nodeInfo`/`/status` storage dumps behind auth.

### LINKED ISSUE #702: Webview: no SSRF/host filtering at TargetUrl::parse (internal/loopback targets admitted)
## Summary

The webview gateway admits any HTTP(S) target host, including loopback, link-local, and private ranges. It provides no SSRF defense of its own; the only host filtering lives in the node's onion egress path, which is not co-located and does not cover the `http://` scheme the webview admits.

## Details

`TargetUrl::parse` validates **only the scheme**, nothing about the host (`crates/webview/src/url.rs:19`):

```rust
match inner.scheme() {
    "http" | "https" => Ok(Self { inner }),
    scheme => Err(WebviewError::UnsupportedScheme(scheme.to_string())),
}
```

So `http://127.0.0.1/`, `http://[::1]/`, `http://169.254.169.254/latest/meta-data/`, RFC1918, and internal DNS names all parse and route straight to the gateway (`crates/webview/src/route.rs:94`). `GatewayRoutePolicy` validates only the controlled-origin shape, never the target host.

Current mitigation and its gaps: the native onion egress path *does* defend well — `is_public_exit_ip` (`crates/node/src/onion/target.rs:182`) rejects loopback / `169.254.0.0/16` / RFC1918 / CGNAT / documentation / multicast / IPv6-embedded-IPv4 and pins resolved addresses to block DNS-rebinding. But:

1. No defense-in-depth — a deployment that wires a different/naïve `GatewayTransport` (the trait in `crates/webview/src/transport/mod.rs` has no concrete impl in this crate) has SSRF wide open.
2. The reviewed guard handles **HTTPS only**; `TargetUrl` also admits `http://`, which is not covered.
3. The filter lives in `node`, not co-located with URL admission in `webview`, so the two can drift.

## Proposed fix

Add a transport-independent host/IP-literal rejection at `TargetUrl::parse` (or in `route`), covering both `http://` and `https://`, as a hard gate independent of the downstream transport.

## Note

CORS credentialed-origin handling, header CRLF validation, cookie jar bounds + PSL checks, and the sandbox CSP (no `allow-same-origin`) in this crate are implemented carefully and are not part of this issue.

RyanKung:
## Summary
- add shared public host/IP policy and apply it to onion targets, remote RPC targets, and webview TargetUrl parsing/deserialization
- require bearer-token auth for native control API routes, tighten CORS/content-type handling, and require explicit opt-in for non-loopback external API binds
- harden connectPeerViaHttp/connectWithSeed against SSRF with static host checks, native DNS resolve/filter/pin, no redirects, and no ambient proxies

Closes #700.
Closes #702.

## Validation
- cargo test -p rings-network-policy
- cargo test -p rings-webview --lib --all-features
- cargo test -p rings-node --lib --all-features
- cargo test --workspace --all-targets --all-features --exclude rings-webview
- cargo clippy -p rings-network-policy -p rings-webview -p rings-rpc -p rings-node --all-targets --all-features -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing
- cargo clippy -p rings-node --features browser --no-deps --no-default-features --target=wasm32-unknown-unknown --lib -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::indexing_slicing
- cargo test --doc -p rings-network-policy -p rings-webview -p rings-node -p rings-rpc --all-features
- cargo +nightly-2026-07-02 fmt --all -- --check
- taplo format --check
- git diff --check

Note: browser-backed webview integration tests need the local Playwright Chromium executable, which is not installed in this environment; the library-level regression coverage passed.


RyanKung:
**Correctness: native `remote_rpc_client` can never connect to an IPv6-literal endpoint**

`validate_remote_rpc_url` accepts IPv6 literals (there's even a passing assertion for `https://[2606:4700:4700::1111]/rpc`), but the native connect path in `crates/node/src/rpc_impl.rs` derives the host from `parsed.host_str()`. For an IPv6 literal that returns the **bracketed** form `"[2606:4700:4700::1111]"`. That string is handed to `OnionProxyTarget::new(host, port)` → `resolve_public_target` → `lookup_host`, and since `"[...]"` does not parse as an `IpAddr`, it is treated as a **hostname** and sent to DNS:

- On a normal resolver the brackets/colons make it an invalid name → `EAI_NONAME` → `Error::OnionTargetResolve`.
- Where a resolver hijacks NXDOMAIN, it resolves to a bogus address (I measured `198.18.6.113`, in the 198.18/15 benchmarking range) which the public-address filter then rejects → `Denied`.

Either way it **fails closed** (no SSRF exposure), but a target that `validate_remote_rpc_url` accepts — and that the test suite asserts is valid — can never actually be reached. IPv4 literals are unaffected: `host_str()` returns them unbracketed, so `lookup_host` parses them directly. The gap is invisible to CI because the tests only exercise `validate_remote_rpc_url` (pure parsing), not the async `remote_rpc_client` resolve path.

Suggested fix: branch on `parsed.host()` (the `url::Host` enum) instead of `host_str()`. For `Host::Ipv4`/`Host::Ipv6`, build the client against the literal directly (a literal URL already bypasses reqwest's `resolve_to_addrs` override, so no DNS/pinning is needed — and `validate_public_url_host` has already vetted the literal as public); reserve `resolve_public_target` + address pinning for `Host::Domain`. That fixes IPv6 and also avoids a pointless DNS round-trip for every literal target.

RyanKung:
Thanks — the IPv6-literal fix (`remote_rpc_resolution_target` branching on `url::Host`, skipping DNS/pinning for literals) is correct, and the `frontend` `TargetUrl → Url` change is safe: egress targets are still the validated `TargetUrl` produced by `policy.route()` and re-checked at `GatewayGuard::send`, while `controlled_origin`/`requested` are only trust-anchor/input values.

But the same refactor introduced a **regression in the native `remote_rpc_client` DNS pin: a trailing-dot FQDN bypasses it and reopens the rebinding vector this code exists to close.**

The pin key changed from the URL host to the normalized `target.host()`, and the two differ for a trailing-dot name:

- `url::Host::Domain` keeps the trailing dot: `"example.com."`. `OnionProxyTarget::normalize_host` strips it (`trim_end_matches('.')`), so `target.host() == "example.com"`.
- v2 pins with `resolve_to_addrs(target.host(), &addresses)` → override map keyed by the **dot-less** `"example.com"`.
- At connect time reqwest's `DnsResolverWithOverrides::resolve` does `overrides.get(name.as_str())`, and `name` comes from the URI host. hyper-util only trims brackets, **not the trailing dot** (`connect/http.rs`: `host.trim_start_matches('[').trim_end_matches(']')`), so the lookup key is the **dotted** `"example.com."` → **miss** → it falls back to the system resolver and **re-resolves at connect time, unpinned**.

v1 was correct here because it keyed the override with `host = parsed.host_str()` (dot preserved), matching what reqwest uses. The refactor swapping the key to `target.host()` is what broke it.

**Exploit:** an attacker-supplied `connectPeerViaHttp` / `connectWithSeed` URL of `https://evil.example./` (note the trailing dot). `validate_public_domain` trims the dot before its checks → passes; `resolve_public_target` resolves the dot-less name to a public IP → passes the filter; but the connect-time resolution is of the **dotted** name and is **not pinned**, so it can rebind to `127.0.0.1` / `169.254.169.254`. This is exactly the #700 SSRF pinning defense, silently bypassed by appending one dot. Native-only (the browser path deliberately does not pin).

**Fix:** the pin key must equal the host string reqwest uses at connect time. Keep resolving via the dot-less `target.host()` (that's what DNS wants), but pin with `parsed.host_str()`:

```rust
builder = builder.resolve_to_addrs(parsed.host_str().unwrap_or(target.host()), &addresses);
```

or normalize the trailing dot out of the URL host up front so validation, resolution, pinning, SNI, and the Host header all agree.

The current tests don't catch this: they only assert `remote_rpc_resolution_target`'s return value — there's no trailing-dot case and no check that the pin key matches the connect-time host. A regression test resolving a domain and asserting the override key equals `parsed.host_str()` (including the trailing-dot form) would lock it down.

PR #716, opened 2026-09-02
build: fix the release pipeline (complete publish set, exclude examples, auto-tag)

RyanKung:
Fixes several problems in the crates.io publish path (`make publish`) and adds real-release automation on version bumps.

## 1. Exclude example crates from publishing

`cargo publish --workspace` for 0.20.0 tried to upload the two example crates, which are not meant to be published:

```
error: failed to publish rings-native-example v0.20.0 to registry at https://crates.io
Caused by: ... missing or empty metadata fields: description
```

Set `publish = false` on `examples/native` and `examples/relay`, matching the existing `examples/dweb` convention. This also clears the `manifest has no description` warnings.

## 2. Complete the publishable crate set in `make publish`

`rings-codec`, `rings-gateway` and `rings-webview` were split out as **new crates in 0.20.0** but never added to `scripts/cargo-publish-crates.sh`, so the script only published 6 of the 9 library crates. Since `rings-codec` is a hard dependency of transport/core/node and `rings-gateway` an optional dependency of node, `cargo publish -p rings-core` cannot succeed for a fresh version until they are indexed — the 0.20.0 release only completed via an ad-hoc `cargo publish --workspace`.

- Add all three to `CRATES` in dependency-topological order (codec/gateway before their dependents).
- Add codec/gateway to `INTERNAL_CRATES` for metadata + index-wait validation (`rings-webview` is an independent leaf with no `[workspace.dependencies]` entry, so it stays out).
- Register their manifest paths.

`scripts/cargo-publish-crates.sh check` now validates all 9 crates, and the script publishes the same set as `cargo publish --workspace`. (crates.io publishing stays a manual `make publish`; this PR does not automate it.)

## 3. Auto-tag version bumps to trigger a real release

New `Auto Tag` workflow: on a master push touching `Cargo.toml`, it reads `[workspace.package].version` and — when the matching `vX.Y.Z` tag does not yet exist — creates and pushes it, then dispatches `Auto Release` for that tag.

A tag pushed with the built-in `GITHUB_TOKEN` does not trigger another workflow (GitHub suppresses recursive runs), so `Auto Release`'s `push: tags: v*` path would never fire on its own; the workflow pushes the tag for provenance and dispatches `Auto Release` explicitly. It runs **alongside** the existing rolling `prerelease` (unchanged), adding a real versioned release only when the version changes.

### DIFF
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -1,0 +1,76 @@
+name: Auto Tag
+
+# Turn a version bump into a real release. When a PR that bumps
+# `[workspace.package].version` lands on master, this creates and pushes the matching
+# `vX.Y.Z` tag and kicks off `Auto Release` for it.
+#
+# Why an explicit dispatch instead of relying on the tag push: a tag pushed with the
+# built-in `GITHUB_TOKEN` does not trigger another workflow (GitHub suppresses recursive
+# runs), so `Auto Release`'s `push: tags: v*` path would never fire. We push the tag for
+# provenance (it points at the merge commit) and then dispatch `Auto Release` ourselves.
+#
+# This runs alongside the existing rolling `prerelease` that `Auto Release` publishes on
+# every master push; it only adds a real, versioned release when the version changes.
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - Cargo.toml
+
+permissions:
+  contents: read
+
+concurrency:
+  group: auto-tag-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  tag:
+    name: tag version bump
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      actions: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@a37ce9120846195fa4ece8f58b268e6043cb2f26 # v3
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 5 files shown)


RyanKung:
Expanded scope beyond the initial `publish = false` fix, per the release-pipeline discussion:

- `9f65867a` — **ci: auto-tag version bumps** — new `Auto Tag` workflow; tags `vX.Y.Z` on a version-bump merge and dispatches `Auto Release`. Runs alongside the existing per-push prerelease.
- `d20e67c5` — **fix(publish): complete the publishable crate set** — added `rings-codec`, `rings-gateway`, `rings-webview` (new 0.20.0 crates) to the publish script in dependency order; `make publish` alone can now do a full release. `scripts/cargo-publish-crates.sh check` passes for all 9 crates.
- `535ef5c1` — original `publish = false` on the two example crates.

RyanKung:
CI failed here on the flaky `test_provider_exposes_sent_and_received_peer_measurements` (`evidence.sent >= 1` at test_network.rs:274) — the same pre-existing wall-clock race that #718 fixes, unrelated to this PR's changes.

Cherry-picked that fix onto this branch (`695e7725`) so this PR is green on its own. The fix also stands alone in #718; whichever merges first, the identical patch dedupes on the other via rebase/merge (and #718 can then be closed).

PR #722, opened 2026-09-02
test(core): await measurement event, not wall clock, in chunk-e2e delta test

### LINKED ISSUE #721: Flaky: test_whole_and_chunked_payloads_have_identical_measurement_delta (1s wall-clock wait on detached sender measurement)
## Symptom

`rings-core` test `tests::default::test_chunk_e2e::test_whole_and_chunked_payloads_have_identical_measurement_delta` fails intermittently in the `Build and test (ubuntu-latest)` job:

```
thread '...test_whole_and_chunked_payloads_have_identical_measurement_delta' panicked at
crates/core/src/tests/default/test_chunk_e2e.rs:175:25:
    ... send must be measured once
```

Line 175 is the `.unwrap_or_else(|_| panic!(...))` of `wait_for_test_condition`, i.e. its **1-second wall-clock `tokio::time::timeout` expired** while polling for the sender-side measurement counter to advance.

Observed on run [33619797818](https://github.com/RingsNetwork/rings/actions/runs/33619797818) (re-run of the same commit then passed — confirming it is timing, not a deterministic failure).

## Root cause

The sender-side measurement is applied on a **detached, asynchronous pipeline**, so `send_direct_message().await` returning does *not* imply the `Sent` measurement has landed:

```
send_direct_message().await returns        // detached: only enqueued to the outbound mailbox,
                                           // not yet dispatched, not yet measured
  -> outbound worker dispatches the frame, then record(OutboundMeasurement::Sent)   // background task
  -> wakes the measurement worker (1-slot wake channel)
  -> measurement worker drains + measure.record_batch(Sent)   // only now the CountingMeasure event exists
```

The test bridges this gap by polling `sender_measure.logical_transfer_count(...) > sent_before` under a **1s timeout** (`wait_for_test_condition`, test_chunk_e2e.rs:168-176). Under a loaded CI runner (this binary runs 659 tests) the three async hops (mailbox -> dispatch/record -> wake -> apply) can occasionally exceed 1s, so the bounded poll times out.

Notes:

- **Not a lost-wake / correctness bug.** The wake channel (`mpsc::channel(1)`, `outbound/measurement.rs`) only carries a wake *signal*; the pending aggregate is held separately and fully drained on the next wake, so a dropped `try_send` never loses state. The counter *will* reach the expected value — just not always within 1s.
- **Asymmetry with the receiver side.** The receiver-side assertions (test_chunk_e2e.rs:237-240, 263-267) use no wait and are stable, because `recv_custom` returning establishes happens-before for the inline receive measurement. The **sender has no such natural synchronization point**, which is why only the sender path needs a wait.
- **Distinct from the already-merged flaky fix.** Commit `e8f91ee` ("stop the sub-second test reliability window from flaking peer-measurement tests") addressed a *windowed-evidence reset* flake in `crates/node/src/measure.rs` — a different mechanism in a different crate. This one is the outbound measurement-application latency in `rings-core`.

## Fix direction (not yet implemented)

The wait itself **cannot simply be removed** and the assertion cannot collapse to a bare `assert_eq!`: because the send is detached, at the moment the send returns the `Sent` event is not yet even in the pending aggregate, so any "quiesce / pending-empty" barrier would pass prematurely. The sender genuinely has no happens-before point for measurement application; a wait for "applied count reached the expected value" is intrinsic unless the outbound + measurement workers are driven on a controllable single-step executor (a larger re-architecture, not a test fix).

What should change is the *shape* of the wait, removing the flaky sub-second wall-clock bound:

1. **Preferred (event-driven, unbounded):** have the measurement worker signal (`Notify`, `#[cfg(test)]`) after each applied batch; the test waits on that notification until `count >= sent_before + 1` with no wall-clock cap (the global `cargo test` timeout still bounds a genuine regression). Keep the exact `assert_eq!(count, sent_before + 1)` afterward to catch over-counting.
2. **Lighter:** drop only the 1s cap from `wait_for_test_condition` and poll unboundedly on the real condition.

Explicitly rejected: merely enlarging the 1s window (e.g. to 5s/10s) — still a race, just rarer.

Affected: `crates/core/src/tests/default/test_chunk_e2e.rs` (the test + its `wait_for_test_condition` helper), with a `#[cfg(test)]` notification hook in `crates/core/src/swarm/transport/outbound/measurement.rs` for option 1.

RyanKung:
## Summary

Fixes the flaky `test_whole_and_chunked_payloads_have_identical_measurement_delta` (`rings-core`, dummy backend) by turning its wall-clock poll into an **event wait**.

The sender-side `Sent` measurement lands on a **detached pipeline** — `send_direct_message` returns after enqueueing, then `outbound dispatch → record → wake → coalescing worker applies`. The test bridged that gap by polling the counter under a **1s `tokio::time::timeout`** (`wait_for_test_condition`); under CI load the async hops occasionally exceed 1s and it panicked `"... must be measured once"`. Root-cause writeup: #721.

## Why the wait can't just be deleted

`send_direct_message` is detached, so at the moment it returns the `Sent` event is not yet even in the aggregate — the sender has **no happens-before point** for measurement application (the receiver does: `recv_custom` returning). A "queue empty / quiesce" barrier would pass prematurely. A wait for *the measurement landing* is intrinsic.

## Change

`CountingMeasure` (the test's own `Measure` double) gains a monotonic `watch` generation, bumped **once each recording is fully applied** (event and counters both visible). `await_recorded(predicate)` parks on that generation and re-checks the predicate **only when a recording lands**:

- **No wall clock** — it cannot flake on a slow runner.
- **No lost wake** — `watch` retains the latest value, so a bump racing the predicate re-check is still observed (`wait_for` checks immediately, then on each change). The bump lands only after the recording's counters are visible, so a predicate over either `events` or `counters` is sound.
- **Better failure mode** — a measurement that genuinely never lands now surfaces as a **deterministic hang** (caught by the CI job `timeout-minutes`), not a 1% timeout. This directly answers the "does a specific condition cause a real timeout?" question: such a condition would now be unmissable rather than intermittent.

Scoped to the two measurement-counter waits in this one test; the other `wait_for_test_condition` call sites gate on different signals (e.g. `dummy_controlled` counts) and are left unchanged. Entirely test-only — no production or `#[cfg(test)]` hook added.

## Verification

- `cargo test -p rings-core --features dummy` (full CI-equivalent suite): **655 passed, 0 failed**.
- The target test run 30× in isolation: 30/30 pass, converging in ~1.3s each with no timeout.
- `cargo +nightly fmt` applied.

Closes #721

### DIFF
--- a/crates/core/src/tests/default/test_chunk_e2e.rs
+++ b/crates/core/src/tests/default/test_chunk_e2e.rs
@@ -9,6 +9,7 @@ use std::sync::Mutex;
 use async_trait::async_trait;
 use rings_transport::connections::dummy_controlled;
 use rings_transport::core::transport::WebrtcConnectionState;
+use tokio::sync::watch;
 use tokio::time::sleep;
 use tokio::time::Duration;
 
@@ -55,10 +56,25 @@ use crate::tests::manually_establish_connection;
 use crate::tests::multi_frame_storage_sync_entries;
 use crate::tests::outbound_capacity_released;
 
-#[derive(Default)]
 struct CountingMeasure {
     counters: Mutex<Vec<(crate::dht::Did, MeasureCounter)>>,
     events: Mutex<Vec<(crate::dht::Did, MeasurementEvent)>>,
+    /// Monotonic generation bumped after every recorded event. It lets a test await
+    /// a specific measurement *landing* rather than poll a wall-clock window: `watch`
+    /// retains the latest value, so a bump that races a waiter's predicate re-check is
+    /// never lost, and a measurement that never lands surfaces as a deterministic hang
+    /// (bounded by the test harness) instead of an intermittent timeout.
+    recorded: watch::Sender<u64>,
+}
+
+impl Default for CountingMeasure {
+    fn default() -> Self {
+        Self {
+            counters: Mutex::new(Vec::new()),
+            events: Mutex::new(Vec::new()),
+            recorded: watch::channel(0).0,
+        }
+    }
 }
 
 impl CountingMeasure {
@@ -96,6 +112,18 @@ impl CountingMeasure {
             Err(_) => 0,
         }
     }
+
+    /// Await until `predicate` holds, re-checking only when a new measurement is
+    /// recorded. This is an event wait, not a timed poll: it parks on the `recorded`
+    /// generation and never consults the wall clock, so it cannot flake on a slow
+    /// runner. `wait_for` evaluates `predicate` immediately and again on each bump.
+    async fn await_event(&self, predicate: impl Fn(&Self) -> bool) {
+        let mut recorded = self.recorded.subscribe();
+        // The sender outlives every waiter (it is owned by this measure), so the
+        // only `Err` is an impossible closed channel; discard it and let the parked
+        // future resolve on the next recorded event.
+        let _ = recorded.wait_for(|_generation| predicate(self)).await;
+    }
 }
 
 #[async_trait]
@@ -122,6 +150,7 @@ impl Measure for CountingMeasure {
         if let Ok(mut events) = self.events.lock() {
             events.push((did, event));
         }
+        self.recorded.send_modify(|generation| *generation += 1);
         self.incr(did, MeasureCounter::from_event(event)).await;
         Ok(ApplyOutcome::Applied)
     }
@@ -138,6 +167,7 @@ impl Measure for CountingMeasure {
         if let Ok(mut events) = self.events.lock() {
             events.push((did, batch.event()));
         }
+        self.recorded.send_modify(|generation| *generation += 1);
         let counter = MeasureCounter::from_event(batch.event());
         for _ in 0..batch.occurrences().get() {
             self.incr(did, counter).await;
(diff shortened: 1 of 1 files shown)


RyanKung:
Self-review pass — pushed cd0972c7 on top of a8e343a3.

**Real defect fixed (soundness):** in a8e343a3 the `recorded` generation was bumped *before* `incr` ran, so it signalled "event pushed" but not "counters updated". A predicate over `counters` (`count()`) would re-check on the bump, observe stale counters, and never be woken again (since `incr` does not bump) — a latent hang. The two predicates in this test read `events` (pushed before the bump) and so happened to be sound, but the invariant was not. The bump now lands after `incr` in both `record` and `record_batch`, so the generation means "this recording is fully applied" and `await_recorded` is sound for a predicate over either `events` or `counters`.

**Also in this commit:**
- `watch::Sender::new(0)` instead of `watch::channel(0).0` (no throwaway receiver; tokio 1.53).
- Rename `await_event` → `await_recorded`: it takes a predicate, not an event, and the name now pairs with the `recorded` field and names the wake source.
- Doc precision: a never-landing measurement is bounded by the CI job `timeout-minutes`, not "the test harness" — `cargo test` has no per-test timeout.

**Verified unchanged:** `send_modify` (not `send`) is required since bumps occur before any subscriber exists; `wait_for` checks the current value immediately even if already seen, and `watch` retains the latest value, so no lost wake; no lock-order cycle (`record` never holds `events`/`counters` while taking the watch write lock).

Full `cargo test -p rings-core --features dummy`: 655 passed, 0 failed. Target test 30× isolated: 30/30.

Issue #737, opened 2026-09-09
Document the communication-layer / privacy-layer split in README.md and SECURITY.md


RyanKung:
## Problem

README.md and SECURITY.md describe Rings as one overlay with a threat model, and mention a "privacy layer" only as roadmap language. The code already has two layers with different security contracts, and the documents do not draw the line between them:

- **Communication layer** (`crates/core`: Chord routing, `MessageRelay`, `MessagePayload`, session-key signatures, ElGamal to a DID). Its contract is payload confidentiality and authenticity. It cannot offer unlinkability: Chord routes by the destination DID, and every payload names its origin through the transaction signature. The most it can do is not leak *more* than that (see #736 for the plaintext hop history).
- **Privacy layer** (`crates/node/src/onion`: layered ElGamal-AEAD circuits over direct edges, fixed-batch cover cells with pacing, replay witnesses, size buckets, route selection from the onion-relay/exit registries). This is where per-hop knowledge is bounded to predecessor + successor and where traffic shaping lives.

Without the split, two failure modes recur: privacy properties get attributed to the plain relay (users assume DID encryption means anonymity), and communication-layer leaks get treated as privacy-layer bugs that cover traffic is expected to fix.

## Proposal

Document the split in both files, with the rule: **the communication layer minimizes leakage; the privacy layer provides privacy.**

### README.md

- In *Architecture*, add a **Privacy** row between *Protocols* and *Overlay* mapped to `crates/node/src/onion` (circuits, cover, pacing, exit registry), and describe the *Overlay* row as routing and confidentiality only.
- In *Features*, add a short *Privacy layer* item stating what circuits hide (route hops from one another; the client from the exit) and what they do not (the client from its first hop; membership; timing against a global observer). Link to the SECURITY.md section.
- Keep the roadmap sentence, but point at the existing module rather than implying the layer is future work.

### SECURITY.md

- Add a top-level section *Layer Contracts* with two subsections:
  - *Communication layer*: what every hop and the destination learn on the plain relay (origin DID, destination DID, predecessor and successor edges, sizes, timing) and the leak-minimization obligations of this layer (no hop history on the wire after #736, no telemetry beyond what routing needs, payload encrypted to the destination DID).
[...]

Issue #710, opened 2026-09-02
DRanking: versioned service receipts and a provisional probe vertical slice


RyanKung:
## Summary

Define a canonical, versioned service-receipt model and land one end-to-end provisional probe-receipt vertical slice.

Before the admission ledger exists, collected receipts are provisional evidence observed by local nodes. They are not finalized DRanking input, do not create rank, and must not change order_peers_by_quality. When ledger-delimited epochs land, the network performs a deliberate receipt-protocol cutover; provisional time-slice receipts are not silently reinterpreted as ledger receipts.

The first vertical slice is probe because its request, authenticated response, unit count, and beneficiary acknowledgement can be defined without claiming durable storage or hidden-path relay delivery. Relay and store require separate evidence profiles after the common carrier is proven.

## Security claim

A valid provisional receipt proves that:

- the provider and beneficiary account identities are fixed by delegated session proofs;
- both delegated session keys signed the same canonical claim in distinct role domains;
- the collector accepted the acknowledgement while both attestations and the provisional epoch were live;
- the receipt has not already been admitted under its epoch-scoped nonce key.

It does not prove:

- external truth beyond the signed probe transcript;
- that a provider stayed live after the response;
- ledger inclusion or finality;
- globally unique issuance from the receipt bytes alone;
- Sybil resistance, honest service value, or trusted weight;
- that a session-signed receipt observed before the ledger remains safe against arbitrary historical backfilling after delegated-key compromise.

The last limitation is why these receipts remain provisional. Future ledger admission supplies the ordering/finality boundary that this issue does not have.

## Canonical model

Add explicit v1 provisional types in rings-core:

    ProvisionalServiceClaimV1 {
        network_id: u32,
        service: ServiceKindV1,
        provider_account: Did,
        beneficiary_account: Did,
        epoch: ProvisionalEpochV1,
        nonce: [u8; 32],
        units: NonZeroU64,
        request_digest: [u8; 32],
        completion_digest: [u8; 32],
    }

    ProvisionalServiceReceiptV1 {
        claim: ProvisionalServiceClaimV1,
        provider_attestation: MessageVerification,
        beneficiary_attestation: MessageVerification,
    }

For this issue, ServiceKindV1 admits only Probe. Do not put unimplemented Relay or Store variants on the accepted wire and then assign them placeholder verification semantics.

The canonical claim digest binds every field. Provider and beneficiary signatures use distinct, explicit domains so the signatures cannot be exchanged:

[...]

PR #745, opened 2026-09-09
Path-less relay: hop budget replaces the hop history, reports route to the origin (#736)

### LINKED ISSUE #736: Replace the plaintext relay hop history with a path-less relay (hop budget + routed reports)
## Problem

`MessageRelay` (`crates/core/src/message/protocols/relay/mod.rs`) carries a push-only `path: Vec<Did>` that every forwarding hop appends itself to. As a result:

- Every intermediate hop sees the full list of nodes that handled the message before it.
- The destination receives the complete hop history of every message it is sent.
- A single relayed message is a topology dump of one Chord route, which composes badly with the topology-diversity goals in `SECURITY.md` and with any future sender-unlinkability work.

This is orthogonal to sender identity: `transaction.verification` already names the origin, so `path` leaks *topology*, not the origin. Hiding the origin is a separate concern (onion messaging) and is out of scope here.

## What `path` is used for today

1. Reverse-path report return (`ReportReturnPolicy::Path` via `MessageRelay::path_report`).
2. Origin lookup through `try_origin_sender` / `origin_sender` (`handlers/connection.rs`, `handlers/stabilization/mod.rs`, `handlers/storage/mod.rs`).
3. Loop detection (`has_infinite_loop`, adjacency check in `validate`).
4. Inspection / debug output.

Note that `path` is **not** under the transaction signature: `hash_transaction` covers `destination`, `tx_id`, `report_return`, and `data` only. The relay is re-signed per hop, so changing its shape does not touch origin authentication.

## Proposal

Replace the accumulating path with a path-less relay:

- `MessageRelay` keeps `next_hop` and `destination` only, plus a monotone hop budget (TTL) that each `forward` decrements. A message whose budget reaches zero is dropped with a distinct error; this replaces `has_infinite_loop`. Chord greedy routing is monotone toward the target, so a loop is a fault, not a normal state, and a budget bound is the right witness for it.
- `ReportReturnPolicy::Path` is removed; `Routed { destination }` (already present since #724, already bound to the transaction signer) becomes the only report-return policy, and the `report_return` discriminant stays in the signed transaction hash.
- `try_origin_sender` / `origin_sender` are removed from the relay. The three protocol handlers that need the origin take it from `transaction.signer()`, which is what they are actually authorizing against.
- Debug output loses the hop list; per-hop tracing (`forward` at each node) covers the inspection use case without shipping the history on the wire.

Each hop then learns exactly its predecessor (transport edge) and its successor (`next_hop`); the destination learns only the last hop. This matches the per-hop knowledge bound the onion circuit data plane already documents for itself in `crates/node/src/onion/circuit/mod.rs`.

## Explicitly not addressed

- The origin DID remains visible to every hop through the transaction signature.
- The destination DID remains visible to every hop; Chord routes by it.
- No timing or size hiding.

`SECURITY.md` should state this boundary when the change lands: payload confidentiality and per-hop topology hiding, but no sender or receiver unlinkability on the plain relay.

## Acceptance

- No `path` field on the wire; no relay-derived origin accessor.
- Hop-budget exhaustion has a test that constructs a forwarding cycle and observes the drop.
- Report return uses `Routed` end to end, including the wasm fixture.
- Existing relay tests that asserted on `path` contents are rewritten around `next_hop` / `destination` / budget.

RyanKung:
Closes #736.

## What changes

`MessageRelay` no longer records the route. The carrier is `{ next_hop, destination, hop_budget }`:

- **`HopBudget`** is a `u8` newtype with invariant `0 ≤ n ≤ MAX_RELAY_HOPS`, established by every constructor and by decoding (`#[serde(try_from = "u8")]`), so a carrier received from a peer can never claim more forwards than a fresh one. `spend : HopBudget → Option<HopBudget>` is the predecessor on the finite chain `MAX > … > 0`; `MessageRelay::forward` spends one unit and returns the new typed error `RelayHopBudgetExhausted` at the bottom. This replaces `has_infinite_loop`: Chord greedy routing is monotone toward the target, so a route that outruns its budget is a loop, not a long ring.
- **Fresh budget** = `MAX_RELAY_HOPS = 64`, a network constant: a greedy route over a finger table that spans the identifier space takes about `log2 N` hops, so 64 covers any overlay the relay routes over; a ring whose table does not span the space (the 8-slot tables of simulated networks, whose slots `did + 2^i` all resolve to the immediate successor) routes by successor walk, whose length is the ring size and is known to no node, so no per-ring derivation exists and 64 covers such rings up to 65 nodes. The sync-storm size matrix (Ring topology, 50 nodes, one edge chosen against the clockwise direction) is the witness. A report is a fresh carrier; `reset_destination` keeps the budget (only `forward` spends it); a chunk frame, which crosses exactly one edge, is sent with `HopBudget::EXHAUSTED`.
- **Reports route to the origin.** `ReportReturnPolicy` and `Transaction::report_return` are removed together with the `*_with_report_return` senders. With the reverse-path policy gone, `Routed { destination }` was forced equal to the signer by `validate_authorized_by` and carried no information on the wire (21 bytes per transaction plus a constant byte in the hash); the receipt shape in #710 (`⟨srv, a, b, k, n, σ_b⟩`, countersignature returned to `a`) needs no other return address either. `send_report_message` now routes to `Transaction::origin`. The transaction hash is `destination ‖ tx_id ‖ data`.
- **`Transaction::origin`** is the account behind the signing session (`Session::account_did`), never the session id: it is the ring position a request is authorized against and the return address of its report. The three handlers that took the origin from the relay path (`connection.rs`, `stabilization/mod.rs`, `storage/mod.rs`) take it from there; the relay-derived accessors `sender`, `origin_sender`, `try_origin_sender` are gone.
- **Errors:** `RelayHopBudgetExhausted` and `RelayHopBudgetAboveMax(u8)` added; `InvalidRelayPath`, `InfiniteRelayPath`, `InvalidRelayDestination`, `CannotInferNextHop` removed (the last two had no callers; `InvalidRelayDestination`'s message still described the path).

## Test transport made causal

Routing reports exposed a causality gap in the dummy transport: its data channel counted as open while the peer connection was still `Connecting`, so an **answerer admitted the joining peer the instant it produced its answer**, before the offerer had even received it. A report for that peer forwarded through such a hop was sent over the half-admitted edge and dropped by the receiver as "pending connection is not admitted yet" (`test_joining_between_bootstrap_and_successor_connects_successor_hint`, `test_fourth_node_connection`, about one run in five). A reverse-path report only ever used edges the request had crossed, which is why this never surfaced before. In real WebRTC the answerer's channel cannot open before the offerer applies the answer, so the dummy now follows the same law: open iff `Connected`, and both ends reach `Connected` together when the offerer accepts. The fourteen tests that model the browser callback order (channel open while `Connecting`) keep it through the explicit override in `open_dummy_data_channel_before_ice_connected`.

## Frames that arrive before admission are held

With real transports the same asymmetry exists as a small window: each end admits an edge on its own data-channel-open callback, and a verified frame that arrives in between used to be dropped at the gate. It is now held and delivered once admission commits. `swarm::callback::pre_admission::PreAdmissionHold` is the pure state: a bounded queue plus a drain flag. `arrive` passes a frame only through an empty hold with no drain in progress, otherwise the frame queues behind everything held; `begin_drain` grants one exclusive drainer until it sees the empty queue, so held frames leave in arrival order and a frame arriving mid-drain follows them. The bound is the per-peer inbound capacity (32). The hold lives on the per-connection `InboundProcessor`, drains at the end of a successful admission and on any admitted arrival that finds frames queued (the judge-then-commit window is closed by re-reading admission after queueing), and is discarded when the pending handshake is cancelled. Held frames keep their transport capacity lease; their authentication is judged at delivery. Tests: the hold's laws; a held message delivered once by admission itself; arrival order kept ahead of later messages; a cancelled handshake discards its hold.

## Not covered by any signature

The carrier is outside both the transaction and the hop signature (as before: `hash_transaction` never covered the relay). The budget bounds the work honest hops do for one payload; it is not a promise a dishonest hop keeps. The module and SECURITY.md say so.

## Tests

- `protocols/relay/test_relay.rs` rewritten: `spend` law, `for_ring` law, a forwarding cycle `a→b→c→a…` that spends exactly `MAX` forwards and then observes `RelayHopBudgetExhausted`, wrong-`next_hop` rejection, `reset_destination` keeps the budget, `report` is a fresh carrier, decoding rejects a budget above the cap.
- `handlers/custom.rs`: a two-node handler test drops an exhausted payload at the hop with the typed error (destination sees nothing), then forwards the same payload with one unit left and the destination receives it with `hop_budget == EXHAUSTED`.
- `payload/test_payload.rs`: `origin` is the account, not the session id.
- `test_sync_entries_handler_ignores_a_forged_relay_origin_for_a_relay_carrier` is removed: there is no relay origin left to forge.
- The wasm three-node relay test now exercises routed reports end to end because every send and every report is routed; no wasm-specific change was needed.

## Docs

Crate doc in `lib.rs` (the MSRP "reverse the path" section), `effects.rs`, `custom.rs`, `storage/mod.rs`, `storage_sync.rs`, `entry/inbox.rs` no longer mention a relay path. SECURITY.md's communication-layer contract (from #737) is stated in the present tense: the carrier is `next_hop`, `destination`, and a hop budget; a route that outruns its budget is dropped as a loop; the carrier is outside every signature, so the budget binds honest hops only; every report is routed back to the origin. ROADMAP.md lists the path-less relay as shipped.

## Version

0.23.0. Wire format and transaction hash are incompatible with 0.22.x; every node in an overlay upgrades together (the node.rings.rs seed after this epoch's PRs land).

## Gates run locally

`cargo clippy --all --tests -- -D warnings`, core wasm clippy (`--features wasm --target wasm32-unknown-unknown --tests`), node `browser_default` wasm clippy, node `ffi` and `node --lib` clippy, `RUSTDOCFLAGS=-Dwarnings cargo doc --all --no-deps`, `cargo +nightly fmt --all -- --check`, `taplo format --check`, `typos`, `cargo test -p rings-core --features dummy`, the sync-storm size matrix (`test_pr_ring_and_hotspot_size_matrix --ignored`), `cargo test -p rings-transport --features dummy`, `cargo test -p rings-node --lib`.


crates/core/src/dht/entry/inbox.rs
```diff
@@ -344,7 +344,7 @@ impl EntryOperation {
 /// anyone else is not invalid, only not this receiver's to take yet, so the law is a predicate
 /// the batch skips on rather than an error it fails with.
 ///
-/// Pre: `sender` is the authenticated signer of the hand-off, not its peer-declared relay path.
+/// Pre: `sender` is the authenticated origin of the hand-off.
 pub(crate) fn relocates_from_predecessor(sender: Did, predecessor: Option<Did>) -> bool {
```
RyanKung, line 348: you sure?

crates/core/src/message/handlers/connection.rs
```diff
@@ -130,7 +130,7 @@ impl HandleMsg<ConnectNodeSend> for MessageHandler {
             tracing::trace!(
                 local = %self.dht.did,
                 tx_id = %ctx.transaction.tx_id,
-                origin = ?ctx.relay.try_origin_sender().ok(),
+                origin = %ctx.transaction.origin(),
```
RyanKung, line 133: what is % means, new syntax?

PR #747, opened 2026-09-09
Generate decode-boundary tests in CI (#741)

### LINKED ISSUE #741: Fuzz the network decode boundaries (wire envelope, chunk framing, JSON-RPC body, handshake SDP) with CI smoke runs
## Problem

Every byte a Rings node accepts from the network passes through a small set of hand-written or serde-derived decoders, and none of them has a fuzz target. The CI gates added in #670 (cargo-deny, Miri on core invariants, ASan/LSan on the FFI lifecycle) check memory safety of code paths that tests already exercise; they do not search for inputs that make a decoder panic, allocate without bound, or recurse. For a 1.0 that claims a reviewed threat model, the decode boundaries are the one place where an unauthenticated peer controls the input byte-for-byte, so they need coverage that is driven by the input rather than by the author's expectations.

## Boundaries to cover

Each target takes `&[u8]` and must return without panicking; the only acceptable outcomes are `Ok` or a typed `Err`.

1. **Wire envelope**: `Encoded::decode::<MessagePayload>` in `crates/core/src/message/encoder/mod.rs` (base58-check → decompress → deserialize), and the `MessagePayload` deserialize + `verify` path in `crates/core/src/message/payload/mod.rs`. This is the whole inbound relay surface.
2. **Chunk framing**: `Chunk::from_wire` in `crates/core/src/chunk/framing.rs` plus reassembly over an adversarial chunk sequence (out-of-order, duplicated, mismatched `ChunkMeta`, oversized indices), against the bounds landed in #601.
3. **Control API body**: the JSON-RPC body decoder feeding `DecodedJsonRpc` in `crates/node/src/native/endpoint/mod.rs`, including the authorization-class classification from #731 (a malformed body must never classify as public).
4. **Handshake**: the SDP / offer / answer decoding used by the manual handshake (`createOffer` / `answerOffer` / `acceptAnswer`, #719) and the `max_message_size` SDP attribute parser from #601.

## Proposal

- A `fuzz/` workspace member built on `cargo-fuzz` (libFuzzer) with one target per boundary above, plus a minimal seed corpus checked in for each (a few valid payloads, so coverage starts past the base58 gate rather than in it).
- Targets are structure-aware where the serde layer would otherwise reject everything at byte zero: use `arbitrary` to build a valid `MessagePayload` and then mutate one field, alongside the raw-bytes target.
- CI: a short bounded run per target on every PR (`-max_total_time` in the low tens of seconds) as a smoke gate in `qaci.yml`, and a longer scheduled run that uploads the corpus and any crash artifacts. Crashes are filed as issues, not fixed silently in the fuzz PR.
- Any panic found becomes a typed error at the boundary (`Error::Deserialize` or a new variant), never an `unwrap` removal in isolation; the fuzz target is the regression test.

## Acceptance

- `cargo fuzz list` shows the four targets; each runs clean for the CI time budget.
- Seed corpora are checked in and documented in `fuzz/README.md`.
- SECURITY.md *Control API* and *Chord Routing* sections mention that the decode boundaries are fuzzed, with the target names.

## Not in scope

Fuzzing the cryptographic primitives (k256 / ed25519 / ElGamal are upstream-tested) or the onion circuit cells, which have their own review track.

Related: #670 (CI gates), #601 (chunk bounds), #731 (authorization class), pre-1.0 security work alongside #736 / #740 / #738.

RyanKung:
Closes #741. No version bump.

## What lands

- Replaces checked-in seed corpus with generated decode-boundary tests.
- Keeps decode-boundary checks inside the crates that own each surface: core message/chunk framing, transport SDP parsing, and native JSON-RPC/auth classification.
- Generates randomized boundary inputs at CI/test runtime through `RINGS_DECODE_BOUNDARY_SEED` and `RINGS_DECODE_BOUNDARY_CASES`; the seed is printed so failures can be replayed without committing generated inputs.
- Adds `crates/core/src/base58_check.rs` so Rings guards the upstream short-input base58-check panic before production decode paths use checked payloads.
- Updates QACI plus a scheduled Decode Boundary Smoke workflow to run the generated boundary tests.

## Cleanup

- The PR branch has been rewritten on current `master` into one clean commit.
- The current PR diff contains only the crate-owned generated boundary tests, the safe base58-check wrapper, CI workflow wiring, and matching docs updates.
- No generated input files or separate fuzz workspace remain in the current PR head.

## Validation

Local validation before publication covered formatting, TOML formatting, actionlint, typos, decode-boundary tests for `rings-core`, `rings-transport`, and `rings-node`, targeted clippy, and cargo-deny. Remote QACI and CodeQL passed on the same tree before the history cleanup; after rewriting the branch to remove stale commits, GitHub Actions is rerunning on commit `9da3cb8a`.

### DIFF
--- a/.github/workflows/decode-boundary-smoke.yml
+++ b/.github/workflows/decode-boundary-smoke.yml
@@ -1,0 +1,49 @@
+name: Decode Boundary Smoke
+
+# Longer scheduled run for the crate-owned decode-boundary tests. The tests generate inputs at
+# runtime from the seed printed by this workflow, so the repository never carries generated corpus
+# files or a separate fuzz workspace.
+
+on:
+  schedule:
+    - cron: "0 3 * * 1"
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_TOOLCHAIN: 1.97.0
+
+permissions:
+  contents: read
+
+jobs:
+  decode_boundary_smoke:
+    name: Decode boundary smoke (scheduled)
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@a37ce9120846195fa4ece8f58b268e6043cb2f26 # v3
+
+      - name: Setup rust toolchain
+        run: |
+          rustup toolchain install "$RUST_TOOLCHAIN" --profile minimal
+          rustup default "$RUST_TOOLCHAIN"
+          rustup show
+
+      - name: Cache cargo registry and build artifacts
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+
+      - name: Generate decode-boundary inputs
+        run: |
+          python3 - <<'PY' | tee -a "$GITHUB_ENV"
+          import secrets
+
+          print(f"RINGS_DECODE_BOUNDARY_SEED={secrets.randbits(64)}")
+          print("RINGS_DECODE_BOUNDARY_CASES=4096")
+          PY
+
+      - name: Run generated decode-boundary tests
+        run: |
+          cargo test -p rings-core --features dummy --no-default-features decode_boundary -- --nocapture
+          cargo test -p rings-transport --no-default-features --features dummy decode_boundary -- --nocapture
+          cargo test -p rings-node --no-default-features --features node decode_boundary -- --nocapture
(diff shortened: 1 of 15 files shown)


fuzz/corpus/chunk_framing/split_second
RyanKung, line 1: what hell is this

fuzz/corpus/handshake_sdp/offer_base58
RyanKung, line 1: you should not put this random shit into repo

RyanKung:
the only correct way to implement fuzz test is to generate those random code in CI/TEST, not hardcode this shit into repo

RyanKung:
crates/fuzz should be remove

PR #754, opened 2026-09-11
Reject oversized ElGamal AEAD wrapped keys

### LINKED ISSUE #749: Onion cells undercharge an unbounded ElGamal wrapped-key vector before decryption
Parent: #743
Severity: High
Review basis: static read-only review of c9d57b581cf1a369327512f2c84fc5ca24a0f482; no local tests/builds were run.

## Finding

The onion wire decoder bounds the total serialized cell and checks only the AEAD ciphertext length for the selected size bucket, but leaves the ElGamal wrapped-key vector unconstrained. The crypto admission gate then charges only the bucket plaintext size before decrypt_aead decrypts every wrapped-key block and only later rejects key material whose length is not exactly 32 bytes.

Evidence:
- decode_wire_message caps the serialized cell at max bucket plus overhead and validates only sealed.ciphertext.len(): https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/circuit/codec.rs#L78-L109
- AeadCiphertext.encrypted_key is a Vec<CiphertextBlock>; one honest 32-byte DEM key is encrypted by encrypt_bytes_with_rng: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs#L61-L70 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs#L97-L107 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs#L401-L427
- decrypt_aead calls decrypt_bytes(&sealed.encrypted_key, ...) before derive_aead_key checks the 32-byte key length: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs#L430-L476
- decrypt_bytes converts/decrypts every block in the vector: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/core/src/ecc/elgamal/impls/secp256k1/mod.rs#L384-L394
- circuit admission charges bucket.plaintext_len() rather than serialized bytes or wrapped-key blocks: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/circuit/shell.rs#L81-L107; the configured limits are byte/op-window based: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/circuit/mod.rs#L88-L93

## Concrete argument

An authenticated immediate peer can send a KiB4 bucket with a 4112-byte AEAD ciphertext and fill the remaining serialized budget with syntactically valid repeated secp256k1 compressed point pairs in encrypted_key. The decoder accepts the cell and the admission gate bills it as one 4 KiB cell, but decryption performs curve conversion plus ElGamal work across roughly 12 MiB / 66 bytes, about 190k wrapped-key blocks, before key-length validation fails.

That violates the limiter model: the visible bucket no longer upper-bounds CPU work.

## Fix direction

Reject wrapped keys whose serialized/block length is not exactly the length required for one 32-byte AEAD key before curve conversion/decryption. Prefer making this a decoded type invariant instead of a late runtime check. Also charge the actual serialized wire bytes, or prove the wrapped-key length is fixed before admission. Add a regression that feeds a small bucket plus an oversized but syntactically valid wrapped-key vector and asserts no per-block decrypt work happens.

RyanKung:
## Summary

- enforce the fixed two-block ElGamal wrapping of the 32-byte AEAD key during deserialization
- repeat the invariant check in `decrypt_aead` before curve conversion for in-memory values
- add core and onion decode regressions for oversized wrapped-key vectors

## Security invariant

Every decoded `AeadCiphertext` now has exactly `ceil(32 / 29) = 2` wrapped-key blocks. Therefore an admitted onion cell cannot hide unbounded ElGamal work behind a small visible bucket.

## Validation

- strict Clippy for all `rings-core` targets/features and `rings-node` node targets, including panic/indexing lints
- secp256k1 adapter tests: 18 passed, 1 ignored
- onion-circuit tests: 45 passed
- full node library tests: 339 passed, 1 ignored
- full core library run: 590 passed, 1 ignored; two unrelated WebRTC topology tests hit their 5-second timeout under parallel load, and both exact serial reruns passed

Fixes #749

### DIFF
--- a/crates/core/src/ecc/elgamal/impls/secp256k1/test_secp256k1.rs
+++ b/crates/core/src/ecc/elgamal/impls/secp256k1/test_secp256k1.rs
@@ -331,8 +331,48 @@ fn test_aead_rejects_empty_wrapped_key() {
 
     assert!(matches!(
         decrypt_aead(&sealed, b"aad", &key),
-        Err(Error::MessageDecryptionFailed(_))
+        Err(Error::AeadWrappedKeyBlockCount {
+            expected: AEAD_WRAPPED_KEY_BLOCKS,
+            actual: 0,
+        })
     ));
+}
+
+#[test]
+fn test_aead_rejects_oversized_wrapped_key_before_curve_conversion() {
+    let key =
+        SecretKey::try_from("65860affb4b570dba06db294aa7c676f68e04a5bf2721243ad3cbc05a79c68c0")
+            .unwrap();
+    let invalid_point = PublicKey([0u8; 33]);
+    let actual = AEAD_WRAPPED_KEY_BLOCKS + 1;
+    let sealed = AeadCiphertext {
+        version: AEAD_VERSION,
+        encrypted_key: vec![(invalid_point, invalid_point); actual],
+        nonce: [0u8; AEAD_NONCE_LEN],
+        ciphertext: Vec::new(),
+    };
+
+    assert!(matches!(
+        decrypt_aead(&sealed, b"aad", &key),
+        Err(Error::AeadWrappedKeyBlockCount {
+            expected: AEAD_WRAPPED_KEY_BLOCKS,
+            actual: observed,
+        }) if observed == actual
+    ));
+}
+
+#[test]
+fn test_aead_decode_rejects_oversized_wrapped_key() {
+    let invalid_point = PublicKey([0u8; 33]);
+    let sealed = AeadCiphertext {
+        version: AEAD_VERSION,
+        encrypted_key: vec![(invalid_point, invalid_point); AEAD_WRAPPED_KEY_BLOCKS + 1],
+        nonce: [0u8; AEAD_NONCE_LEN],
+        ciphertext: Vec::new(),
+    };
+    let encoded = rings_codec::serialize(&sealed).expect("encode invalid envelope fixture");
+
+    assert!(rings_codec::deserialize::<AeadCiphertext>(&encoded).is_err());
 }
 
 #[test]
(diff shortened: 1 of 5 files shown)


crates/core/src/ecc/elgamal/impls/secp256k1/test_secp256k1.rs
```diff
@@ -331,10 +331,50 @@ fn test_aead_rejects_empty_wrapped_key() {
...
+            expected: AEAD_WRAPPED_KEY_BLOCKS,
+            actual: 0,
+        })
     ));
 }
 
+#[test]
+fn test_aead_rejects_oversized_wrapped_key_before_curve_conversion() {
+    let key =
+        SecretKey::try_from("65860affb4b570dba06db294aa7c676f68e04a5bf2721243ad3cbc05a79c68c0")
```
RyanKung, line 344: where is the secret key comefrom?

PR #755, opened 2026-09-11
Epoch-bind onion exits across process restarts

### LINKED ISSUE #750: Persist or epoch-bind onion replay witnesses across exit restarts
Parent: #743
Severity: High
Review basis: static read-only review of c9d57b581cf1a369327512f2c84fc5ca24a0f482; no local tests/builds were run.

## Finding

The replay invariant is process-local, while the exit decryption key is persistent by default. After an exit process restarts with the same session key, already-consumed forward nonces are forgotten but still-valid old cells can still decrypt.

Evidence:
- the replay cache and replay partitions are in-memory HashMaps with empty defaults: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/replay.rs#L110-L150 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/replay.rs#L210-L288
- the circuit law says replay witnesses live for the same interval as still-valid forward layers; the max validity is 150 seconds: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/circuit/mod.rs#L78-L87
- TCP exit consumes the nonce before opening the external target, but the witness lives only in OnionTcpRuntime: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/tcp/mod.rs#L267-L305 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/tcp/mod.rs#L396-L425 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/tcp/mod.rs#L484-L501 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/tcp/mod.rs#L522-L560
- HTTPS exit has the same process-local replay state and consumes before the outbound fetch: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/https/mod.rs#L157-L204 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/onion/https/mod.rs#L616-L648
- native config reads a persisted/raw session_sk, and the CLI default writes ~/.rings/session_sk with a 30-day TTL before passing it into the onion runtime: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/native/config.rs#L156-L168 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/src/native/config.rs#L259-L280 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/bin/rings.rs#L478-L528 and https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/crates/node/bin/rings.rs#L814-L874
- SECURITY.md describes the intended fail-stop model as restart with a new session, which the native default does not enforce: https://github.com/RingsNetwork/rings/blob/c9d57b581cf1a369327512f2c84fc5ca24a0f482/SECURITY.md#L39-L42

## Concrete argument

A malicious or colluding immediate predecessor records a valid first-hop-to-exit TCP Open or HTTPS POST onion cell. The exit process restarts inside the 150-second forward validity window and reloads the same session key from disk. The attacker then wraps and sends the same onion cell bytes inside a fresh authenticated overlay message. Decryption succeeds, the empty replay cache returns Consumed, and the exit opens the TCP target or repeats the HTTPS request. The response path may fail after restart, but the external side effect has already happened.

## Fix direction

Either persist the one-shot witness set durably until the maximum forward validity window expires, or bind a boot/process epoch into advertised exit descriptors and the forward AEAD transcript so old cells cannot decrypt after restart. Add a regression that constructs two runtimes with the same SessionSk, replays the same valid first forward cell after the first runtime consumes it, and expects a replay rejection or epoch failure before any exit side effect.

zkjoie:

Security bug fix for onion-exit replay admission across process restarts.

Forward replay witnesses are process-local, while a delegated `SessionSk` can persist across restarts. Restarting an exit therefore clears its replay cache and can allow the same still-valid authenticated cell to reach the exit adapter again.

- Generate a fresh random `OnionExitEpoch` for every processor lifetime.
- Sign that epoch into onion-exit descriptors and authenticate it inside the final forward layer.
- Reject an epoch mismatch in the pure reducer before emitting any exit-adapter effect.
- Carry the epoch through native and browser registration/RPC boundaries.
- Keep epoch-bearing exit registration crate-internal so advertisement and admission cannot be configured with unrelated lifecycle values.
- Add a regression test that reuses the same `SessionSk` and ciphertext with an empty replay cache after restart, then proves that no exit effect is emitted.
- Document the fail-stop restart contract in `SECURITY.md`.

Yes, at the pre-v1.0 onion boundary: the exit descriptor schema moves from v2 to v3, the exit descriptor RPC shape gains `process_epoch`, and `OnionExitRegistration` is no longer exported as a public custom-registration task because it cannot safely choose an epoch independently of circuit admission. Stale descriptors fail closed until refresh. The onion AEAD namespace remains v1; no backward-compatibility guarantee is made before v1.0.

Validated locally with:

- `cargo test -p rings-node --all-targets --all-features`
- `cargo check -p rings-node --all-targets --all-features`
- strict native and wasm/browser Clippy profiles
- `RUSTDOCFLAGS=-Dwarnings cargo doc --all --no-deps`
- `git diff --check`

Closes #750


crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
RyanKung, line 1: I found that almost all tests in this file are mark with `rings_native`, why not just make the mark at `mod.rs`

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
```diff
@@ -88,6 +97,10 @@ pub(super) fn test_payload(label: &str) -> OnionCircuitPayload {
     )
 }
 
+fn process_epoch() -> OnionExitEpoch {
```
RyanKung: Why 17;16, need comment here

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
```diff
@@ -998,6 +1035,72 @@ async fn test_two_relays_peel_fixed_size_cells_through_the_exit_reducer_and_shel
...
 }
 
+#[cfg(rings_native)]
+#[tokio::test]
+async fn test_restarted_exit_rejects_old_epoch_before_replay_state_or_side_effect() {
+    let client = session();
+    let exit = session();
+    let route = route(&[], &exit);
+    let original_epoch = route.exit().process_epoch;
+    let circuit_id = OnionCircuitId::new([41; 16]);
```
RyanKung: Same reason, need comment, why 41;16

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
```diff
@@ -998,6 +1035,72 @@ async fn test_two_relays_peel_fixed_size_cells_through_the_exit_reducer_and_shel
     assert_eq!(second_transition.state.relay_return_count(), 1);
 }
 
+#[cfg(rings_native)]
+#[tokio::test]
+async fn test_restarted_exit_rejects_old_epoch_before_replay_state_or_side_effect() {
+    let client = session();
```
RyanKung, line 1010: need a comment about how this test works and why

crates/node/src/onion/circuit/crypto.rs
```diff
@@ -203,17 +205,26 @@ pub async fn send_backward(
 fn build_forward_layers(
     client: OnionClientReturn,
     hops: &[OnionRouteHop],
+    process_epoch: crate::onion::OnionExitEpoch,
```
RyanKung: `use crate::onion::OnionExitEpoch` , import first, never use crate::type in a fn sig.

crates/node/src/onion/circuit/mod.rs
```diff
@@ -91,7 +91,7 @@ pub(super) const MAX_ONION_CRYPTO_OPS_GLOBAL_PER_WINDOW: u32 = 8192;
 pub(super) const MAX_ONION_CRYPTO_BYTES_PER_WINDOW: u64 = 256 * 1024 * 1024;
 pub(super) const MAX_ONION_CRYPTO_BYTES_GLOBAL_PER_WINDOW: u64 = 512 * 1024 * 1024;
 pub(super) const MAX_ONION_CRYPTO_PEERS: usize = 64;
-pub(super) const ONION_AEAD_NAMESPACE: &str = "rings-node:onion-circuit:v1";
+pub(super) const ONION_AEAD_NAMESPACE: &str = "rings-node:onion-circuit:v2";
```
RyanKung: We don’t guarantee backward compatibility before v1.0.

crates/node/src/onion/route/tests/test_route_selection.rs
```diff
@@ -56,6 +56,7 @@ fn signed_exit_for_session_network_at(
                 .account_verification_pubkey()
                 .map_err(Error::CoreError)?,
             session_public_key: session_sk.session_public_key(),
+            process_epoch: crate::onion::OnionExitEpoch::new([29; 16]),
```
RyanKung: need fix, same reason

crates/node/src/onion/tests/test_exit_registry.rs
```diff
@@ -201,6 +202,17 @@ fn test_exit_descriptor_signature_covers_schema_version() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_exit_descriptor_signature_covers_process_epoch() -> Result<()> {
```
RyanKung, line 213: need description

RyanKung (changes requested):

crates/node/src/onion/circuit/crypto.rs
r2d2-rs: Fixed in b44930ee. Imported `OnionExitEpoch` and use the short type in both function signatures.

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
r2d2-rs: Fixed in b44930ee. Replaced the literal with documented `RESTART_REPLAY_CIRCUIT_ID`, which keeps this replay witness distinct from neighboring fixtures.

crates/node/src/onion/tests/test_exit_registry.rs
r2d2-rs: Fixed in b44930ee. Added a description of the signature-substitution property the test witnesses.

crates/node/src/onion/route/tests/test_route_selection.rs
r2d2-rs: Fixed in b44930ee. Imported `OnionExitEpoch` and replaced `[29; 16]` with documented `TEST_EXIT_PROCESS_EPOCH`.

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
r2d2-rs: Fixed in b44930ee. Replaced the literal with documented `TEST_PROCESS_EPOCH`; the concrete non-zero bytes only make the fixture deterministic.

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
r2d2-rs: Fixed in b44930ee. Added a test contract explaining the same `SessionSk` plus empty replay cache plus new process epoch, and why rejection must happen before an exit effect.

crates/node/src/onion/circuit/mod.rs
r2d2-rs: Fixed in b44930ee. Restored the v1 AEAD namespace. The per-process epoch remains the restart replay boundary.

crates/node/src/onion/circuit/tests/test_circuit_protocol.rs
r2d2-rs: Fixed in b44930ee. Moved the native cfg to the circuit test module boundary and removed the redundant item-level gates.

crates/node/src/onion/mod.rs
```diff
@@ -772,11 +798,30 @@ impl OnionExitRegistration {
...
         services: Vec<OnionExitService>,
         policy: OnionExitPolicy,
+    ) -> Self {
+        Self::with_process_epoch(
+            heartbeat_interval,
+            ttl,
+            node_type,
+            services,
+            policy,
+            OnionExitEpoch::random(),
```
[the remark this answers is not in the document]
r2d2-rs: Fixed in 85355739. `OnionExitRegistration` and its constructor are now crate-internal, the unsafe prelude export is removed, and `ProcessorBuilder` remains the sole constructor path that injects the same epoch into descriptor publication and circuit admission.

RyanKung (approved):
LGTM

PR #764, opened 2026-09-13
Add provisional Probe service receipts

### LINKED ISSUE #710: DRanking: versioned service receipts and a provisional probe vertical slice

zkjoie:

- [x] Tests for the changes have been added
- [x] Docs have been added or updated

## What kind of change does this PR introduce?

Protocol feature, bounded durable evidence persistence, executable formal models, and documentation.

## What is the current behavior?

Probe liveness uses timestamp-only `PeerLiveness` messages and does not produce portable, jointly signed service evidence. See #710.

## What is the new behavior?

- Defines canonical `ProvisionalServiceClaim` and `ProvisionalServiceReceipt` types with distinct provider and beneficiary signing domains.
- Replaces the legacy liveness wire with the three-step `ProbeRequest`, `ProbeOffer`, and `ProbeAcknowledgement` flow.
- Separates pure receipt transition planning from clock, sequence, pending-state, send, and persistence effects through an explicit effect interpreter.
- Verifies the embedded transactions and all four delegated-session boundaries at one explicit observation instant.
- Adds bounded receipt storage plus independently bounded durable replay markers. Receipt eviction cannot reopen an active freshness key; marker saturation and clock regression fail closed.
- Commits evidence and its replay marker before admission returns success. Storage failure rolls the transition back, and successful admission survives process restart without an explicit flush.
- Requires browser IndexedDB evidence storage to open successfully; constructors without a durable evidence backend keep liveness but cannot return a successful receipt admission.
- Keeps provisional evidence isolated from `CreditRecord` and route ordering.
- Adds executable Stateright models for time, loss, duplication, reordering, replay, independent session expiry, admission outcomes, pair/global/per-beneficiary bounds, eviction, clock regression, restart normalization, and hard-crash recovery. Production-refinement tests replay the model through real signatures and the real evidence store.

## Breaking change

Yes. This is a deliberate hard cutover: legacy `PeerLiveness` wire variants are removed with no fallback, negotiation, or dual decoder. Mixed-version peers are unsupported. The workspace and npm package versions are bumped to `0.25.0`.

## Validation

- `rings-measure`: 48 passed
- `rings-core` all features: 811 passed, 4 ignored; doc tests 1 passed, 2 ignored
- `rings-node` all features: 358 passed, 1 network test ignored; CLI target 7 passed
- Native evidence restart succeeds without an explicit flush; missing durable evidence storage fails closed
- Strict Clippy passes for `rings-measure`, `rings-core`, native `rings-node`, core Wasm, browser lib, and browser tests
- `rings-core` rustdoc passes with warnings denied
- Formatting, patch checks, typos, and locked root/frontend metadata pass

Closes #710


r2d2-rs:
Required follow-up evidence profiles are now tracked separately: #765 (Relay) and #766 (Store). This PR continues to accept only ProbeV1.

crates/core/src/dht/stabilization.rs
```diff
@@ -585,14 +586,31 @@ impl Stabilizer {
...
+        let unix_seconds = u64::try_from(now_ms).unwrap_or(0) / 1_000;
         let candidates = self.transport.liveness_probe_candidates(now_ms)?;
         for attempt in candidates {
             let peer = attempt.peer();
             let state = self
                 .transport
                 .get_connection(peer)
                 .map(|conn| conn.webrtc_connection_state());
-            let msg = Message::PeerLivenessProbe(PeerLivenessProbe { sent_at_ms: now_ms });
+            let mut nonce = [0_u8; 32];
```
RyanKung: use help function instead

crates/core/src/dht/stabilization.rs
```diff
@@ -585,14 +586,31 @@ impl Stabilizer {
...
+            let payload = self
+                .transport
+                .signed_payload(Message::ProbeRequestV1(request), peer, peer)
+                .await?;
+            let tx_id = payload.transaction.tx_id;
+            if !self
+                .transport
+                .register_pending_liveness_probe(attempt, tx_id, request)?
+            {
+                continue;
```
RyanKung: use filter, not `continue`
code in more functional way

crates/core/src/dht/stabilization.rs
```diff
@@ -601,19 +619,23 @@ impl Stabilizer {
                 idle_ms = PEER_LIVENESS_IDLE_MS,
                 "STABILIZATION peer liveness probe send start"
             );
-            match self.transport.send_direct_message(msg, peer).await {
-                Ok(tx_id) => {
-                    self.transport
-                        .record_peer_liveness_probe_sent(attempt, now_ms)?;
+            match self.transport.send_payload(payload).await {
+                Ok(()) => {
+                    let pending = self
```
RyanKung: bad var name. what is pending means? is pending? pending list?

crates/core/src/message/handlers/connection.rs
```diff
@@ -30,34 +28,6 @@ use topology_view::connect_successor_hint;
 #[cfg(all(test, not(target_family = "wasm")))]
 use topology_view::topology_has_confirmed_peer;
 
-/// PeerLivenessProbe is a direct overlay liveness probe.
-#[cfg_attr(all(feature = "wasm", target_family = "wasm"), async_trait(?Send))]
-#[cfg_attr(not(all(feature = "wasm", target_family = "wasm")), async_trait)]
-impl HandleMsg<PeerLivenessProbe> for MessageHandler {
```
RyanKung, line 36: This is a break change, version bump is required

crates/core/src/message/handlers/service_receipt.rs
```diff
@@ -0,0 +1,151 @@
...
+use crate::message::ProvisionalServiceClaimV1;
+use crate::message::ProvisionalServiceReceiptV1;
+use crate::message::ServiceReceiptError;
+use crate::message::Transaction;
+use crate::utils::get_epoch_ms;
+
+/// ProbeV1 request is a direct authenticated overlay liveness probe.
+#[cfg_attr(all(feature = "wasm", target_family = "wasm"), async_trait(?Send))]
+#[cfg_attr(not(all(feature = "wasm", target_family = "wasm")), async_trait)]
+impl HandleMsg<ProbeRequestV1> for MessageHandler {
```
RyanKung: is `V1` label necessory? if we have v2, then we can flag it as v2, but why we now have `v1`

crates/core/src/message/handlers/service_receipt.rs
```diff
@@ -0,0 +1,151 @@
...
+    async fn handle(&self, ctx: &MessagePayload, msg: &ProbeRequestV1) -> Result<()> {
+        if ctx.should_forward_from(self.dht.did) {
+            return self
+                .run_effects([CoreEffect::forward_payload(ctx, None)])
+                .await;
+        }
+
+        let now_seconds = u64::try_from(get_epoch_ms() / 1_000)
+            .map_err(|_| ServiceReceiptError::ObservationTimeOverflow)?;
+        if !msg.epoch.is_accepted_at(now_seconds) {
```
RyanKung: use match { foo => cond_a(), bar=> cond_b() } ,do not do early return

crates/core/src/message/handlers/service_receipt.rs
```diff
@@ -0,0 +1,151 @@
...
+            .into());
+        }
+        let beneficiary = ctx.transaction.origin();
+        let provider = self.dht.did;
+        let network_id = self.transport.network_id;
+        let request_digest = ctx.transaction.digest()?.into_bytes();
+        let completion_sequence = *self
+            .transport
+            .reserve_transaction_sequences(beneficiary, NonZeroU64::MIN)
+            .await?
```
RyanKung: is this a side-effect? do not mix side-effect and pure func.
abstract sideEffect with monad

crates/core/src/message/handlers/service_receipt.rs
```diff
@@ -0,0 +1,151 @@
...
+            ctx,
+            self.transport.network_id,
+            beneficiary,
+            get_epoch_ms(),
+        )?;
+        if !self
+            .transport
+            .consume_pending_probe(provider, ctx.transaction.tx_id, request)?
+        {
+            return Err(Error::InvalidMessage(
```
RyanKung: you should map it to error, but dont do early return

crates/core/src/message/service_receipt/model.rs
```diff
@@ -0,0 +1,503 @@
+//! Executable finite-state model for the provisional ProbeV1 receipt flow.
```
RyanKung: what is this mod for， if it's a formal model it should be ignored when we compile product code

RyanKung (changes requested):

crates/core/src/dht/stabilization.rs
r2d2-rs: Fixed in `126156f4`: probe nonce and epoch construction now goes through `ProbeRequest::random_for_epoch` instead of open-coded random-byte setup in stabilization.

crates/core/src/dht/stabilization.rs
r2d2-rs: Fixed in `126156f4`: candidates now flow through sequential `then` -> `try_filter_map` -> `try_for_each`; superseded generations are filtered without `continue`.

crates/core/src/dht/stabilization.rs
r2d2-rs: Fixed in `126156f4`: renamed the ambiguous `pending` binding/log field to `matching_probe_recorded`, which states exactly what the boolean means.

crates/core/src/message/handlers/connection.rs
r2d2-rs: Fixed in `126156f4`: this breaking wire cutover now bumps the Rust workspace, internal path constraints, npm package, root lockfile, and frontend lockfile to `0.25.0`; the changelog records the coordinated-upgrade requirement.

crates/core/src/message/handlers/service_receipt.rs
r2d2-rs: Fixed in `126156f4`: removed the `V1` suffix from all public Rust receipt types and `Message` variants. `V1` remains only inside canonical byte markers and signing domains as a cryptographic domain separator.

crates/core/src/message/handlers/service_receipt.rs
r2d2-rs: Fixed in `126156f4`: forwarding/local dispatch and epoch admission are expressed as `match` transitions, with no handler early return.

crates/core/src/message/handlers/service_receipt.rs
r2d2-rs: Fixed in `126156f4`: pure plan/build/role-validation functions are separated from an explicit `ProbeEffect` algebra and `ProbeEffectInterpreter`. Clock reads, sequence reservation, pending-state mutation, network sends, and durable admission now live only in the interpreter boundary.

crates/core/src/message/handlers/service_receipt.rs
r2d2-rs: Fixed in `126156f4`: the pending-probe boolean is converted with `then_some(()).ok_or_else(...)`; the handler no longer uses an early return for this rejection.

crates/core/src/message/service_receipt/model.rs
r2d2-rs: Confirmed and made explicit in `126156f4`: the parent module already used `#[cfg(all(test, not(target_family = "wasm")))]`; the model file now also has the same file-level gate. `stateright` is absent from the normal `rings-core` dependency graph.

RyanKung (approved):

PR #770, opened 2026-09-14
Bound finger-table convergence by proved ranges

### LINKED ISSUE #768: Avoid linear 160-round finger-table convergence
### Motivation

Finger-table convergence is currently coupled to the full stabilization cadence and advances by exactly one slot per round.

`finger_join` may populate a contiguous set of slots with a newly admitted peer, but those `Some(peer)` values are only best-known hints. The state does not distinguish an inferred slot from a slot verified by a `find_successor(local + 2^i)` result. `step_fix_finger` nevertheless increments `fix_finger_index` by one, and `finger_set` updates only that selected slot.

As a result, a complete validation sweep takes:

```
RING_BITS * stabilize_interval
```

With 160 slots and the default 15-second interval, the worst-case sweep takes about 40 minutes, regardless of how many distinct finger destinations exist.

Prioritizing only empty slots is insufficient. For example, a bootstrap peer may have already filled all 160 slots, while closer peers discovered later should replace some lower slots. Every slot is non-empty, but the table is still inferred and potentially stale.

Reducing `stabilize_interval` to one second is also the wrong control. That interval drives the full maintenance cycle, including predecessor notifications, finger repair, topology queries, cleanup, and periodic storage repair. It increases steady-state control traffic and can synchronize bursts across a fleet without fixing the missing evidence model.

### Current Behavior

The relevant paths are:

- `finger_join` fills multiple slots from an admitted peer using local distance.
- `step_fix_finger` selects `(fix_finger_index + 1) % fingers.len()`.
- `finger_set` writes one slot.
- `Option<Did>` represents both inferred and lookup-verified values.

The liveness expectation therefore depends on 160 fair stabilization rounds even for a sparse ring whose final finger table contains only a few distinct peer ranges.

A concrete five-node browser session demonstrated this behavior. Connecting to the public seed inferred slots 0 through 158 as the seed. The only other distinct target was at slot 159, but the round-robin cursor had to revisit the preceding slots before reaching it.

### Proposed Direction

Make bootstrap and post-change finger convergence range-aware and bounded independently from the steady-state stabilization period.

Rings defines finger slot `i` as the nearest forward node whose distance from the local DID is at least `2^i`. If a verified lookup for slot `i` returns peer `p` at distance `d = dist(local, p)`, then `p` is also the correct result for every consecutive slot `j >= i` for which `2^j <= d`. No peer can lie between `2^i` and `d`, otherwise `p` would not have been the verified successor for slot `i`.

A convergence pass can therefore:

1. Query the lowest unverified slot `i`.
2. Apply the returned peer to the entire proven consecutive range.
3. Advance the cursor to the first slot whose threshold is greater than the returned peer distance.
4. Repeat until the table is covered.
5. Rate-limit and jitter bootstrap work independently from the full stabilization loop.

The state must preserve the difference between inferred hints and verified evidence, for example with a verification epoch or an equivalent range-level representation. Join, admission, removal, and successor changes should invalidate only the affected evidence ranges.

This should remain compatible with the existing sparse/no-wrap finger definition. The exact representation is open; the important property is that convergence advances by proven ranges rather than by slot occupancy or an unconditional one-slot cursor.

### Acceptance Criteria

- Finger convergence is not gated on 160 full stabilization rounds.
- Non-empty inferred slots are not treated as verified merely because they contain a DID.
- A verified lookup result can update every consecutive slot justified by the finger-distance invariant.
- A sparse ring converges in a number of lookups proportional to its distinct finger ranges, not the fixed 160-slot width.
- The five-node case above converges in approximately two range lookups rather than waiting for slot 159 after 159 timer ticks.
- A case where one bootstrap peer initially fills all slots still discovers and installs closer peers.
- Topology admission, removal, and successor changes trigger bounded revalidation of affected ranges.
- Bootstrap convergence has an explicit per-node rate bound and jitter; it does not require lowering the full `stabilize_interval`.
- Property tests compare the converged result with `topology::finger_table(all, local)` across sparse and dense rings, including insertions and removals.
- Deterministic schedule tests cover concurrent topology changes while a range lookup is in flight, so stale results cannot overwrite newer evidence.

### Related

This is separate from #763. That issue owns run-scoped bootstrap reachability and automatic redial after transport loss. This issue concerns how quickly the Chord finger table converges once the node is connected to the overlay.

zkjoie:

## :large_blue_circle: What kind of change does this PR introduce?

Protocol/DHT convergence fix, bounded topology-report effects, browser lifecycle hardening, documentation, and a version bump to 0.26.0.

## :brown_circle: What is the current behavior?

Finger maintenance advances one slot per stabilization round, so a 160-slot table can take roughly 160 rounds. The previous evidence model could also retain false verification across isolation/join, reuse numeric request IDs after restart, and lose a timely proof during a longer connection handshake.

The old network and formal gates did not cover the asynchronous effect path: they modeled topology state but not report ownership, connection-admission leases, per-candidate effects, or overlapping browser listener generations. The network test multiplied a fixed four-leg estimate rather than executing the actual transport cascade.

Closes #768.

## :green_circle: What is the new behavior?

- Separates inferred finger hints from verified ranges and lets one proof verify every consecutive slot covered by its successor.
- Keeps an isolated node dormant and unverified. The first successor admission activates convergence; losing the final successor invalidates all membership evidence.
- Proves the local successor interval without a ring traversal, using only a correlated stabilization report from the current successor whose predecessor is the local node.
- Uses fresh UUID request tokens for finger and topology reports. Unsolicited, duplicate, superseded, replayed, and post-churn reports are rejected.
- Preserves a timely finger proof before awaiting transport admission. Lookup reports expire after 10 seconds; retained admission proofs have a separate 210-second lease, strictly longer than the 180-second handshake generation that owns an attached proof, so the lease is only a backstop for a proof that never gains a generation.
- Counts only actual timeout, invalid report, cancellation, send/handshake failure, or admission-lease expiry as failure. Retries use 2/4/8/16/32/60-second floors plus full jitter; topology churn itself does not increase the failure streak.
- Keeps at most one finger operation outstanding per node and at least one second between emissions. Lookups are issued for unverified ranges in cyclic order, resuming after the range the previous attempt proved or failed, so one slot whose successor never completes a handshake costs one attempt per rotation instead of starving every range above it; periodic revalidation is deferred by pending work only while that work is succeeding. Initial work is spread over a lifecycle-randomized 10-second window, drawn only on activation, with no catch-up burst after browser suspension.
- Correlates and bounds topology-report effects. Successor sync admits at most K deduplicated candidates; stabilization admits at most K+1 including the predecessor; both order the bounded set by transport quality. Both revalidate the production token before each candidate. A claim is an RAII guard released on every exit path of its handler, a new maintenance round never supersedes a report that is still being processed, and a claimed report keeps its remaining budget while its reporter stays a successor (its own admissions extend the list without revoking it). Once the reporter leaves, at most the already-issued candidate effect finishes; no further candidate is issued.
- Serializes WASM listener generations across the full cooperative shutdown. Repeated stop/listen cycles cannot run duplicate maintenance daemons and reuse the same lifecycle phase.
- Restores the public `Chord::fix_fingers()` begin-and-advance contract.
- Ignores the complete `.claude/` directory.

## :radioactive: Does this PR introduce a breaking change?

Yes. `FingerFixRequest` now carries a mandatory UUID, and `QueryForTopoInfoSend` / `QueryForTopoInfoReport` carry a mandatory UUID request ID. There is no legacy decoder or fallback. Nodes in one overlay must upgrade together to 0.26.0. `CorrectChord::stabilize` is removed: a successor's report changes the successor list only through the token-checked report path.

## :information_source: Why the previous gates missed the failures

- The retry model was a hand-written shadow state machine and the scope manifest only classified events; neither guaranteed that the production transition or asynchronous effect existed in the model.
- The model omitted the two-phase reserve/execute boundary, pending connection ownership, the 180-second admission lease, and listener restart generations.
- Happy-path fixtures joined before their first convergence pass, so they missed isolation-before-join, last-successor loss, restart replay, clock-domain changes, and repeated browser stop/listen.
- The network test assumed four routed legs and had a tautological sample/lower-bound assertion. It never executed `TopologyAction -> CoreEffect -> transport`, WebRTC admission, successor queries, or their follow-up connections.

## Replacement gates and network-cost result

- Finger and stabilization checks execute the production `topology::step`; stabilization and successor-sync effect checks execute the same bounded production plans used by the handlers.
- The model separates reserve from execute. It deliberately allows one already-issued candidate effect after churn and proves that the stale plan cannot reserve another.
- A three-node dummy-transport test drains the full missing-finger admission cascade to quiescence and caps the observed causal trace at 20 outbound control submissions. A five-node local-successor proof emits zero finger submissions.
- A 200-node deterministic schedule test checks deadline dispersion. A real Chrome `wasm_bindgen_test` checks that an overlapping listener cannot start until the previous generation releases the lifecycle gate.
- Per report, successor sync is bounded by K candidates and stabilization by K+1. Pending handshakes are separately capacity-limited, and every routed carrier retains its 64-hop budget.

These are per-node and per-scenario bounds, not an N-independent global rate limit. Healthy aggregate startup work remains O(N), and the 10-second phase window smooths a synchronized start but does not provide a hard global seed-QPS guarantee.

## Local validation on `5de67b48`

- Full core dummy suite: 867 passed, 4 ignored, 0 failed.
- Default and dummy core Clippy with `-D warnings`.
- Strict core WASM, browser-node WASM, and native-node Clippy, including panic/unwrap/expect/indexing lints.
- Finger/stabilization and successor-sync production-transition model tests.
- Full transport cascade, five-node zero-submission, correlated successor-sync, and register-before-send/send-failure rollback tests.
- Real headless Chrome listener stop/listen test: 1 passed.
- Cargo cooldown passes with a reviewed exact exception for `rustls 0.23.45`; the 2880-minute global window is unchanged.
- `cargo deny --locked check -A unmatched-source advisories licenses bans sources` passes.
- `cargo check --locked --workspace --all-targets` and all 70 `rings-transport` tests pass.
- Root, dweb, and frontend Cargo metadata with `--locked`.
- `cargo fmt` and `git diff --check`.

Remote CI was intentionally not inspected during this self-audit.


r2d2-rs:
Thanks for the full-diff review. I treated R1–R7 as runtime findings rather than only updating the proof core. The fixes are now on `be3114ae`.

### Review response

- **R1 fixed — no local-range ring traversal.** The current successor's topology report is accepted only when it matches the exact outstanding stabilization UUID and reports `predecessor == local`. That authenticated current-head evidence verifies the local successor interval directly. The five-node dummy-transport regression observes zero finger submissions for that interval.
- **R2 fixed — timely proof survives the handshake.** A valid missing-candidate report closes the 10-second lookup immediately and moves the proof into a distinct 180-second admission lease before awaiting WebRTC work. Dropped or stuck handlers cannot retain it forever; lease expiry releases it and records one failed attempt.
- **R3 fixed — churn is not failure.** Hint/topology invalidation cancels stale evidence without incrementing the failure streak. Timeout, invalid report, cancellation, send/handshake failure, and admission-lease expiry do increment it; only applied evidence resets it.
- **R4 fixed — deadline-based wakeup.** The scheduler uses the exact remaining lookup/admission duration rather than a 1–2 second poll. Remaining durations cross listener clock origins safely, including repeated stop/listen.
- **R5 addressed.** Disposition conversion is centralized, hint mutation and topology outcomes share the production transition, connection-effect cursors are production plans reused by tests/models, and oversized modules/tests were split below 1000 lines. I did not force stabilize/repair/finger into one generic scheduler phase because their reservation and completion semantics differ.
- **R6 fixed — no synthetic four-leg model or lower bound.** The gate executes the actual effect/interpreter/transport path and drains the missing-peer admission cascade. The controlled three-node trace is capped at 20 outbound control submissions; the test no longer fails when an optimization reduces traffic.
- **R7 retained intentionally.** Ignoring the full `.claude/` directory was an explicit follow-up requirement for this PR, so it remains in scope.

### Additional audit fixes

The self-review found and fixed several gaps beyond R1–R7:

- Finger and topology reports now use mandatory fresh UUIDs; restart/replay cannot match a new in-flight operation through a reset numeric counter.
- `Chord::fix_fingers()` again performs begin + advance.
- Stabilization and successor-sync reports use exact `Requested -> Processing` single-use claims, bounded/deduplicated candidate plans, and per-candidate token revalidation. Sync is bounded by K candidates; stabilization by K+1. After churn, an already-issued permit may finish one candidate effect for each active plan, but that stale plan cannot issue another.
- Successor queries register before send and cancel the exact token on send failure.
- WASM providers serialize listener generations across complete cooperative cleanup, so rapid stop/listen cannot create duplicate maintenance daemons.
- `FingerTable` equality includes convergence/cursor state.

### Why the old formal/model gates stayed green

The old retry model was a shadow implementation, while its scope manifest only asserted that an event had a label. The network test similarly asserted arithmetic over a fixed constant. Neither was coupled to the production async effect boundary. They omitted pending-attempt ownership, lookup-vs-admission time, reserve-vs-execute interleavings, topology report tokens, and browser listener generations. Happy-path setup also joined before the first convergence pass, excluding the isolation-before-join counterexample.

The replacement gates directly call production `topology::step` and the production connection plans, explicitly separate reserve from execute, run the real dummy transport cascade, and run the overlap test in headless Chrome. The scope manifest now explicitly says it is routing metadata, not a completeness proof.

### Local audit result

- Full core dummy suite: **867 passed, 4 ignored, 0 failed**.
- Default/dummy core, strict core WASM, strict browser-node WASM, and strict native-node Clippy passed.
- Real Chrome stop/listen test passed.
- Root/dweb/frontend locked metadata, fmt, and diff checks passed.
- Three independent final reviews (network cost, protocol/formal, Rust/API) found no P0/P1/P2 blocker on the exact staged snapshot.

Remaining honest coverage boundaries: the real finger transport gates are N=3/N=5 rather than N>=10; successor-sync has production-transition interleaving witnesses rather than a complete BFS over every multi-reporter schedule; existing Miri filters do not cover the new topology paths. The documented network guarantee is therefore per-node/per-scenario. Fleet startup remains O(N), and phase jitter is not a global seed-QPS hard limit.

Per request, I did not inspect remote CI and have not merged the PR.

r2d2-rs:
Fixed the reported `cargo deny` failure in `5de67b48`.

- Updated the root lockfile from `rustls 0.23.29` to `0.23.45`, the first version accepted by `RUSTSEC-2026-0285`.
- Cargo also selected the compatible `rustls-webpki 0.103.15` patch release.
- `rustls 0.23.45` is only about two hours old, so the repository's independent 2880-minute supply-chain cooldown correctly rejected it. I kept that global window unchanged and added a reviewed `allow.exact` entry for only `rustls = 0.23.45`, with the advisory recorded next to it. There is no package-wide or time-window exception.

Local validation on the exact pushed tree:

- `cargo cooldown-check -v` — passed; only the exact security release is skipped.
- `cargo deny --locked check -A unmatched-source advisories licenses bans sources` — advisories, bans, licenses, and sources all passed.
- `cargo check --locked --workspace --all-targets` — passed, including rustls → dtls/webrtc and rustls → hyper-rustls/reqwest dependency paths.
- `cargo test --locked -p rings-transport --all-targets` — 70 passed, 0 failed.
- Locked metadata and `git diff --check` — passed.

I did not inspect or rerun remote CI.

r2d2-rs:
Architectural follow-up after the final self-review:

- Split the pure finger-convergence core into `proof`, `evidence`, `attempt`, `retry`, and scheduler-only `status` modules. The coordinator owns no network, lock, clock, or RNG effects.
- Replaced parallel flags and optional owners with an algebraic `Idle | AwaitingReport | AwaitingAdmission` state, so dual ownership is unrepresentable.
- Moved finger-specific topology integration into `dht/topology/finger.rs`; transport remains responsible only for admission and async effects.
- Documented the exact Chord-derived range lemma and explicitly separated it from Rings-specific UUID correlation, evidence epochs, admission leases, backoff, jitter, and browser lifecycle policy.
- Made `FingerTable` equality cover the complete serialized protocol state and documented the 0.26 behavior change.

Network-focused self-review found and fixed two semantic regressions:

1. A same-token report for a conflicting successor could evict a retained admission proof.
2. The active-but-unroutable path could cancel by token before validating the reported successor.

The latter is now one atomic pure transition: token, successor, deadline, range, and evidence epoch are validated before ownership is retired and backoff is recorded. A conflicting stale report has no state or network effect.

Local validation (no remote CI polling):

- full `rings-core --features dummy`: 869 passed, 0 failed, 4 explicitly ignored
- deterministic/model suites: finger retry model 6/6; topology 50/50; transport finger 14/14; sync-storm 11/11
- strict native and WASM Clippy with warnings, unwrap, expect, panic, todo, unimplemented, and indexing denied
- nightly fmt, rustdoc warnings, cargo-deny advisories/licenses/bans/sources
- three independent final static reviews (Rust/API, protocol/formal, network cost): no remaining P0-P3 findings

Final refactor commit: `b4765c04`.

crates/core/src/dht/chord/tests/test_finger.rs
```diff
@@ -103,3 +103,35 @@ async fn test_finger_table_tracks_clockwise_and_wrapped_joins() -> Result<()> {
     assert_eq!(node_d.successors().list()?, vec![a]);
     Ok(())
 }
+
+#[test]
```
RyanKung, line 116: need description/document here

r2d2-rs:
Final exact-head validation update for `b4765c04c4ee15488ec40fc7620968e062758c2d`:

- `cargo test -p rings-core --features dummy`: **870 passed, 0 failed, 4 explicitly ignored** (215.38s)
- doc tests: **1 passed, 0 failed, 2 ignored**

The extra passing test is the new regression proving that an unroutable conflicting successor cannot retire another successors admission proof. No remote CI was polled.

crates/core/src/dht/chord/peer_ring.rs
RyanKung, line 1: you made peer_Ring very complex, the code is not clean and resonable, and it's not readable .
you should add more document and makesure this functions are readable

crates/core/src/dht/chord/peer_ring.rs
```diff
@@ -316,20 +371,225 @@ impl PeerRing {
...
+    /// A browser provider can stop and restart its listener without rebuilding
+    /// the ring. Keeping this value on the ring prevents every listener restart
+    /// from rerolling its phase while still assigning a new phase after a full
+    /// provider reconstruction.
+    pub(crate) fn finger_jitter_entropy(&self) -> uuid::Uuid {
+        *self.finger_jitter_entropy.get_or_init(new_uuid)
+    }
+
+    pub(crate) fn finger_convergence_status(&self) -> Result<FingerConvergenceStatus> {
+        let now_ms = self.finger_now_ms();
```
RyanKung, line 466: self.with_topology_state(lambda |state| xxxx)?
i cant understand it without document

crates/core/src/dht/chord/peer_ring.rs
RyanKung, line 98: you can use seprated impl to group funcs, 
e.g:
impl PeerRing {fn a, fn b, fnc}
impl PeerRing {fn d, fn e, fn f}

RyanKung (approved):
LGTM

PR #782, opened 2026-09-15
Add run-owned bootstrap reachability supervision (#763)

### LINKED ISSUE #763: Add run-owned bootstrap reachability and automatic redial
### Motivation

`rings run` does not currently own ongoing bootstrap connectivity. Commands such as `rings connect node` and `rings connect seed` establish a connection once, but the running node does not retain a desired bootstrap target and redial it after a later transport failure.

If a WebRTC transport becomes `Disconnected`/`Closed` and is removed from the DHT, a healthy long-running process can remain partitioned indefinitely. We observed this with three 0.23.0 native nodes: all three initially connected to the same public seed, later lost that transport while their processes stayed healthy, and recovered only after the node processes were restarted.

Chord stabilization should continue to manage ordinary neighbor and finger-table edges. The missing behavior is a run-scoped bootstrap reachability supervisor, not a mechanism that pins or redials every peer in `listPeers`.

### Proposed Solution

Allow `rings run` to own a set of managed bootstrap targets for the lifetime of the process. Prefer reusing the existing `SeedPeer`/seed-file descriptor instead of introducing a second peer format. If imperative connect RPCs can register a managed target, that behavior should be explicit so existing one-shot semantics remain available.

For each managed bootstrap target:

1. When the node becomes ready, or when a previously reachable target becomes unreachable, perform a bounded rapid retry burst (for example, five attempts separated by short delays).
2. If the rapid attempts are exhausted, retry once every five minutes until reachability is restored.
3. Reset the retry state after a successful connection so a later loss starts with the rapid retry burst again.
4. Do not start overlapping handshakes for the same target. Multiple targets must have independent retry state.
5. Treat successful routing/reachability through the current Chord overlay as healthy; do not force a direct seed edge merely because the seed is not currently a finger-table peer. When the target is unreachable through the overlay, use its HTTP endpoint to bootstrap again.
6. Keep transient failures inside the supervisor; they must not terminate `rings run`. Invalid configuration and unsafe endpoint validation should still fail fast.
7. Cancel retry tasks cleanly during shutdown. Log state transitions and retry outcomes without exposing remote API tokens.
8. Use a five-minute base interval with bounded jitter if needed to prevent synchronized fleets from redialing at exactly the same instant.

Once bootstrap reachability is restored, normal Chord stabilization should repopulate the appropriate finger lists. `listPeers` should remain a view of the local Chord/transport peers; this issue does not propose a global node registry.

### Acceptance Criteria

- A native node reconnects to a configured bootstrap target after the transport is dropped, without restarting `rings run`.
- The initial rapid retry burst is bounded and cannot become a hot loop.
- After the burst, retries use the five-minute base cadence.
- No retry is attempted while the target remains reachable through the overlay.
- Concurrent timer and disconnect events cannot create duplicate connection attempts.
- Retry state is bounded by the configured bootstrap target list.
- Tests use controlled/paused time to cover initial failure, transition to the slow cadence, recovery, a second disconnect, shutdown, and multiple targets.
- Native-node and CLI documentation describes how bootstrap targets are configured and the automatic redial policy.

### Alternatives

An external watchdog can poll peers and restart the process, but that discards healthy node state and duplicates overlay semantics outside Rings. Blindly reconnecting every disconnected peer would also fight Chord stabilization and create unnecessary connection churn.

### Additional Context

The current `ConnectWithSeed` path in `crates/node/src/rpc_impl.rs` filters already-connected DIDs, runs one set of connection futures, returns the first error, and retains no retry state. `crates/node/src/seed.rs` already provides a reusable bootstrap peer descriptor with redacted token debugging.

zkjoie:
Closes #763.

## What

`rings run` owns a set of **managed bootstrap targets** for the life of the process and keeps them reachable. Targets come from the new `bootstrap.peers` config section (written empty by `rings init`) and from `rings run --bootstrap-seed <file-or-url>`, both in the seed-document shape (`{ did, url, api_token? }`). `rings connect node|seed` keep their one-shot semantics; nothing registers a managed target imperatively.

## How reachability is judged

The issue asks for "reachable through the current Chord overlay", not a direct edge. The probe is a **routed successor lookup** for the target's DID (`FindSuccessorSend { did: t, strict: false }` sent toward `t`): in a Chord ring every present node is the successor of its own identifier, so a `FindSuccessorReport` under the probe's transaction id naming `t` proves `t` is in the overlay. A partition that lost `t` answers with the node now succeeding `t`'s position; an isolated node cannot route the request at all; both read as unreachable. A direct `Connected` transport short-circuits the probe, so a target is never forced into a direct edge merely because it is not a finger; a key in this node's own successor range is refuted locally without sending anything. Reports are correlated by `tx_id` (reports reuse the request's id), so they cannot be spoofed off-path; on-path trust is exactly Chord's. Validated end to end by a real-WebRTC test over a `C — B — A` chain (`test_bootstrap_probe.rs`): present, absent, direct, and locally refuted.

## Departures are a swarm fact (core)

The core emitted `ConnectionStateChange { Failed | Closed }` only for an admitted generation and only after its own `leave_dht`; every locally decided retirement (data-channel close, liveness/stabilization removal, eviction, `Swarm::disconnect`) retired the record first and returned before emitting. `SwarmEvent::PeerRetired { peer }` is now emitted where an admitted record is retired — the retirement transition yields a `#[must_use] Retirement` witness announced by the one async path `retire_announced_if`, with the peer's delivery turn taken before the retirement — and only when that admission was announced as `ConnectionStateChange { Connected }` (the mark is set atomically under the lifecycle boundary), so for every generation `Connected` delivered ⟺ `PeerRetired` delivered. A topology prune that keeps the record emits nothing. The transport is born with the callback slot (`SwarmTransportParts`); the swarm shares it.

## Retry policy

Per target, a pure state machine over explicit time (`schedule.rs`, phase diagram and laws in its module doc):

- unreachable → HTTP redial: `Processor::connect_peer_via_http` owns the pinned-dial policy — refused as `HandshakeFailure::InFlight` (no request sent) while the local core holds an unadmitted handshake to the target, whichever side started it (`Swarm::has_unadmitted_connection`), or when the local core refuses the exchange because the slot is owned by another attempt; otherwise it pins the answering DID *before* any offer is created and runs the exchange against the generation its offer reserved (`Swarm::offer_connection` / `accept_answer_for`); the port then waits for admission (30 s) and abandons the reserved generation on every exit short of an admission, cancelling it iff it is still pending; a refusal or a failure is a *miss* that consumes a burst slot alike, so a handshake the peer keeps in flight cannot hold the target in the rapid cadence; an *announced* admission, ready or recovering, counts as reachable and is left to the swarm; every remote request has a timeout; 5 attempts, each ≥ 2 s after the previous settled, then every 300 s + U[0, 30 s] jitter (seeded, replayable);
- success resets the count; a later loss restarts the burst;
- a `PeerRetired` for a target triggers an immediate reassessment, also while the target waits in the slow cadence (it was admitted meanwhile and left again);
- a loss noticed while a turn is in flight is never lost: whatever the turn's outcome, its settlement restarts the burst at once;
- at most one turn per target in flight; targets retry independently; the endpoint must answer as the configured DID;
- stop aborts in-flight probes and handshakes; only configuration fails fast (unparsable DID, non-public URL, one DID with differing endpoints, self); an entry repeated verbatim is merged.

## Plumbing

- `Swarm::offer_connection` / `accept_answer_for` / `cancel_pending_connection_attempt` — a handshake pinned to the generation it reserved (the answer is refused as `ConnectionAttemptSuperseded` whenever another generation owns the slot, from one `AnswerSlot` read; the cancellation releases a still-pending generation only); `ConnectionAttempt::peer` names the peer; `Swarm::is_peer_admitted` (announced admission, whose retirement is therefore promised) and `has_unadmitted_connection` — the facts a dialer needs, none naming a physical transport state; `Swarm::lookup_successor` — the local step of a successor lookup, routed through the hop that step chose when the topology cannot decide; `SwarmEvent::peer_transition` — both halves of the retirement law as one `PeerTransition` (core). `seed::ValidatedSeedPeer` / `validate_seed_peers` / `SeedPeerError` — one validated seed entry and one seed-set law shared by `connect seed` and the managed targets; an entry naming the node itself is skipped on both paths; `Error::HandshakeInFlight` (819) reports a refused handshake to an RPC caller (node). `RemoteRpcEndpoint` carries the endpoint policy as a type (node).
- `Backend::observed_by(BackendObserver)` — the backend already decodes each inbound payload once; the observer receives application lookup reports (`FindSuccessorReportHandler::None`), `peer_admitted` (from the admission `Connected`) and `peer_retired` (from `PeerRetired`) — logical facts only, the backend is the single translation point. `ReachabilityEvidence` is the node's implementation: three rendezvous records (reports, admissions, losses) read by the probe, the dial and the supervisor; `BootstrapSupervisor::over_processor` returns it together with the supervisor and nothing is installed when there are no targets.
- Bumped to 0.27.0; CHANGELOG, `config.yaml.md`, `cli.md`, and the `rings run --help` dump updated.

## Tests

- `schedule.rs`: laws (bounded burst, reset, no overlap, drops never lost, totality, jitter bounds/determinism, saturation).
- `bootstrap/tests.rs`: shell under tokio's paused clock with a scripted port — initial burst → slow cadence, recovery, second disconnect, transient-disconnect ignored, independent targets, shutdown aborting a hanging dial, drop during a turn; target validation; ledger rendezvous in both orders and bounded early buffer.
- `config.rs`: `rings init` writes the section; old configs load empty; peers round-trip.
- `test_bootstrap_probe.rs`: real-network routed probe (present, verified routed not direct / absent / direct / locally refuted), backend translation of every physical state, and `PeerRetired` observed on a local `Swarm::disconnect`.

Local gates: `clippy --all --tests -D warnings`, nightly fmt, `rustdoc -D warnings`, `taplo`, `mdbook build`, full node lib suite, full core lib suite (its real-WebRTC tests with 5 s waits show the same load-induced timeouts on the base commit; see the comment).


crates/core/src/swarm/mod.rs
```diff
@@ -170,6 +171,19 @@ impl Swarm {
...
+    /// Whether `peer` has an admitted direct transport in the `Connected` state.
+    ///
+    /// Law: `is_peer_connected(p) ⟹ p ∈ peer_dids()`. The converse fails while a routable
+    /// transport is still `Connecting`, so this is the strict readiness predicate rather than
+    /// the routability one.
+    pub fn is_peer_connected(&self, peer: Did) -> bool {
+        self.transport
+            .get_connection(peer)
+            .is_some_and(|connection| {
+                connection.webrtc_connection_state() == WebrtcConnectionState::Connected
```
RyanKung: this should be abstract, how if we add another type of transport future, lick QUICTransport?

crates/node/bin/rings.rs
```diff
@@ -23,6 +23,9 @@ use rings_node::native::api_auth::load_api_token;
 use rings_node::native::api_auth::load_api_token_file;
 use rings_node::native::api_auth::load_or_create_api_token;
 use rings_node::native::api_auth::ApiSecurity;
+use rings_node::native::bootstrap::BootstrapObserver;
```
RyanKung: why it's bootstrapObserver, why it's named such a comfused name

RyanKung (approved):
LGTM

PR #784, opened 2026-09-17
Reference sessions by content address after one inline send per link (#738)

### LINKED ISSUE #738: Every payload ships two full Session delegations; reference sessions by id after a one-time announce
## Observation

Every `MessagePayload` carries two `MessageVerification` values: `transaction.verification` (origin) and the outer `verification` (current hop). Each embeds a full `Session`:

```
Session { session_id: Did, account: Account, ttl_ms, ts_ms, sig: Vec<u8> }  // account delegation, ~150 B
MessageVerification { session: Session, ttl_ms, ts_ms, sig: Vec<u8> }        // + 65 B signature
```

So each payload ships two account-delegation signatures and two session descriptors in addition to the two message signatures, on every hop, for the lifetime of a peer relationship whose session does not change. The origin's `Session` is repeated on every message the origin sends; the hop's `Session` is repeated on every message the hop forwards.

Rough wire cost (postcard, before gzip): about 220 bytes per `MessageVerification`, so roughly 440 bytes of identity material per payload, of which only the two 65-byte message signatures are per-message data.

## Why signature aggregation does not address this

`rings_core::ecc::signers::bls::aggregate` exists and aggregates BLS signatures over distinct messages. It is not applicable here:

- Message signatures are secp256k1 ECDSA from `SessionSk::sign`; ECDSA does not aggregate. BLS is only an `Account` variant, i.e. the delegation signature inside `Session`.
- The relay never accumulates signatures: the hop signature is verified and replaced at each hop, so there are always exactly two, not N.
- Aggregation cannot merge a delegation signature (over the session) with a message signature (over the payload) unless both are BLS and the verifier holds both messages, which they do anyway.

Aggregation is a fit for a future batch of receipts signed by many peers (DRanking V2, #710-#714), not for the relay.

## Proposal

Send the `Session` once per peer relationship and reference it by `session_id` afterwards:

- A `SessionAnnounce` message (or handshake field) carries the full `Session`; the receiver caches it under `(account, session_id)` bounded by the session `ttl_ms`.
- `MessageVerification` becomes `{ session_id: Did, ttl_ms, ts_ms, sig }` plus an `Option<Session>` that is populated only when the sender has not yet announced to that peer, or on a cache miss reply.
- The transaction-level verification (origin) needs the same mechanism at the destination, which may never have seen the origin. Either the first message to a destination carries the inline `Session`, or the destination requests it on miss. The miss round-trip is bounded by session TTL and is a one-time cost per (origin, destination) pair.
- Verification semantics do not change: a cached `Session` is exactly the value that would have been inline, judged against the same `ts_ms`/`ttl_ms` rules.

## Acceptance

- Steady-state relayed payload carries two 20-byte session ids and two 65-byte signatures of identity material.
- Cache miss path is tested for both hop verification and origin verification.
- Session expiry evicts the cache entry and forces re-announcement.

RyanKung:
Closes #738

## What changes

Every `MessagePayload` carried two full `Session` delegations (origin and current hop) on every hop. A link now sends each delegation once and its 20-byte content address afterwards. A steady-state relayed payload carries two 20-byte digests and two 65-byte signatures of identity material.

This is a wire-format change and the last PR of the pre-1.0 protocol epoch, so it bumps the workspace to **0.28.0**. Payload frames carry the `RINGS-PAYLOAD` marker, link-control frames `RINGS-LINK`; every 0.27.x frame fails closed. Seed nodes need the upgrade after merge.

The protocol is not versioned before 1.0, so the version suffixes earlier releases had put on wire markers, signing domains, AEAD namespaces, HKDF labels and storage keys are removed in the same cutover (`RINGS-TX-V2` → `RINGS-PAYLOAD`, `RINGS-PROVISIONAL-SERVICE-{CLAIM,RECEIPT}-V1` → without suffix (golden vectors regenerated), `rings-core:message-verification:transaction:v2` → `...:transaction`, `...:payload:v1` → `...:payload`, the inbox, receipt, descriptor, onion and ElGamal domains, `rings-core:transaction-replay:v2` → `...:transaction-replay`, `rings-node:onion-entry-guards:v1` → `...:onion-entry-guards`). Each name is one domain; a change to it is a total cutover, never a version beside another. Consequently every signature domain changes in this release, and a replay window or entry-guard set persisted by 0.27.x is not read, which loses nothing since no 0.27.x frame is accepted either.

## Where the issue's proposal was changed, and why

**1. The reference is a content address, not `session_id`.**
A delegation needs no proof of possession of the session key, so any account can sign `Session { session_id: <someone else's key>, account: <mine> }`. With inline sessions that is harmless (it is plagiarism of a message one could have authored anyway). With a cache keyed by `session_id` it becomes misattribution: the victim's by-reference messages resolve to the attacker's account wherever the attacker's delegation was cached first. The same account can also legitimately hold two live delegations of one key, and `Transaction::digest()` covers the session bytes, so resolving "the other one" would change the digest behind replay and fork evidence. `SessionDigest` is the trailing 20 bytes of `keccak256(encode(Session))`, the same size as a DID, and makes the round-trip law a theorem: a resolved frame is exactly the payload that would have travelled inline.

**2. The cache is scoped to one direction of one connection generation, not node-wide `(account, session_id)`.**
Checked in the code before designing: every hop verifies *both* proofs (`verify_transaction_and_payload`), not only the hop proof. So the origin session is needed on every link of the path, and the issue's "one-time cost per (origin, destination) pair" would really be per (origin, hop). Per-link tables answer the open questions at once:
- *Admission*: a receiver table is populated only by the peer at the other end, only from frames that verified or from an announcement this end asked for whose delegation verified. Unsolicited announcements are ignored. A peer can spend only its own bounded tables and hold.
- *Origin sessions*: inline-first, performed by the **last hop**, not by the origin. A forwarding hop re-encodes the origin slot from the table of its own next link (neither signature covers the session slot). A destination that never met the origin gets the session inline from its last hop until it confirms it: no routed "request the session from the origin" message, which would have needed replay/quota admission and a return path.
- *Restart*: both tables die with the connection generation. There is no state to resynchronise.
- *Rotation*: a new delegation is a new digest, sent inline until confirmed; old entries leave by expiry or LRU.
- *Stale announcements*: accepted only while some held frame awaits them and only if `Session::verify_self_at` accepts the delegation now.

**3. The link is a datagram link; the sender references only what the receiver confirmed.**
Nothing assumes ordered or reliable delivery (WebRTC data channels happen to be both; QUIC datagrams or UDP would carry the protocol unchanged). The sender keeps a session inline until the receiver sends `Known(digest)`, which the receiver sends for every inline session it verified. So a lost inline frame, a lost confirmation, or any reordering costs inline frames and never a stall: a reference is sent after the confirmation, which was sent after the session was learned.

*Superset law.* The receiver's table holds 128 sessions to the sender's 64, under one least-recently-referenced order over the frames both ends saw (the sender touches on every frame it encodes, the receiver on every frame it resolved or verified), so on a lossless link the sender goes back to inline before the receiver could have forgotten. A frame lost on the link touches only the sender's order, so under loss the two drift and the receiver may evict a session the sender still references; that miss is answered from the sender's table at the cost of one round trip and no charge. A miss the sender cannot answer needs something outside the tables (the two ends disagreeing on expiry, or a misbehaving peer). The miss path is the safety net for all of these, repaired on the link: the receiver holds that frame alone (no head-of-line queue: there is no order to preserve), asks `Request(digest)` for what it misses, the sender answers with exactly one `Announce(session)` or `Unknown(digest)`, and the frame leaves when the session it awaits is learned, from the announcement or from any later frame carrying it inline.

*Bounds and charging.* The hold keeps at most 16 frames per connection (half the pre-admission hold, so both holds leave a quarter of the transport's per-peer frames for the control frames that release them), each for at most twice the transport's delivery timeout plus one period of the inbound actor's periodic cleanup, which sweeps what waited longer, whatever lifetime the frame's unverified proof claims. Every held frame asks for its own missing digests once on arrival, one `Request` per missing session (one lost question is repaired by the next frame that misses the same digest); a frame that finds the hold full is dropped and the oldest held frame's question is asked again, so the question side is bounded by inbound frames. A frame the peer was asked about and did not back (disclaimed or invalid announcement, hold timeout, failure on release) is charged to the peer as a receive failure, as a frame that fails verification is; a frame that finds the hold full is a loss at this end's capacity, like one the pre-admission hold cannot take, and is not charged, since under loss-induced drift an honest sender can burst more than 16 references to a session the receiver evicted before its round trip returns. A held frame whose question this end never managed to send (budget spent, generation ended) is swept uncharged: the hold records which digests were actually asked about (`note_asked`), and the sweep tells unanswered frames from unasked ones. The announcement verdict is a three-way `Announcement { Ignored, Admitted, Refused }`, and a release is started only when some held frame awaits what was learned.

*Link-control frames* (`Known`, `Request`, `Announce`, `Unknown`) are unsigned and idempotent on purpose: they only mean something on the authenticated connection they arrive on, an announced session authenticates itself, a digest names a value, and a duplicate or stale one changes nothing. They are link-layer signals, not messages, so they go straight to the data channel rather than through the transfer lanes, which serialize a class behind each frame's 300 ms flush poll; routing them through the DHT-control lane delayed relayed handshakes enough to fail the three-node WebRTC test. Their emission (`transport/link_control.rs`) obeys three laws: a frame is sent on exactly the connection generation it was judged on and dropped if that generation is no longer current (both ends' tables are per generation, so a late `Known` or `Unknown` on a newer generation would confirm or disclaim what its tables never saw); the send runs in a task of its own under the same frame-send discipline as every payload frame (cancelled while revocable if the generation is superseded, retired through the termination path if it became irrevocable and timed out) and is never awaited from the transport's read loop, which on native holds the data channel's send lock across the SCTP window and would otherwise stall this end's reading, and so the peer's sending, on a link both ends still use; and nothing reserves lane or memory capacity for them: each inbound frame causes at most two (a confirmation or question per slot of a frame verified, held, or dropped for want of room; one answer per question), and at most `LINK_CONTROL_IN_FLIGHT_CAPACITY` = 2 × `rings_transport::callback::INBOUND_PEER_FRAME_CAPACITY` = 128 of them, twice the raw frames the peer may have in flight at this end's transport, are in flight to one peer at a time, counted by a per-peer budget the send task holds until it ends; a send beyond it is refused, and repeated by the next frame that misses or teaches the same session, so what this end spends on a peer's link control is bounded by the frames it accepts from that peer. The permit is taken inside the same lifecycle-locked step that proves the generation current, so a control frame never looks a worker up for a peer retirement is removing (and never spawns one: `link_of` is a lookup). A runtime that cannot carry the task refuses the send (typed) rather than running it inline. The budget and the announced table form `PeerLinkState`, which survives a replacement of the outbound worker under an unchanged generation, so a worker restart no longer disclaims what the old worker announced.

On the receiving side the link law is: a reference resolves only on the link this callback is bound to (its handshake's peer and generation), every control frame names that generation, only frames on the link teach the link's table, and every held frame was admitted under that peer, so every drop charges one identity. A payload frame from any other peer, or on a callback bound to no handshake, is judged self-contained (`WirePayload::into_self_contained`), a reference in it is refused, and it teaches nothing. Verification (`verify_frame`) and learning (`admit_verified`) are separate steps, and what the link learns from a verified frame is confirmed, and held frames released against it (in a task of their own: a release admits up to 16 frames and must not pace the read loop), before that frame is gated: a held frame never waits on the fate of the frame that taught it. A frame released or delivered after its connection generation was superseded is dropped by the gate (`InboundGate::Superseded`), never held for an admission that will not come. Every task the swarm detaches (inbound actor, pre-admission drain, session release, control send) goes through one boxed boundary, `swarm/detached.rs`, whose `run_detached_or_inline` states the law of the inline fallback: only a task whose omission would leave claimed state or charge a peer for this end's inaction (a drain, a release) runs inline when no runtime is current; a control send is refused.

This is the shape the sync-storm harness demanded and I initially dodged with a `FrameDelivery` capability: under its adversary (bootstrap frames discarded, LIFO / seeded / control-last delivery within one connection) the first design produced holds and extra control frames that the harness's 1:1 frame model could not follow. With confirmations, the same adversary only ever delays the switch to references, so every sync-storm scenario runs with references enabled, asserts that references were actually used on links of that scenario (per-link counts before and after, since replays reuse the deterministic node keys), and converges unchanged. The simulation classifies a link-control frame as its own `LinkControl` delivery class, with no deadline and outside the transfer-class model (like a lifecycle event): a delayed confirmation costs inline frames, never liveness, so it is neither a control-starvation verdict nor protected from the adversary; the LIFO and control-last strategies now delay these frames freely and still converge. `trace.rs` is unchanged from master.

**4. Expiry.** An expired session is absent from both tables. A reference to it is a miss, and re-announcing the expired delegation is refused by the same predicate that refuses it inline, so expiry forces a *fresh* delegation and never resurrects an old one.

## Structure

- `crates/core/src/session/digest.rs`: `SessionDigest`, `Session::digest`.
- `crates/core/src/message/payload/wire.rs`: `SessionRef`, `WirePayload` (borrowed view to encode, owned to decode, `resolve` is a traverse by move, `into_self_contained` the all-inline reading), `LinkControl`, `LinkFrame`, `PerSlot<T>`. `MessagePayload`/`MessageVerification` are untouched, so inbox carriers, service receipts, descriptors, onion authentication, and handshake payloads stay self-contained; `to_wire`/`from_wire` remain the all-inline encoding and refuse a reference with `SessionReferenceUnresolved`.
- `crates/core/src/swarm/session_link.rs`: the two pure state machines, time as an argument, effects returned as data (`FrameArrival`, the released frame, the digests to confirm, the carriers to fail).
- Shell: `transport/outbound/session_encoding.rs` (the worker encodes against the peer's table immediately before the send; the inbound side marks confirmations and answers questions on the same table; the per-peer link-control budget) and `callback/link_stage.rs` (interprets arrivals, releases, control frames; the link and learning laws above), with `transport/link_control.rs` emitting the control frames under the three laws above and `swarm/detached.rs` as the one spawn boundary. Chunked payloads are cut from the self-contained encoding because they are decoded after reassembly, outside the order of the link; the chunk frames themselves are link frames.
- To keep `outbound.rs` under 1000 lines, its test-only counters and `SwarmTransport` test accessors moved into the existing `test_trace.rs` / `simulation_pressure.rs` unchanged.

Placement: everything new is protocol and lives in `crates/core`; `rings_transport` only exports its existing per-peer raw-frame bound (`INBOUND_PEER_FRAME_CAPACITY`) so the link-control budget derives from the transport's own quantity, `crates/rpc` is untouched, and `crates/node` loses version suffixes on two storage directory names, two storage keys, and its onion/descriptor domain tags and AEAD namespaces. Exposure: only `SessionDigest`, `Session::digest`, `Error::SessionReferenceUnresolved` and `Error::LinkControlOutsideLink` are public; the wire view, control frames, and both tables are crate-private, and the receiver table is reachable only by the peer of its own connection.

## Tests

- `swarm/session_link/test_session_link.rs`: soundness (nothing unconfirmed is referenced), unannounced confirmations mark nothing, steady-state size saving equals the two delegations minus two 20-byte digests, generation scope, both table bounds with LRU and the **superset law** (sender forgets first, receiver still resolves) and its failure **under loss** (lost frames drift the two orders, the receiver evicts what the sender still references, the sender answers), round trip incl. `Transaction::digest`, **loss and reordering before confirmation never miss**, **origin miss**, **hop miss**, miss answered from the sender table, held frames never wait for each other and resolvable ones leave earliest first, overflow drops the newcomer and re-asks the oldest question, unsolicited announcement ignored, disclaimer fails exactly the awaiting frames, idempotent re-announcement, **expiry evicts and refuses the expired delegation** (sender and receiver), the sweep drops frames past the hold timeout or their proof. All over explicit instants.
- `transport/tests/test_inbound/test_session_reference.rs`: the link stage over a dummy connection (a stranger's session referenced, held, released by announcement or failed by disclaimer; a resolvable frame passes a held one; the hop miss asks once on arrival and the released frame's inline slot is confirmed, witnessed on the dispatched control frames; the hold overflow drops the newcomer uncharged and re-asks the oldest question; the receiver table does not outlive the connection generation; **a control frame judged on a superseded generation is not sent** (and the late frame is neither charged nor held); **with the dummy's dispatch gate closed, an inline frame still completes and its confirmation is dispatched** (the detachment law); **with the peer's whole in-flight budget held, a confirmation is refused and the frame still delivered, and confirmed again once a send ends**; **a question is answered from the announced table** (`Announce` for a session this end sent, `Unknown` for one it never sent); a referenced frame on an unbound callback, and from a peer other than the bound one, is refused with the typed `SessionReferenceUnresolved` and a stranger's announcement releases nothing; **a frame teaches the link before it is gated** (pre-admission: an inline then a referenced frame both held for admission, the session hold empty); an expired announcement is refused and charged (the delegation stamped before the inbound clock is captured, its lifetime shorter than the hold timeout); a frame held past the timeout is swept by the inbound actor under an injected clock and charged), waiting on the application callback's delivery event or on two cleanup passes. The pure layer also witnesses that the sweep tells unasked frames from unanswered ones, and `test_outbound.rs` that the link's tables outlive a replaced worker.
- `tests/default/test_session_link.rs`: two real swarms reach references through the confirmation exchange without any question being asked; three swarms relay an origin's messages until a frame on the relay's link to the destination carries the origin's own session digest in its origin slot, with no question anywhere on the path.
- `message/payload/test_payload.rs`: the self-contained decoder refuses a referenced frame and a link-control frame with their typed errors; an unmarked frame and a `RINGS-TX-V2`-marked 0.27.x frame both fail closed. `session/test_session.rs` freezes a golden vector of `SessionDigest`. `transport/outbound/session_encoding.rs` tests the link-control budget; the simulation classification test covers the `LinkControl` class.
- `test_decode_boundaries.rs`: generated, mutated link frames (including `Known`, and announcements of sessions earlier generated frames referenced, so releases run under generated input) keep the receiver within its bounds under arrivals, releases and sweeps (runs in the CI decode-boundary job).
- The whole sync-storm suite (ring/hotspot, LIFO, seeded, adversarial-control-last, N=10 ablation, thirty identical replays) runs with references on and asserts, per scenario, that some of its links went by reference.

## Docs

`SECURITY.md` Layer Contracts, new mdBook page `advanced-topic/session-references.md`, the cutover note in `transaction-replay.md`, `CHANGELOG.md`. The papers do not describe the envelope. There is no `fuzz/` directory on master (#741/#747 is not merged), so there are no fuzz targets to update; the decode-boundary test above is the in-tree equivalent.

## Verified locally

`cargo +nightly fmt --all -- --check`; clippy `-D warnings` for native (`--all --tests`), core wasm, transport wasm, node browser, node ffi, and core `--features dummy --no-default-features --all-targets`; `RUSTDOCFLAGS=-Dwarnings cargo doc --all --no-deps`; `taplo format --check`; `typos .`; `cargo test -p rings-core --features dummy --no-default-features` (921 passed, doctests included); `cargo test -p rings-transport --no-default-features --features dummy` and `cargo test -p rings-transport --lib`; `cargo test -p rings-node --lib`; the three-node native WebRTC test `test_handle_connect_node` (`cargo test -p rings-core --lib`), which the lane-scheduled control frames had made fail.

Not verified locally: the browser (wasm-pack/headless Chrome) suites, which only run on CI; the release-profile `cargo test --all --all-targets` matrix.


RyanKung:
74faf709 — the link is now treated as a datagram link, and `FrameDelivery` is gone.

- The sender references a session only after the receiver confirms it (`SessionControl::Known(digest)`, sent for every inline session the receiver verified). Until then every frame stays inline, so loss or reordering on the link costs inline frames and never a stall or a miss; a miss needs the receiver to have forgotten a confirmed session and is repaired on the link as before.
- Held frames are independent (no head-of-line queue): a frame that resolves never waits behind one that does not. A held frame leaves when a later frame or an announcement teaches its session, or lapses.
- The sender table needs no commit-on-accept any more (`FramePlan` removed); `AnnouncedSessions::encode` is the one step, and the confirmation marks the entry.
- `ConnectionInterface::frame_delivery` is removed from `rings_transport`; nothing in the protocol depends on delivery order.
- Every sync-storm scenario (LIFO, seeded, adversarial control-last, N=10 ablation, thirty identical replays) now runs with references enabled. The harness needed one statement: a link-control frame has no transaction identity and no outbound-submission boundary; link-control transfers are not counted as message submissions.

Tests restated as datagram-link laws (loss and reordering before confirmation never miss; held frames do not block resolvable ones; questions once per awaited digest and again on overflow); the two end-to-end tests witness the switch to references without any question asked. Local gates green again: nightly fmt, clippy (native, core wasm, transport wasm, node browser, ffi, core dummy all-targets), rustdoc `-Dwarnings`, taplo, typos, mdBook, `cargo test -p rings-core --features dummy --no-default-features` (904 passed), `cargo test -p rings-transport --no-default-features --features dummy`. The PR description is updated to match.

RyanKung:
43291e6e — the protocol is not versioned before 1.0, so every version suffix is gone.

- Wire markers: `RINGS-PAYLOAD`, `RINGS-LINK`, `RINGS-PROVISIONAL-SERVICE-CLAIM`, `RINGS-PROVISIONAL-SERVICE-RECEIPT` (the receipt golden vectors are regenerated).
- Signing domains: `rings-core:message-verification:transaction`, `...:payload`, `rings-core:relay-inbox:held-message`, both provisional service-receipt role domains, `rings-node:online-node-descriptor`, `rings-node:onion-exit-descriptor`, `rings-node:onion-backward-payload`.
- AEAD namespaces and HKDF labels: `rings-node:onion-circuit`, `rings-node:onion-cell`, the secp256k1 ElGamal AEAD salt and info.
- Storage keys: `rings-core:transaction-replay`, `rings-node:onion-entry-guards`. A replay window or entry-guard set persisted by 0.27.x is not read; nothing is lost, since no 0.27.x frame is accepted either.

Each name is one domain; a change to it is a total cutover, never a version beside another. SECURITY.md, the Session References and Transaction Replay pages, and the CHANGELOG say so; the PR description is updated. Local gates green: nightly fmt, clippy (native, core wasm, node browser, ffi, core dummy all-targets), rustdoc `-Dwarnings`, typos, taplo, mdBook, `cargo test -p rings-core --features dummy --no-default-features` (904 passed), `cargo test -p rings-node --lib` (385 passed), `cargo test -p rings-transport --no-default-features --features dummy`. In the native `cargo test -p rings-core --lib` run only the three-node WebRTC test `test_handle_connect_node` fails locally (connection never reaches `Connected` within 5s), which is the known local limitation of that test; CI is the witness for it.

zkjoie:
ef175035 — fixes from a four-angle review (protocol/security, concurrency/lifecycle, abstraction, tests/docs) and the CI failure.

**CI failure (`test_handle_connect_node`, three-node WebRTC).** Not the local limitation I claimed earlier; it passes on master and failed on this branch. Cause: link-control frames were scheduled as DHT-control transfers, and a transfer lane admits its next frame only after the previous one's 300 ms flush poll, so every `Known` confirmation delayed the relayed handshake past its window. Link-control frames are link-layer signals, not messages: they now go straight to the data channel (bounded by the send-accept timeout), outside the transfer lanes; `OutboundTransfer::link_control` and `FrameSource::LinkControl` are gone. The test passes again (≈5.7 s, faster than master's 11 s).

**Protocol findings fixed**
- Held frames were bounded in count but not in time (an attacker-chosen proof TTL pinned transport leases for up to 100 min, and the sweep ran only on control events). A held frame now waits at most `SESSION_HOLD_TIMEOUT` (twice the delivery timeout), swept by the inbound actor's periodic cleanup on the injected inbound clock; hold capacity is half the pre-admission hold (16), so both holds leave a quarter of the per-peer transport frames for the control frames that release them.
- Overflow re-asked every awaited digest (up to 66 questions per dropped frame) and included the dropped frame's own digests. Now: every held frame asks its own missing digests once on arrival (a lost question is repaired by the next frame missing the same digest); an overflow drops the newcomer and re-asks only the oldest held frame's question.
- The sender could answer `Unknown` for a digest it had itself referenced after evicting it (silent loss charged to the honest peer). Superset law: the receiver's table holds 128 to the sender's 64 under one LRU order, so the sender goes back to inline before the receiver could have forgotten; the miss path is a safety net (expiry disagreement, misbehaving peer), not a steady state.
- Charging: every frame the link drops (overflow, disclaimed/invalid announcement, timeout) is now a receive failure, like a frame failing verification; before, only `Unknown` drops were.
- Held frames leave in arrival order (`remove`, not `swap_remove`), so a released span of one stream fits the replay window as it did.
- Released held frames are delivered detached, so one announcement releasing many frames does not pace the transport's read loop behind each frame's handlers.
- `Session::is_expired_at` saturates; an unsigned announcement with `ts_ms = u128::MAX` cannot panic this end.
- Two per-peer-bound docs contradicted the code (pre-admission plus session hold); stated and sized as above.

**Abstraction findings fixed**: `LinkControl` is the one name for control frames (marker, enum, sender, handler); `SlotEncoding` replaces `PerSlot<bool>`; `InboundFrameLease` is the transport residue shared by both holds; `Digests` is a `BTreeSet`; `authentication_of`, `announced_of`, `bound_attempt` name the repeated predicates; `AnnouncedSessions` carries `Option<u64>` instead of a zero generation; table admission clones the session only on first sight; a link-control frame decoded outside a link is the typed `Error::LinkControlOutsideLink`; `handle_link_control` is four arms of effects. Not taken: making `Transaction`/`MessagePayload` generic over the slot to remove the wire mirror (public-type surface; the round-trip law is witnessed by tests instead).

**Tests/docs findings fixed**: the relayed end-to-end test now witnesses the origin slot by reference on the relay→destination link (per-link slot counters), the two-swarm test witnesses the switch; the sync-storm outcome asserts references were used; the receiver table bound/LRU and the superset law have a witness; the self-contained decoder's typed refusals are tested; the sweep is tested pure and through the actor; the answer counter counts disclaimers too; stale `RINGS-LINK-V1`, the `V1` sentence in the DRanking receipts page, the replay page's "no acknowledgement protocol" sentence and the two `-v2`/`-v1` storage directory names are fixed.

Local gates green: nightly fmt, clippy (native, core wasm, node browser, ffi, core dummy all-targets), rustdoc `-Dwarnings`, typos, taplo, mdBook; `cargo test -p rings-core --features dummy --no-default-features` (907 passed), `cargo test -p rings-node --lib` (385), `cargo test -p rings-transport --no-default-features --features dummy` (81), `test_handle_connect_node` native. Not run locally: the doctests (this machine's Xcode license currently blocks the linker) and the browser suites. The PR description is updated.

zkjoie:
Second four-angle review round (protocol/security, concurrency/lifecycle, Rust abstraction, tests/docs), fixed in 834319e9. Full `origin/master...HEAD` diff re-read, not a closure check.

**Generation and read-loop laws for link control** (three agents converged on this). Control frames were sent on the peer's *current* generation with the bound attempt used only to pick the table, and the send was awaited inside the transport's read loop. Consequences: a `Known`/`Unknown` judged on a superseded generation could leave on the new one and confirm or disclaim what its tables never saw (a soundness-law violation masked only by data-channel ordering, which the PR disclaims); on native the awaited send holds the channel's send lock across the SCTP window, so a peer with a closed receive window stalled this end's reading and, symmetrically, its own; and a timed-out send that had already become irrevocable was dropped by `select!`, which retires the connection through the guard rather than `terminate_accepted_connection`. Now `send_link_control(attempt, control)` sends on exactly that generation's `AdmittedConnection` (refused as `ConnectionAttemptSuperseded` otherwise) and runs the send in its own task under `send_data_with_timeout`, the same discipline as every payload frame; module `transport/link_control.rs` states the generation, detachment and bound laws.

**Learn before gate, link law, one charging identity.** Release was gated on the teaching frame's delivery: an admission error skipped `learned_sessions`, so frames held for a session the peer did back sat until the sweep charged the peer for it. `admit_resolved_frame` is now `verify_resolved_frame` + `gate_prepared_frame`; the shell confirms and releases between them. A reference resolves only on the bound link (`bound_attempt`): a payload from another peer or on an unbound callback is judged self-contained (`WirePayload::into_self_contained`), so `Held`/`Overflow` cannot occur without a link, and the sweep's charge (bound peer) is the identity every held frame was admitted under. `authentication_of` is the one predicate; `peer_authentication` is gone.

**Simulation class.** Link-control frames were `ScheduledDeliveryClass::Control` with a deadline, so the harness both produced false control-starvation verdicts for delayed confirmations and shielded them from the adversary. They are now their own `LinkControl` class, no deadline, outside the transfer model; LIFO and control-last delay them freely and still converge. The reference witness is per scenario and per link direction (replays reuse deterministic keys, so a thread-wide counter was vacuous after the first run). `trace.rs` is back to master.

**Laws restated to what the code does**: superset law under loss (orders drift, sender still answers, no charge); order law ("a held frame never waits for one held before it; among resolvable frames the earliest arrival leaves first", the old "arrival order" claim was false, the module's own test released 1,3,2); hold bound is 2×delivery + one sweep period; one `Request` per missing session, not per frame; CHANGELOG 64-vs-128 contradiction; "confirms every inline arrival it knows" → "that verified"; sender table dies lazily on the next generation's first frame.

**Abstraction/vestiges**: `SessionRef::inline` constructor (five spellings), `KnownSessions` alias, non-allocating `awaits`/`is_resolvable`, `waited_past`/`proof_lapsed_at` split with the hop-only choice stated, free `drop_held`, `LogicalCompletion` (no unused derives), `SESSION_HOLD_CAPACITY` const, `Error::UnmarkedFrame`, digest cut as an array pattern, `record_session_answer` without its ignored argument, phantom "same fields in the same order" comment removed, `reserve_outbound_capacity` back to the master signature, worker-restart note on the fresh announced table. Remaining version suffixes removed: `MeasurementLedger`/`ProvisionalEvidence` storage keys in the node (CHANGELOG names the loss), receipt prose, a test sled dir.

**Tests added**: shell overflow (charged, oldest question re-asked, newcomer's never asked), receiver table not outliving the generation, request-once + confirm-on-release witnessed on the emitted control frames (new `emitted_link_control_for_test`), relay witness on the origin's own digest (`referenced_origins_for_test`), sweep waits for two cleanup passes, decode-boundary announcements of previously referenced sessions.

Not taken: `admit` returning `&A` (double scan at ≤128 entries), `Overflow` carrier binding (the drop is the lease release; kept explicit), enqueue-failure charging on the detached path (consistent with the pre-admission drain, out of this PR's charging law).

Verified locally: nightly fmt; clippy `-D warnings` native, core dummy all-targets, core wasm, node browser, node ffi; rustdoc `-D warnings`; taplo; typos; core dummy suite 909 passed (doctests included); node lib 385; transport dummy 81; native three-node WebRTC test.

zkjoie:
Third four-angle review round (protocol/security, concurrency/lifecycle, Rust abstraction, tests/docs) on the full `origin/master...HEAD` diff, fixed in 2b765149.

**Learning separated from verification; off-link frames teach nothing** (three agents). `verify_resolved_frame` learned into the bound link's table for every frame it saw, including frames from a peer other than the bound one and frames on an unbound callback, which `submit_off_link` then discarded as `_unconfirmed` while the sessions had already been admitted. Now `verify_frame` verifies only; `admit_verified` is the link stage's own step on the link path; an off-link frame goes `into_self_contained` → verify → gate and never touches the table. `VerifiedFrame` is gone.

**Link-control sends bounded per peer.** The transport lease of a control frame was released before its spawned answer existed, so a peer that never read its window could hold one task per inbound frame for the whole accept timeout: the bound law was vacuous. Each peer now has `LINK_CONTROL_IN_FLIGHT_CAPACITY` = 2 × `INBOUND_PEER_CAPACITY` = 64 sends in flight, a permit the send task holds until it ends (`LinkControlBudget` next to the announced table); beyond it, `Error::LinkControlInFlightCapacity`. No runtime to carry the task is `Error::LinkControlRuntimeUnavailable`, never an inline await in the read loop (the detachment law now matches the code). The budget and the announced table survive an outbound-worker replacement under the same generation, closing the honest-charge path where a restarted worker answered `Unknown` for what the old one announced.

**Release detached.** `release_held_frames` ran inline in the read loop for the resolved and `Announce` paths (each released frame still awaits lane admission before the actor owns it). It now runs in a task of its own; every detached task in the swarm (inbound actor, pre-admission drain, session release, control send) goes through one boxed boundary, `swarm/detached.rs` (boxed because a release admits frames that may start a release: the future type is recursive otherwise). A released frame's learning is confirmed but starts no further release; the loop that freed it re-scans. `FrameProvenance { Arrived, Released }` replaces `LogicalCompletion`: provenance decides both the wait and the release.

**Overflow uncharged.** Under loss-induced drift an honest sender can burst more than 16 references to a session the receiver evicted before its round trip returns; charging frame 17+ contradicted the superset law's "no charge", and the pre-admission overflow is uncharged. The charging law now reads: frames the peer did not back (disclaimed, invalid or expired announcement, hold timeout, failure on release) are charged, through one `charge_dropped_frames`; a frame that finds the hold full is a capacity loss. Docs, CHANGELOG and SECURITY.md restated accordingly, including the overflow re-ask in the questions bound.

**Named predicates, vestiges.** `PendingConnectionAttempt::is_with`, `AnnouncedSessions::is_current` (three raw generation comparisons; `announced_on`/`live_on` inlined), `SessionTable::touch_live` (one scan instead of `live` + `touch`), `bound_attempt(Option<Did>)`, `resolve_slots` removed, duplicated doc sentence removed, `record_receive_failure_now` on the processor, one `submit_held_frame`. Test observables: one `referenced_slots_for_test()` projection and `dispatched_link_control_for_test()` (named for what it witnesses: judged current, within budget, handed to its task); the answer counter is derived from it.

**Witnesses added**: superseded-generation control frame not sent; detachment (every dummy send pending forever, an inline frame still completes and its `Known` is dispatched: the regression the three-node WebRTC failure was); unbound callback and non-bound peer references refused, stranger's `Announce` releases nothing; learn-before-gate under pre-admission; expired announcement refused and charged; overflow uncharged; superset under loss at the pure level; budget capacity; `SessionDigest` golden vector; `RINGS-TX-V2` 0.27.x frame fails closed; `LinkControl` simulation class. `SWITCH_WITHIN_MESSAGES` derivation corrected; storm comment weakened to what is asserted (some link direction of the scenario switched).

Not taken: `ReassemblyClock` → `InboundClock` rename (nit, touches unrelated files); `AEAD_VERSION`/schema-version fields are in-band format bytes, not name suffixes, left for the owner to rule on; the dummy send hooks firing on spawned control sends (`PendingAfterSentCountGuard` tests leave one hung `Known` task per test process) is noted, not changed.

Verified locally: nightly fmt; clippy `-D warnings` native, core dummy all-targets, core wasm, node browser, node ffi; rustdoc `-D warnings`; taplo; typos; core dummy suite 918 passed (doctests included); node lib 385; transport dummy 81; native three-node WebRTC test.

zkjoie:
Fourth four-angle review round (protocol/security, concurrency/lifecycle, Rust abstraction, tests/docs) on the full `origin/master...HEAD` diff, fixed in fb772721.

**Permit under the lifecycle lock; no spawn from the control path** (three agents). `link_control_permit` went through `handle(peer)`, a get-or-spawn, after `admitted_send_connection` had released the lifecycle boundary: a control frame arriving as retirement removed the worker could spawn a fresh worker + measurement task for a retired peer. The permit is now taken inside `with_current_connection`, from `link_of(peer)`, a lookup that never spawns; a peer with no worker is `SwarmMissDidInTable`. The announced table and the budget are one `PeerLinkState` (`outbound/link_state.rs`), carried across a worker replacement under an unchanged generation; `test_outbound.rs` witnesses that the replacement names the same tables.

**Budget derivation.** 2 × `INBOUND_PEER_CAPACITY` (the core mailbox's 32) was the wrong quantity: the link stage runs under the transport's raw-frame bound, 64 per peer. `LINK_CONTROL_IN_FLIGHT_CAPACITY` = 2 × `rings_transport::callback::INBOUND_PEER_FRAME_CAPACITY` = 128, the transport now exporting its own constant (the one line changed in `rings_transport`); the callback-side re-export and the widened `peer_capacity` are gone. SECURITY.md, the page and CHANGELOG say 128.

**Charge only what was asked.** A held frame whose `Request` was refused (budget, runtime, generation) was charged by the sweep as unanswered though the peer never had its round trip. The hold records which digests were actually asked about (`note_asked`, fed by what `request_sessions` dispatched) and the sweep returns `Swept { unanswered, unasked }`: unanswered charged, unasked dropped. Learning or disclaiming a digest spends its question.

**Announcement verdict and release fan-out.** `announce` returns `Announcement { Ignored, Admitted, Refused(frames) }`; a release starts only on `Admitted`, and on an arrived frame only when some held frame awaits what it taught (`awaits_any`), so an unsolicited `Announce` or a re-confirmation no longer spawns a task that scans an empty hold.

**Superseded generation.** `InboundGate::Unadmitted` conflated "not yet" with "no longer": a frame released by a late `Announce` on a retired generation sat in the pre-admission hold with its lease until the callback was dropped. The gate now answers `Superseded` (the registry's generation for the peer is no longer this one, in any state) and the frame is dropped, uncharged; the superseded-generation test asserts the late frame is verified, uncharged, not held, not delivered.

**Laws and names.** `run_detached_or_inline` states the inline-fallback law (a drain or a release runs inline when no runtime is current, since omitting it leaves claimed state or charges a peer for this end's inaction; a control send is refused); the release task's doc states why it is built outside the future (the `Send` proof would otherwise recurse). `is_with` at the four remaining raw peer comparisons; `FrameProvenance::releases_what_it_teaches` (no `PartialEq`); `HeldFrameDrop` names the three charging reasons; `answer` as two propositions; the unreachable `release_next` error is reported, not charged. Superset law names every drift source (loss, transport refusal, a frame dropped before it verified) and the sender's own eviction within a round trip; the overflow re-ask comment no longer claims the question overdue; CHANGELOG no longer lists `RINGS-LINK` among de-versioned markers.

**Witnesses.** Detachment now closes the dummy's dispatch gate (a send awaited in the read loop would sit behind it); the budget test holds the whole budget through the transport's permit, shows the refusal and the recovery; a `Request` is answered from the announced table (`Announce` / `Unknown`); the expired announcement is stamped before the inbound clock and judged while the frame is not yet sweepable; the sweep tells unasked from unanswered (pure); off-link refusals are matched by typed error; the storm's clock caveat is in its module doc.

Not taken: per-digest question dedup across held frames (bounded, within the questions law, a design choice); the global raw-frame pin by holds (pre-existing in kind with the pre-admission hold); the dummy send hooks seeing spawned control sends; `ReassemblyClock` rename; in-band schema/AEAD version bytes (owner's call).

Verified locally: nightly fmt; clippy `-D warnings` native, core dummy all-targets, core wasm, transport wasm, node browser, node ffi; rustdoc `-D warnings`; taplo; typos; core dummy suite 921 passed (doctests included); node lib 385; transport dummy 81 and default 70; native three-node WebRTC test.

Issue #775, opened 2026-09-14
Persist an age-ranked peer cache and re-join automatically when the successor list empties


r2d2-rs:
### Motivation

A node whose successor list empties never comes back on its own. `SuccessorSeq::min()` returns `self` (`crates/core/src/dht/successor.rs`), `notify_predecessor` and `pre_stabilize` skip, `find_successor` answers `Local(self)`, and nothing re-dials: `ConnectWithSeed` is a one-shot RPC (`crates/node/src/rpc_impl.rs`) and there is no persisted DHT peer state at all. The only persisted peer-shaped state is the onion entry-guard set and the measurement ledger, neither of which is used to reconnect.

#763 adds a run-owned redial supervisor for the native daemon's configured seeds. That leaves two gaps:

- **Browser nodes**: the provider has no equivalent, and a tab that loses its links has nothing but whatever seed the page last used.
- **Seed-only recovery**: re-dialing seeds concentrates every recovering node on the same few endpoints and discards what the node already learned about who stays online.

The literature is explicit about which peers to remember. Maymounkov & Mazières, "Kademlia" (IPTPS 2002), §2.2 keep the oldest live contacts because "the longer a node has been up, the more likely it is to remain up another hour" (Fig. 1, measured on Gnutella). Stutzbach & Rejaie, "Understanding Churn in Peer-to-Peer Networks" (IMC 2006), §4.5 confirm on three systems that observed uptime predicts residual lifetime (Weibull sessions with k ≈ 0.34–0.38). Godfrey, Shenker & Stoica, "Minimizing Churn in Distributed Systems" (SIGCOMM 2006) show that selecting by longest observed uptime beats random selection, and that "preference-list" selection driven by another metric (here: ring position) systematically raises churn — which is why the *cache* must be age-ranked even though the *ring* cannot be. BitTorrent's Mainline DHT and libp2p's Kademlia both persist their routing tables across restarts for the same reason.

### Proposal

1. **Age-ranked peer cache.** Maintain a bounded set (e.g. 32) of `(did, first_seen, last_seen, reachability hint)` records for admitted peers, ranked by observed uptime `last_seen − first_seen`, refreshed from the measurement ledger's `Connected`/`Disconnected` events. Persist it through `KvStorageInterface` next to the entry guards (`crates/node/src/onion/entry_guard.rs` is the pattern; browser localStorage is already used there).
[...]

PR #785, opened 2026-09-17
Model-check Chord rejoin races across connection generations (#772)

### LINKED ISSUE #772: Model-check bounded Chord disconnect/rejoin convergence across connection generations
### Motivation

Rings treats browser/WASM and native nodes as equal Chord peers. Correctness under repeated disconnect and rejoin should therefore be expressed as runtime-independent topology and lifecycle invariants, rather than assigning a special role to either runtime.

The connection lifecycle registry already assigns a local generation to each peer attempt and rejects stale callbacks for a superseded generation. The DHT tests also cover fixed-set convergence and several scoped StateRight models. What is not covered is their composition when a peer disconnects, is removed, rejoins under a new connection generation, and delayed work from the retired generation is delivered afterward.

This is distinct from:

- #770 / #768, which cover finger-range evidence, request correlation, and bounded finger maintenance.
- #763, which covers run-owned bootstrap reachability and automatic redial.
- #686, whose deterministic congestion scenario explicitly excludes crash and restart behavior.

### Goal

Add a bounded model that checks Chord topology safety throughout disconnect/rejoin races and conditional convergence after churn stops.

The model must not distinguish WASM from native peers. A rejoin is a lifecycle transition of an ordinary peer.

### Required transitions

The explored executions should include:

- connection reservation and admission;
- disconnect or send-terminal observation;
- unavailable-peer removal and successor replacement;
- rejoin of the same peer under a newer local connection generation;
- delayed disconnect, terminal, or report events from the retired generation;
- stabilization request and report delivery, duplication, bounded loss, and reordering;
- a quiescent suffix in which churn has stopped and enabled protocol work is delivered fairly.

At least one bounded configuration must contain more eligible peers than successor capacity, so successor truncation and replacement are exercised.

### Safety properties

For every reachable state:

- An event belonging to a retired connection generation cannot revoke, remove, or mutate the newer active generation.
- A stale event cannot remove topology evidence installed by the newer generation.
- Successor entries are clockwise ordered, distinct, bounded by capacity, and exclude self.
- Predecessor and finger state preserve the existing Chord topology invariants.
- Every remote routing step remains a strict clockwise advance.
- The model reports a minimal counterexample trace when a property fails.

### Conditional liveness

After the final churn event, if the remaining overlay retains a reachable successor path and enabled protocol messages are delivered fairly, every node must eventually reach the production Chord fixpoint for the final live-node set.

The issue does not claim convergence during endless adversarial churn or after a permanent partition.

### Fidelity requirements

- Exercise the production topology transition through topology::step where applicable.
- Exercise production connection-generation checks where practical. If a hashable snapshot or specification operator is required, prove its observable transition equal to the production transition over the modeled domain.
- Do not maintain an unvalidated shadow retry or lifecycle state machine.
- Include deterministic replay of failing traces.
- Document the chosen node count, successor capacity, search depth, state count, and CI wall-clock limit.
- Keep the bounded checker suitable for normal CI.

### Non-goals

- Distinguishing or assigning different responsibilities to WASM and native nodes.
- Defining a globally authoritative latest session for one account or DID.
- Proving arbitrary permanent partition, Byzantine behavior, or unbounded churn.
- Storage durability, replica quorum acknowledgement, or handoff deletion semantics.
- Reimplementing the finger-range scheduler from #770.
- Implementing bootstrap redial from #763.
- Traffic or throughput benchmarking.

### Acceptance criteria

- A checked trace covers disconnect, removal, same-peer rejoin with a newer generation, and delivery of an old-generation event.
- The old-generation event is proven unable to affect the current connection or topology state.
- Safety properties hold across all states within the documented bound.
- Conditional convergence holds for every fair quiescent suffix within the documented bound.
- Tests fail with a useful counterexample if the exact-generation guard or successor-replacement invariant is deliberately removed.
- The model and its scope limitations are documented next to the existing DHT model-checking tests.

### LINKED ISSUE #786: stabilize_successors drops the reporter's last real successor (no terminal self entry exists)
### Observation

`crates/core/src/dht/topology/stabilization.rs`, `stabilize_successors`:

```rust
let mut known = vec![local];
for candidate in current.iter().copied().chain(topo_predecessor).chain(
    topo_successors
        .iter()
        .copied()
        .take(topo_successors.len().saturating_sub(1)),
) {
    push_unique(&mut known, candidate);
}
successors(&known, local, capacity)
```

The doc says the last entry is "the reporter's terminal self entry". No such entry exists on the wire: the `QueryForTopoInfoSend` handler (`message/handlers/connection.rs`) answers with `TopoInfo::try_from(&PeerRing)`, which is `topology_state().successors` verbatim, a list that never contains the reporter itself. So `take(len - 1)` drops the reporter's last *real* successor.

### When it matters

`successors(&known, local, capacity)` already truncates to capacity. If the reporter is the immediate head and its list is full (`K` entries), the dropped entry sits at ring position `K + 1` from the requester and would have been truncated anyway; the defect is invisible. It bites exactly when the dropped entry would have fitted: the reporter's list is shorter than `K` after churn (before its own refill), or the reporter is not the requester's immediate head. In both cases the requester's successor list stays one entry short for at least one more round, or until successor-list sync (`QuerySuccessorList`) supplies it.

Safety is unaffected (an omission, never a wrong entry). Convergence is delayed, not prevented.

### Why the rejoin model (#772 / #785) did not catch it

The model's configurations start from a full mesh and admit every peer through the transport, so successor lists are computed from admitted peers by `Admit` and refined by the report's *predecessor*; reported successor lists never carry information the requester lacks (with `K = 1` the reported list has one entry, which `take(0)` drops entirely, and with `K = 2` in a 4-ring the dropped entry is at position 3). The model also checks eventual convergence under fairness, not round complexity, and does not drive a bootstrap in which a peer is discoverable only through a report. A configuration that would expose it: a chain bootstrap (each peer connected only to its dialed contact) with `K ≥ 2`, where the requester must learn its second successor from the head's report.

### Proposal

- Drop `take(len - 1)` and let `successors(&known, local, capacity)` be the only truncation, matching classic Chord's `[s] ++ s.succ_list` followed by truncation to `r`; fix the doc.
- Add a unit law: every reported successor within capacity distance of the requester appears in the stabilized list (`∀p ∈ reported. rank(p) ≤ K ⇒ p ∈ stabilize_successors(...)`), which the current code violates whenever the reported list is shorter than `K`.
- Optionally extend the rejoin model with a chain-bootstrap configuration so report-driven discovery is under the exhaustive check.

Found while reviewing #785; the `take(len - 1)` dates back to the correct-chord default (#604).

RyanKung:
Closes #772
Closes #786

## What

Stage 6 of the DHT model checks: the **composition** the existing stages leave out. A peer disconnects, is removed (with successor replacement), rejoins under a newer local connection generation, and lifecycle events / reports bound to the retired generation arrive afterwards.

- `dht::topology::invariants` (test build) — named predicates on `TopologyState` (`successors_are_well_formed`, `predecessor_is_well_formed`, `fingers_are_well_formed`, `is_well_formed`, `routes_clockwise_toward`, `is_chord_fixpoint_of`), each with a falsifying witness per clause; quoted by this model and by the finger-retry model.
- `swarm::transport::test_rejoin_model` — the model. It lives inside `swarm::transport` because its carrier holds the production `ConnectionLifecycleRegistry`; the Stage list in `tests::default::test_dht_stateright` points to it. Module docs carry the TLA+-style spec (variables, `Init`, `Next`, laws, per-instance fairness), scope limits and the bounds table.

## Fidelity

- The carrier stores the production `TopologyState` and the production `ConnectionLifecycleRegistry`. Every topology change is `topology::step`; every lifecycle change is a registry method (`reserve`, `begin_admission`, `admitting_connection().activate()`, `mark_send_terminal`, `retire_active_if`, `remove_unadmitted`); a report's candidate budget is the production `StabilizationConnectionPlan`; confirmation is `TopoInfo::confirmed_by`; the removal flavours are the production `DhtPeerRemoval`. No snapshot, no shadow lifecycle or retry machine.
- The model owns only the composition `SwarmTransport` performs under its lifecycle boundary (`node.rs`, each step names the production path it interprets) and the physical environment (`overlay.rs`: links, envelopes, budgets, two bootstraps).
- The one place the model hands `step` a differently shaped argument (unnormalized replacement candidates) is proved observationally equal over every candidate subset, for removed heads and non-heads.
- Conformance: the acceptance trace is driven through a real `SwarmTransport` (production `K = 3`, 160 finger slots) and through the model on the close path, and under the dummy transport (routable data channels) on the send-terminal path through `disconnect_unavailable`; successors, predecessor, fingers and the admitted generation must agree after every step, and the retired generation's late events are inert on both sides.
- The search is plain single-threaded Rust (no Stateright): the same exhaustive checks run natively and in the browser test job against the wasm build. Only the conformance tests need the native harness.

## Properties

Safety (`□`, every reachable state, breadth-first, exhaustive):
- `RetiredGenerationsAreInert`: an event of a generation that no longer owns its slot leaves `(topology, lifecycles)` unchanged (history variable; a failure is a minimal trace);
- `UnavailableHeadsAreReplaced`: retiring the head as unavailable leaves `succ = Successors(Sendable ∖ {head}, n, K)`;
- `TopologyReferencesOnlyAdmitted`: `Referenced(n, p) ⇒ Active(n, p)`;
- `TopologiesAreWellFormed`: successor/predecessor/finger well-formedness (falsifiable only through `step`, kept so the composition inherits it).

Coverage (`◇`): a retired generation's event waits beside a newer admitted generation of the same peer; a head replacement has at least `K` sendable candidates (more than `K` in the `replacement` configuration, so it truncates).

Conditional liveness: for every reachable state in which every live peer's head is live, sendable and linked (`RetainsLiveHeads`), every fair churn-free behaviour satisfies `◇□ ChordFixpoint(live set)`. Fairness is per action instance (`WF` for dials, deliveries and events; `SF` for the periodic `Stabilize`, because the search's one-round-at-a-time bound disables an action production keeps continuously enabled). Decided on the explicit graph by Streett-style fair-SCC search; a trap counts only if it contains an unconverged state. The trap decision's branches (plain cycle, starved weak label, strong-label pruning, converged-only trap, dead state) each have a hand-built-graph unit test.

Non-vacuity (each mutant's minimal trace replays to the violation under the mutated shell and to none under the faithful shell; `Init` is always built faithfully):
- `CallbackIgnoresGeneration` (exact-generation guard removed) → `RetiredGenerationsAreInert`;
- `ReplacementPreserves` (unavailable head removed without replacement) → `UnavailableHeadsAreReplaced`;
- `RetireWithoutRemove` (registry retires, topology keeps the peer) → `TopologyReferencesOnlyAdmitted`;
- premise dropped → a replayable starved suffix (the #775 orphan).

## Production fix (#786)

`stabilize_successors` dropped the last entry of every reported successor list as "the reporter's terminal self entry", which the wire never carries (the `QueryForTopoInfoSend` handler reports `TopoInfo::try_from(&PeerRing)` verbatim); the requester lost the reporter's last real successor whenever the reported list was shorter than capacity. The `take(len - 1)` is removed (normalization to capacity is the only truncation, as in Chord), a unit law pins `p ∈ result ⇔ rank(p) ≤ K` for every reported `p`, the stale unit expectation and the TLA+ spec comment (`ButLast`) are corrected. The exhaustive claims are insensitive to this defect (it delays convergence by rounds; it does not change the fixpoint), so #786 is pinned by its unit law; the `chain` configuration puts report-driven discovery under the check but does not distinguish the fix from the defect.

## Bounds

Exhaustive, no depth cut-off; each test asserts its exact state count, depth and premise counts; the search aborts above two million states.

| configuration | peers | K | budget | states | depth | premise states (unsettled) |
|---|---|---|---|---|---|---|
| departure | 3 | 1 | 1 departure | 4 862 | 21 | — (premise-necessity test) |
| replacement | 4 | 1 | 1 cut | 109 840 | 35 | 91 312 (25 472) |
| truncation | 4 | 2 | 1 cut | 237 328 | 36 | 219 664 (174 048) |
| restart | 3 | 1 | 1 departure, 1 rejoin | 542 589 | 47 | 464 848 (408 404) |
| chain | 4 | 2 | none (chain bootstrap) | 509 815 | 60 | 509 815 (495 191) |
| lossy flap | 3 | 1 | 1 cut, 1 loss, 1 duplication, 1 refusal | 1 034 264 | 42 | 890 762 (653 760) |

Lossy flap runs only in the release jobs (native and browser). Wall-clock figures are in the module docs.

Scope limits (in the module docs): no claim under endless churn, permanent partition or Byzantine peers; remote finger ranges, per-candidate plan revalidation and cross-round stale tokens are Stage 5; not modeled: capacity eviction, successor-list sync, connect lookups, `NotifyPredecessorReport`, periodic notification beyond the head, storage repair, bootstrap pacing and seed selection; one stabilization round in flight overlay-wide; claim and commit in one step; readiness identified with the registry's sendable projection. The model does not distinguish wasm from native peers.

## Findings

No safety violation in production. The liveness premise is necessary: when a peer's only successor goes down and its close event wins the race with the unavailable-peer sweep, `SuccessorRemoval::Preserve` leaves an empty successor list although other admitted connections remain, and stabilization never refills it (#775). #786 was found while reviewing this PR and is fixed here.

No workspace version bump.


RyanKung:
420e110d — the model now runs in the browser test job as well, so wasm and native peers are checked by the same exhaustive search.

- Stateright is dropped for this stage (it spawns threads and is a non-wasm dependency). `search.rs` decides safety (minimal counterexample by breadth-first order), coverage, and fair-suffix liveness in one pass per configuration; plain single-threaded Rust.
- The module moves from the non-wasm `tests` tree to `swarm::transport::test_rejoin_model`, built for every test target. Only the shell conformance test stays native (it needs the `SwarmTransport` harness).
- Peers are shared between a state and its successors (`Arc<NodeState>`): peak memory of the million-state configuration drops from 960 MB to 625 MB, which is what lets it run in a tab.
- Measured on an M-series laptop: native release 52 s for the module, dev profile 139 s, headless Chrome (wasm32, release) 37 s for all nine browser tests. State counts unchanged and still asserted exactly; BFS depths are now deterministic (36 / 47 / 42).
- Local gates: fmt, clippy (native, dummy, wasm, ffi), rustdoc, taplo, and the browser run itself (Chrome 153 with a matching chromedriver).

zkjoie:
d61594f2 — addresses a four-angle review (fidelity/search soundness, strict Rust style, test non-vacuity/determinism, placement/exposure).

Substantive:
- **Liveness soundness.** The trap search ran inside `¬Stable`, so a fair trap consisting of `Converged` states only could be reported although its tour satisfies `□Converged`. The Streett recursion now runs over the whole protocol graph and a trap counts only if it contains an unconverged state; `Stable` is reported as the inhabited target, and the `◇□Converged ⇔ ◇Stable` claim is withdrawn.
- **Replacement mutant was a straw man.** Removing the replacement altogether (`Preserve` where production selects `Unavailable`) passed every law, because the liveness premise excluded the orphan it creates. New history-variable law `UnavailableHeadsAreReplaced` (`succ' = Successors(Sendable ∖ {head}, n, K)` after an unavailable head's retirement) with `ReplacementPreserves` as its mutant, found after 46 states; `AdmitBeforeActivation` witnesses `TopologyReferencesOnlyAdmitted`. Every mutant trace must also satisfy its law under the faithful shell.
- **Fidelity.** `Notify` is applied only from a sendable generation (`notify_admitted_predecessor`); the glare rule abandons the own offer only while it is unanswered (production's `New` raw state, reported by the environment from the link table); removal flavours are the production `DhtPeerRemoval`; docs corrected where they overstated (production does not await handshakes; the inbound gate holds rather than refuses).
- **Premise weakened** from strong connectivity to `RetainsLiveHeads` (every live peer's head is live, sendable, linked). The claim holds in every configuration and now covers the post-churn recovery states: restart 466 396 premise states (409 952 unsettled) vs 250 550 before. Premise counts are in the bounds table.
- `routing advances clockwise` is no longer a checked law: it is `find_successor`'s unconditional postcondition (true of every representable state), so no reachable state could falsify it; its unit theorem stays in `invariants.rs`.
- New `replacement` configuration (4 peers, K = 1, 1 cut; 114 640 states) where head replacement has more sendable candidates than capacity; the coverage law is the pre-state of such a replacement.
- Conformance now also drives the send-terminal path through `disconnect_unavailable` on the dummy transport (routable data channels), comparing the replacement composition with production; the close path stays as before.

Search: 128-bit fingerprints (collision assumption stated), `EXPLORATION_BOUND` = 2 000 000 states failing deterministically, depths asserted with state counts, replay panics on a disabled index.

Style/placement: `Budget` resources as an enum with named predicates; `next_state(&action)`; law identities as an enum; `SearchReport` sum type; `LifecycleEvent`/`ProtocolMessage` vocabulary disjoint from transport frames/callbacks; `dht::topology::invariants` is now a test-build module with `pub(crate)` predicates; `Hash` on the lifecycle enums is test-only; the tests tree exposes one `transport_with_key` helper.

State counts of the previous configurations are unchanged. Measured: native release 30 s (12 tests), headless Chrome 41 s (11 tests), core dummy suite 896 passed.

Not changed on purpose: dev-profile cost (restart exploration stays in the dev job, as agreed); one pre-existing observation surfaced by the review, `stabilize_successors` dropping the last reported entry as a "terminal self entry" that `TopoInfo::try_from` never includes, is outside this PR.

zkjoie:
24c5a7cd — the remaining review items.

- `search.rs`: `StateIndex` and `Label` are newtypes and every dense per-state / per-label array is a `Table<K, T>` indexed by its own key type, so the two index spaces are no longer interchangeable for the type checker; `explore` interns states and labels through the `Entry` API.
- The finger-retry model (Stage 5) now asserts `successors_are_well_formed`, `predecessor_is_well_formed` and `fingers_are_well_formed` at every explored state, so "invariants of `step`" is witnessed across the finger-lookup events the rejoin model does not drive; the `invariants` module doc names which model witnesses which events, and no longer claims the unit tests quote it.

Verified locally: native dev/dummy (rejoin model, finger-retry model, invariants: 35 passed; state counts, depths and premise counts unchanged), headless Chrome wasm32 release (11 passed, 42 s), core dummy suite (896 passed), fmt, clippy native/dummy/wasm/ffi, rustdoc, taplo.

Out of scope, for a separate issue: `stabilize_successors` drops the last reported successor as "the reporter's terminal self entry", but the `QueryForTopoInfoSend` handler reports `TopoInfo::try_from(&PeerRing)`, which never appends one, so the requester loses the reporter's last real entry each round.

zkjoie:
d7915877 — fixes #786 and adds the chain bootstrap.

- `stabilize_successors` no longer drops the last reported entry; normalization to capacity is the only truncation. Unit law added (every reported successor within capacity rank is retained); the one unit expectation that had encoded the dropped entry is corrected (`[20, 30]` → `[20, 30, 40]` at `K = 3`).
- New `chain` configuration (4 peers, K = 2, no churn, `ring[i]` dials `ring[i-1]`): report-driven successor discovery is under the exhaustive check; 509 815 states, depth 60, every reachable state satisfies the premise and the liveness claim holds. Its two churn witnesses are unsatisfiable by construction and asserted as such.
- `truncation` grows from 233 488 to 242 128 states with the fix; the other configurations are unchanged.
- Verified: native dev/dummy (model, topology, convergence, Stateright stages: 102 passed; the earlier full core suite run had 899 passed, its only failures being the two stale bounds and the stale expectation corrected here, plus the env-only sync-storm replay that was run under `--include-ignored`), headless Chrome wasm32 release (12 passed, 54 s), fmt, clippy native/dummy/wasm/ffi, rustdoc, taplo.

zkjoie:
Second four-angle review round addressed in the latest commit (see `git log`, "second review round of the rejoin model"); the PR description is rewritten to match the code.

Substantive:
- **Lost messages on a cut link.** `cut` now drops the envelopes the link carried, as production does; the suffix fairness no longer obliges their delivery. Every mesh configuration's graph shrinks (restart 552 309 → 542 589, truncation 242 128 → 237 328, replacement 114 640 → 109 840, lossy flap 1 057 016 → 1 034 264, departure 5 834 → 4 862; chain unchanged); all laws and the liveness claim hold at the new bounds, which are asserted with premise counts.
- **`OwnOfferState` was dead code** (an answered own offer always sits on a link whose far end the offerer still owns); removed, and the one real divergence from production's `New` guard (a pending generation on an already-dead link: production refuses until the close, the model abandons at once) is documented as a superset schedule.
- **Mutant witnesses.** `Init` is built faithfully under every mutation, so no counterexample can be the empty trace; `AdmitBeforeActivation` (whose witness was `Init`) is replaced by `RetireWithoutRemove`, found after 10 states with a trace ending in the retiring event; the replacement mutant asserts the departed head by identity.
- **Trap decision witnessed.** Hand-built graphs test each branch: plain cycle, weak label enabled throughout and untaken, strong-label pruning (both the residual-cycle and residual-path cases), converged-only traps excluded, dead states included. These run in the browser too (16 wasm tests now).
- **Spec/doc accuracy.** Per-instance fairness; `Deliver(offer)` spends the refusal budget; periodic notifications beyond the head and seed selection listed as unmodeled; the `chain` row states which laws are vacuous there and that it does not distinguish the #786 fix (pinned by the unit law); the wall-clock paragraph gives what was measured on this revision and marks the older release figures as such.
- **#786 follow-through.** `stabilize_successors` doc states the HMCC merge (not Chord's replacement), defines `rank`, and the law `p ∈ result ⇔ rank(p) ≤ K`; the TLA+ comment's `ButLast` is gone; `test_correct_stabilize_ignores_last_topo_successor` is renamed to what it witnesses; the merge law lives in `topology/tests/stabilization_tests.rs` (the parent file is back under 1000 lines).

Style: typed `Table<K, T>` keys construct themselves, the BFS frontier is a struct, `Budget` has one slot accessor, `configuration` takes the bootstrap, `TopologyState::is_well_formed` names the conjunction both models assert, conformance shares its setup and its inertness assertion, test-gated derives are consistent across the lifecycle types.

Verified locally: native dev/dummy (model, topology, convergence: 92 passed; full core suite 902 passed), headless Chrome wasm32 release (16 passed, 54 s), fmt, clippy native/dummy/wasm/ffi, rustdoc, taplo.

PR #813, opened 2026-09-23
Bound outbound cancellation using existing futures channels

RyanKung:
## Change

Addresses the outbound Mailbox acceptance item in #787 on master `43c22c9e` (including #789, #803 and #812). Keeps the existing futures submission channel, validation/close gate, wakeups and transfer-queue policy.

Collected submissions retain their permits, so the existing FIFO drain is bounded to 256 even with concurrent producers. Repeated cancellation commands lacked a corresponding bound. A one-slot futures channel coalesces those notifications; each drain reads one, then the existing queue scan visits at most 256 transfers. Control submitted before the FIFO drain is visible before frame selection; a submission racing the final empty read may enter the next iteration. There is no arbitrary cutoff ahead of control. The unchanged 4:1 policy requires runnable heads and executor/delivery progress, not a wall-clock guarantee.

One command handler now accepts either a batch or a single idle input. It collects cancellation results and releases all shutdown ownership before publication. This replaces the previous separate command, submission and scan handlers. No new files, custom queue, wakeup primitive or mailbox reducer are introduced.

Extend existing mailbox and cancellation fixtures rather than maintaining parallel test modules. Native/browser contracts cover bulk-before-control, notification coalescing and wakeups, cancellation-before-submit at the worker, queued cancellation behind a waiting head, and shutdown publication order. Existing native end-to-end tests cover the actual submit predicate and paused transport paths. Native thread contention remains platform-specific. The existing production queue model explores 13^6 six-action traces with eight slots; this is finite safety evidence, not arbitrary-schedule liveness.

## Validation

- Native outbound: **42 passed** (`cargo test -p rings-core --no-default-features --features dummy swarm::transport::outbound --lib`).
- Native end-to-end scheduler: **20 passed** (same command with filter `tests::default::test_outbound_scheduler`).
- Real headless Chrome outbound: **18 passed** (`cargo test -p rings-core --release --target wasm32-unknown-unknown --no-default-features --features wasm swarm::transport::outbound --lib`; matching ChromeDriver 153).
- Clippy passed for default native, dummy/no-default (`--all-targets`), and WASM (`--tests`), with `-D warnings`.
- Repository-pinned `cargo +nightly-2026-07-02 fmt --all --check`, scoped `typos`, and `git diff --check` passed.

The full workspace suite was not run locally; CI is tracked at the latest head. Native/browser features are mutually exclusive. No wire changes, DHT notify, Docker expansion, merge, release, or umbrella-issue closure.


zkjoie (approved):
LGTM

RyanKung (commented):
currently treats a full 0-capacity notification slot as a successful no-op. If another  fires before the worker drains the already-pending , that later wakeup is silently discarded, so the queue scan never runs again until some unrelated input arrives. That can leave already-stopped transfers sitting in  until a later unrelated wakeup, which is exactly the kind of cancellation gap this patch is trying to close.

RyanKung (commented):
send_coalesced() currently treats a full 0-capacity notification slot as a successful no-op. If another StopSource fires before the worker drains the already-pending CancelStopped, that later wakeup is silently discarded, so the queue scan never runs again until some unrelated input arrives. That can leave already-stopped transfers sitting in ready until a later unrelated wakeup, which is exactly the kind of cancellation gap this patch is trying to close.

PR #819, opened 2026-09-23
Move listener generation ownership into Processor

### LINKED ISSUE #809: Move listener generation ownership from the browser wrapper into Processor
## Problem
Browser `Provider::listen` serializes listener generations using a Provider-owned gate, including cooperative cleanup. Native `Provider::listen_with` forwards directly to `Processor::listen_with`.

The public Processor method creates a new stabilizer and registration loops on every invocation and has no equivalent ownership gate. Concurrent native listen calls on the same processor can therefore start duplicate maintenance/registration loops. Browser callers can also bypass the wrapper through Processor; independently constructed Providers wrapping the same Processor own different gates.

The invariant belongs to the shared processor: at most one listener generation owns maintenance and cleanup for that processor.

## Evidence
- [crates/node/src/provider/browser/provider/listener.rs:103](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/node/src/provider/browser/provider/listener.rs#L103) — browser wrapper gate.
- [crates/node/src/provider/mod.rs:376](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/node/src/provider/mod.rs#L376) — native forwarding.
- [crates/node/src/processor/mod.rs:523](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/node/src/processor/mod.rs#L523) — new loops per invocation.
- [crates/core/src/dht/stabilization/maintenance.rs:625](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/core/src/dht/stabilization/maintenance.rs#L625) — independent schedule per invocation.

## Acceptance
- [ ] Enforce generation ownership in shared Processor lifecycle state.
- [ ] Define whether a duplicate start queues or returns a typed already-running result; preserve/document browser startup semantics.
- [ ] Hold ownership through graceful cleanup and measurement flushing.
- [ ] Make native futures and browser promises thin adapters over the same lifecycle.
- [ ] Test concurrent starts, Provider clones, independently wrapped shared Processors, queued cancellation, and restart after cleanup.
- [ ] Verify that stopping one generation cannot stop or overlap a replacement generation.

## Review boundary
Static wasm/native review at `daa3c26570166dc3f57be184c5db9e5204aeb9bd`. No build, test execution, or runtime reproduction was performed for this finding.

RyanKung:
Closes #809

## Changes

- Move the listener serialization gate into shared `Processor` lifecycle state.
- Preserve queued startup semantics for browser listeners; `started()` resolves only after processor ownership is acquired.
- Retain ownership through cooperative loop shutdown and measurement flushing.
- Add regression coverage for queued cancellation, restart after cleanup, and isolation between listener stop sources.

## Validation

- `rustfmt --check` on the changed Rust files passed.
- `git diff --check` passed.
- Tests were added but not run locally.


crates/node/src/processor/builder.rs
```diff
@@ -228,6 +228,7 @@ impl ProcessorBuilder {
             stabilize_interval: self.stabilize_interval,
             online_node_registration,
             measure,
+            listener_gate: Arc::new(futures::lock::Mutex::new(())),
```
0xworm: what is listener_gate

0xworm: Why it's a "gate"

PR #826, opened 2026-09-23
Extract a shared onion HTTPS client with a native entry point (#810)

### LINKED ISSUE #810: Extract a shared onion HTTPS client and expose a native Rust entry point
## Capability gap
The HTTPS onion exit has native and browser executors, but the HTTPS onion client workflow is tied to the browser build.

Client request/response types, the pending-request table, completion handling and URL parsing are guarded by `cfg(any(test, rings_browser))`. Routing, sealing, sending and waiting for the response live in `BrowserOnionProxy::request_http`. Native's circuit handler sends client responses to the TCP runtime rather than an equivalent HTTPS client runtime.

This is a missing native client capability and misplaced shared logic, not a claim that native cannot serve HTTPS exits or that native TCP CONNECT is broken.

## Evidence
- [crates/node/src/onion/https/mod.rs:147](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/node/src/onion/https/mod.rs#L147) — browser/test-only client state.
- [crates/node/src/provider/browser/provider/onion_proxy.rs:40](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/node/src/provider/browser/provider/onion_proxy.rs#L40) — request workflow.
- [crates/node/src/onion/tcp/mod.rs:257](https://github.com/RingsNetwork/rings/blob/daa3c26570166dc3f57be184c5db9e5204aeb9bd/crates/node/src/onion/tcp/mod.rs#L257) — native client response dispatch.

## Acceptance
- [ ] Extract target-independent HTTPS onion client request lifecycle, pending state and response verification.
- [ ] Expose a native Rust client API and adapt existing browser bindings to the same implementation.
- [ ] Keep fetch/HTTP egress, runtime timers and JS conversions at platform boundaries.
- [ ] Preserve expected-peer/exit/return-id validation, drop cancellation and timeout cleanup.
- [ ] Test success, exit error, wrong return peer, invalid authenticated response, cancellation and timeout through the shared client.
- [ ] Verify both native and browser callers use that client; adding a public RPC endpoint is a separate scope decision.

## Review boundary
Static wasm/native review at `daa3c26570166dc3f57be184c5db9e5204aeb9bd`. No build, test execution, or runtime reproduction was performed for this finding.

zkjoie:
Closes #810.

## Problem
The HTTPS onion client was browser-only. Its types, pending table, completion and URL parsing were gated on `cfg(any(test, rings_browser))`. Routing, sealing, sending and waiting lived inside `BrowserOnionProxy::request_http`. Native nodes could serve HTTPS exits, but they could not send HTTPS requests over a circuit.

## Change
- **`onion/https/client.rs`**: new `OnionHttpsClient`, the same on native and browser builds (no target `cfg`).
  - `OnionHttpsCall::from_url(url, request) -> (OnionProxyTarget, OnionHttpsCall)`: the call holds a normalized method and path and **names no target**.
  - `begin(route, call)`: addresses the call to `route.target`, so a request cannot disagree with its route by construction. It then registers the circuit and seals the first forward cell into an `OnionHttpsFlight`.
  - `PendingOnionHttpsRequest::within(deadline)`: waits for the response. The deadline is passed in: `Ok(())` means the request timed out (`OnionProxyRequestTimedOut`), and a timer error is returned as-is.
  - `claim(peer, id) -> Option<OnionHttpsClaim>` takes the request that owns `(circuit id, return peer)`. `OnionHttpsClaim::resolve(payload, network)` checks the payload against that request's return id, the selected exit and the network, then delivers the outcome.
  - `request(scope, route, call)`: the shell = `begin` → `send_sealed` → `within(rings_runtime::sleep(30s))`.
- **`onion/native/`** (new): `NativeOnionCircuitHandle`, the native circuit handler, and the constructor for the shared resources. `onion::tcp` is TCP-only again.
  - New `NativeOnionCircuitHandle::request_https(&OnionProxyRoute, OnionHttpsCall) -> OnionHttpsResponse`.
  - Backward frames: the HTTPS client claims first; anything it does not claim goes to the TCP streams. The module doc states the client-disjointness law.
- **Browser**: `request_http` now does `OnionHttpsCall::from_url` → its own route builder → `OnionHttpsClient::request`. Method signatures and JS response shape are unchanged.
- **Platform boundaries**: route selection stays with the caller. The timer goes through `rings_runtime::sleep` (#828). Fetch and JS conversion stay where they were.
- **Subtractions**: `OnionHttpsClientResponse` is removed in favour of the identical `OnionHttpsResponse`. `OnionHttpsPayload` and `encode_https_payload` are now private to the `https` module.
- **New error variant**: `OnionRouteError::HttpsResponseClosed`, for a claim dropped before it is resolved.
- No public RPC endpoint is added; that stays a separate scope decision, as the issue says.

## Compatibility
- **Wire**: unchanged, no version bump.
- **Rust API**:
  - `rings_node::onion::tcp::NativeOnionCircuitHandle` moved to `rings_node::onion::native::NativeOnionCircuitHandle`; the in-repo bin is updated.
  - `OnionHttpsClientResponse` → `OnionHttpsResponse`, with the same fields.
  - `OnionRouteError` gains `HttpsResponseClosed`. The enum is not `#[non_exhaustive]`, but the in-repo matches have wildcards.
- **Behaviour**:
  - In the browser, a failing timer now returns its error instead of being reported as a timeout.
  - A poisoned lock during a claim now surfaces as `Error::Lock`.
  - On native, a backward frame from the wrong peer for an HTTPS circuit is no longer dropped silently; it goes to the TCP streams, which report it like any other unknown frame.

## Tests
These run through the shared client and are target-independent (polled once with a noop waker, no wall-clock waits):
- success, exit failure, wrong return peer (no claim; the request stays pending, then times out), payload signed by another exit, payload for another return id, request echoed back as a response
- cancellation while waiting (the circuit is released and a late reply finds no claim), timeout
- URL → call normalization
- native: the handler passes TCP circuits through the HTTPS client untouched, and delivers HTTPS circuits to the shared client
- duplicate install is refused
- browser (headless Chrome): the browser handler delivers HTTPS circuits to the shared client, and every `BrowserOnionProxy` shares the installed runtime's client (`Arc` identity)

Local: `cargo test -p rings-node --lib onion::` (175 passed); clippy `-D warnings` native and `browser_default` wasm32 `--tests`; `RUSTDOCFLAGS=-D warnings cargo doc -p rings-node`; frontend wasm check; `cargo +nightly fmt --check`; node browser suite `cargo test -p rings-node --target wasm32-unknown-unknown --features browser_default --no-default-features --lib` (12 passed).


zkjoie:
Self-review round addressed in c437b4c3, restructured at the boundaries instead of patched per finding:

1. **Target binding by construction.** `OnionHttpsCall` carries no target, and `begin` addresses it to `route.target`. The runtime `HttpsTargetMismatch` check and its variant are removed. Native callers can no longer pass un-normalized wire requests.
2. **Placement.** The native handle, handler and shared-resource wiring moved to `onion::native`. `onion::tcp` is TCP-only again and drops from 1055 to 891 lines, below the 1000-line limit.
3. **Disjoint client circuits.** The HTTPS client claims by `(circuit id, return peer)`, not by id alone. The collision bound is documented as a law on the native module.
4. **Consistent dispatch.** `claim` / `resolve` replaces the `Option`-encoded hand-back. A wrong-peer frame now falls through to TCP and gets TCP's normal unknown-frame error, and no ~240-byte payload is boxed or shuttled back.
5. **Duplicate type.** `OnionHttpsClientResponse` is removed in favour of `OnionHttpsResponse`.
6. **Closed channel.** It is now the named `OnionRouteError::HttpsResponseClosed`, and the doc states when it is reachable.
7. **Test gaps.**
   - Added a test that the native handler delivers HTTPS circuits to the shared client.
   - The `request` doc now states the timeout scope.
   - Browser remains compile-verified only: driving `BrowserOnionProxy` needs a wasm provider fixture.

The PR description is updated to match, including the public path move `onion::tcp::NativeOnionCircuitHandle` → `onion::native::NativeOnionCircuitHandle`.

zkjoie:
83f0e8f5 closes the browser test gap. It also corrects my previous comment: no new fixture was needed. `tests::wasm::{prepare_processor, new_provider}` already builds an IndexedDB-backed processor/provider in the browser, and CI's "Run node browser tests" step runs it under headless Chrome.

Added:
- `test_browser_circuit_handler_delivers_https_circuits_to_the_shared_client`: `BrowserOnionCircuitHandler` delivers an HTTPS circuit's backward payload to the shared client (the browser twin of the native handler test).
- `test_browser_proxies_send_through_the_installed_runtime_client`: two `BrowserOnionProxy` instances hold the same client `Arc` as the node's single installed onion runtime.

Local run: `cargo test -p rings-node --target wasm32-unknown-unknown --features browser_default --no-default-features --lib` in headless Chrome 153 → 12 passed, 0 failed (the whole node browser suite).

An end-to-end browser `request_http` test is still out of scope: route selection needs live relays and an exit in the directory, which the browser fixtures do not set up.

zkjoie:
Rebased onto master at 2341fc56 to resolve the conflict with #828. History is linear; the three commits are re-signed and now end at 2d4905f3.

- The conflict was in `BrowserOnionProxy::request_http`: #828 moved its inline timeout onto `rings_runtime::sleep`, and this PR moves that whole wait into the shared client. It is resolved in favour of the shared client.
- `OnionHttpsClient::request` now uses `rings_runtime::sleep(..).map_err(Error::Timer)`, because the removed `extension::transport::platform::sleep` no longer exists. This is folded into the first commit so every commit builds.

Re-verified on the rebased tree:
- `cargo test -p rings-node --lib onion::` (175 passed)
- clippy `-D warnings` native and `browser_default` wasm32 `--tests`
- node browser suite in headless Chrome (12 passed)
- `RUSTDOCFLAGS=-D warnings cargo doc -p rings-node`
- frontend wasm check

RyanKung (approved):

Issue #834, opened 2026-09-23
OnionMonad: onion circuits as client-sealed loops of registered operation symbols


RyanKung:
# OnionMonad: onion circuits as client-sealed loops of registered operation symbols

## Motivation

The onion data plane today interprets exactly two forward layers
(`crates/node/src/onion/circuit/mod.rs`, `OnionForwardLayer`):

- `Relay`: peel one layer, record a return edge in `relay_returns`, `SealAndSend` the inner layer;
- `Exit`: run the adapter selected by `OnionCircuitPayload::service` (`tcp`, `https`);

and answers along the same path reversed (`OnionWireMessage::Backward`).

This is a degenerate instance of a more general structure: a circuit is a term over a
signature of operation symbols, evaluated one symbol per hop, where `Relay` is the identity
symbol and `Exit` is a symbol that happens to be last. This issue generalises the circuit to
that structure (**OnionMonad**): any hop may evaluate a registered symbol (`add(a, b)`,
`jev(state, q)`) and hand the result on, the circuit is a loop that returns to the client
through fresh hops, and no hop learns another hop's arguments or results.

This is a **breaking change**: no compatibility with pre-change nodes, one total cutover
(no protocol versioning pre-1.0).

## Definitions

**D1 Signature.** `Σ` is a finite table of operation symbols, defined statically in code.
Each `f ∈ Σ` has the specification
`spec(f) = (name, Args_f, In_f, Out_f, W(f), L(f), pos(f), quota(f))`:

- `name ∈ OnionServiceName`. A name identifies the *backend*, not only the operation
  (e.g. `jev.typesafe`), so that a client can choose independent backends.
- `Args_f`: client-supplied constants; `In_f`: carried input; `Out_f`: carried output.
- `W(f) ∈ ℕ` — **output width**: the exact number of bytes `Out_f` is encoded into before
  sealing. Every result of `f` has width `W(f)` regardless of its value (e.g. `add` over
  u256: `W = 32`; `jev.typesafe`: answer padded to `W = 64`).
- `L(f)` — **latency class**: the exact delay between receiving an input and releasing the
  result (e.g. `L(add) = 10 ms`, `L(jev.typesafe) = 500 ms`). A fast evaluation waits; a slow
  one is dropped.
- `pos(f) ⊆ {Intermediate, WorldFacing}`. `WorldFacing` symbols (`tcp`, `https`) exchange
  bytes with the outside world and may hold a session; `Intermediate` symbols are one-shot.
[...]

PR #837, opened 2026-09-23
Add the Onion Kleisli Pipelines specification paper (#834)

RyanKung:
Companion spec paper for #834, titled *Onion Kleisli Pipelines* (renamed from OnionMonad: L6 excludes bind, so the exposed structure is a static composite in Kl(M), not a monad). It freezes D1–D9, L1–L10 and L5′ before the Phase 2 code. Docs only: no code change, no version bump. Supersedes #835, which was closed so the PR could be reopened under this account.

Adds `papers/onion-kleisli-pipeline.{tex,bib,pdf}` and a row in `papers/README.md`.

Contents:
- **Signature:** Σ as a signature and nodes as partial Σ-algebras. Registering any symbol implies registering `relay` (D1–D3).
- **Pipelines:** morphisms of the free category on Σ, denoted in Kl(M), with `relay` = η (D4a/b, L1, L2). At most `MAX_ONION_LOOP_SYMBOLS = 4` symbols.
- **Loop:** the guard-closed loop with `s = 2`. The guard counts as relay 1 of segment 0 and relay s of segment n (D4c, D5, L7). Also:
  - `H(n, s) = (s+1)n + s`, `H − 1` distinct DIDs, and `E = H + 1` cells;
  - `MAX_ONION_LOOP_HOPS = 14`, derived rather than written as a literal;
  - why s = 1 is rejected: the guard alone would link the client to h₁.
- **Layer and header:** a uniform layer inside a fixed-length Sphinx-style header of 14 layers. It uses per-hop PRG filler and client-precomputed MACs, and a correctness proposition shows every MAC verifies. L5′ (position hiding) is proved by a hybrid argument over the filler blocks.
- **AEZ carry:** the producer wraps s+1 layers, each relay removes one with τ=0, and the consumer checks τ=16 (D7). The AEZ v5 implementation is in-repo on `aes::hazmat` and uses raw 48-byte keys. Keys with a weak subkey are rejected, and `C ≥ 32`. L9 is load-bearing: it guarantees each key enciphers exactly one value.
- **SURBs:** credit invariant r(t) ≤ u(t) (D8).
- **Laws:**
  - L5 noninterference;
  - L6 path authority: the Arrow composition fragment is admitted and bind is excluded, while Selective needs L6 weakened to a client-sealed set of next hops;
  - L8 unlinkability, stated at the byte level;
  - L9 replay;
  - L10 wire.
- **Tables:** accepted leakage (the guard's RTT gives an estimate of H and ΣL(f_i)), loop shape at s = 2, threat → law, obligation → test, and the cutover phases.
- **Current design:**
  - `advance_backward` forwards one byte-identical client-bound ciphertext, so any two colluding relays can link their positions;
  - relays keep `relay_returns`;
  - nested layers grow by 404 B per downstream hop, which violates L5′.

Requirements the paper states beyond the issue text:
- L5's premise also fixes `from_i`, `e_i`, and the per-loop expiry `x`. L5′ additionally assumes that only the guard observes the client's departure time.
- `W` factors through the output type, so the `0x80` padding does not reveal the predecessor's symbol.
- The Selective combinator requires weakening L6 (see above). Its law is deferred to its own issue, together with Selective itself.
- Leakage row: a producer holds its outbound segment's keys, so a producer colluding with a relay of that segment can link the two positions by bytes. No value leaks.
- Open parameters: key transport (a per-layer envelope or one blinded element), raw DID/key encodings, σ, and `C ≥ σ|χ_υ|` when SURBs ride in the carry.

Build: `latexmk -xelatex`, 0 Overfull, no undefined citations.


RyanKung:
Pushed 00d371a3, which addresses the two remaining review points.

**Replay bounds (L9).** L9 now states the bounds the replay cache enforces today. A layer is admitted iff `arr < x ≤ arr + V`, with `V = ONION_FORWARD_MAX_VALIDITY_MS = 150 s`, and its tag is retained until `arr + V`. Each sender's partition holds at most `B = 4096` tags, across at most 64 senders. A full partition rejects the layer and never evicts a live tag. A new proposition derives:
- no admitted layer outlives its tag;
- `|R_i| ≤ 64B`;
- a per-neighbour admission rate of `B/V`, about 27 cells/s;
- a SURB lifetime of at most `V`.

The rate bound caps loop throughput per neighbour, including SURB replies. `B` is listed as a tuning constant, not a law.

**SURBs out of the carry (D8).** Under L5′, a SURB header must have the full header length, so a carried SURB would force `C > Ĥℓ` on every edge (L3) and at least double every cell. A session symbol now keeps the suffix of each forward loop as its SURB. Credit therefore arrives only as forward loops: one forward cell per reply cell, and the data is optional. A session idle for longer than `V` holds no credit. σ and `C ≥ σ|χ_υ|` are removed from the open parameters.

**Size correction.** Layer sizes now count the 48-byte AEZ keys decided in #834, not 32-byte keys:
- `ℓ(2,64) = 649 B` (≈ 531 B with raw DID and key encodings);
- `|β| = 9086 B`.

This still fits the 16 KiB bucket.

The algorithm, the obligations table, and the current-vs-here table are updated to match. Build: 0 Overfull, no undefined citations.

RyanKung:
Review of 580bb081 addressed in 00d371a3 and 21b4b184. What each finding became:

1. **Single carry width (HIGH):**
   - `C` is fixed per loop. The client picks a class `b` from today's cell buckets above 4 KiB (16 KiB to 12 MiB), with `C_b = b − |χ| − F`.
   - The class holds on every edge of the loop, SURB replies included (D6, L3).
   - `tcp`/`https` carry `Bytes<C₀` chunks, one per loop or reply; larger bodies span several loops.
   - `b` is added to `view_i`, to L5's `π_i`, and to L5′'s `π°`, and a "loop class b" row is added to Table II.
   - SURBs no longer ride the carry (00d371a3, Prop. "No SURB in the carry"), so `σ` is gone.
2. **Relay restart (HIGH):**
   - Every published symbol, `relay` included, carries a process epoch `e_n` drawn at each start (D2). A restart that empties `R_i` therefore rejects every older layer.
   - Prop. 4 now has a proof that uses this. A new threat row covers replay across a restart, and a new L9 obligation covers the same case.
3. **Expiry as a link token:**
   - `x` is one per loop and quantised to `Q = 30 s`.
   - A Table II row states that all hops of a loop and all loops sealed in one quantum share it.
   - The Table IV row "colluding relays match bytes" now says "up to the shared x".
   - D9 uses `x`.
4. **Key width:** fixed in 00d371a3 with 48-byte keys: `ℓ(2,64) = 649`, `|β| = 9086 B`. The issue's segment-analysis numbers are corrected in https://github.com/RingsNetwork/rings/issues/834#issuecomment-5802552904.
5. **Client demultiplexing:**
   - New Def. "Client position": the client is position `H+1`.
   - `λ_H` carries a fresh 128-bit tag in place of `α_{H+1}`. The client looks up the loop state by tag, accepts it once, and drops unknown tags. Every SURB suffix ends in its own tag.
   - Prop. 3 now covers `i = H`.
6. **L5′ guard exception:** the prose now names the guard (`from = ◇` or `next = ◇`) as the exception.
7. **`In = ()`:** `()` is encoded as `ε`, so every consumer authenticates `pad(ε)`. The "ignores the slot" rule is removed.
8. **Algorithm 1:** it gains the session branch (store the SURB, write to the world). New Algorithm 2 `Reply_h` handles SURB consumption and expiry. Prop. 4 is reworded (enciphered once by the producer, deciphered once by the holder) and proved.
9. **`W(relay)`:** removed. `relay` has no `W` and preserves `C`.
10. **Font fallback:** left as is. It is the same as `finger-convergence.tex`, and changing fonts for one paper would diverge from the house style.

Items 1–2 change D2, D6, L3, and the Phase 2a wire format. The #834 body is not edited; these are proposed there for the user's approval. Build: 0 Overfull, no undefined citations, all 7 pages rendered and checked.

RyanKung:
Pushed aed2a78f: D1 now states that Σ is closed, per the ruling in #834. A name outside Σ is rejected at configuration, at descriptor decode, and at route construction. Nothing falls back to `tcp`, and operator-named services are not accepted. The Phase 1 row in the cutover table notes that this changes configs, not wire bytes. I checked the paper against the updated #834 body: loop class, process epoch, client position, `pad(ε)`, SURBs kept out of the carry, same-class L5/L5′ premises with the guard exception, and the shared-`x` leakage row all match. Build: 0 Overfull, no undefined citations.

RyanKung:
Pushed 6c569b82. It addresses a full re-read of `develop...paper/onion-monad` at aed2a78f and the failing `typos` CI step.

**High**
1. **Session replies were outside L4.** L4 now names this as its one exception: a `WorldFacing` session reply leaves when world bytes are ready, at any time before the SURB's `x`. A Table II row records that reply timing follows the world.
2. **The expiry rule was unspecified, so the L5′ proof's claim about `x` was unsupported.** D6 now fixes `x = ⌈t_build/Q⌉·Q + X₀`, with `Q = 30 s` and `X₀ = V − Q = 120 s`, which is today's `ONION_FORWARD_PAYLOAD_TTL_MS`. `x` therefore depends only on the build quantum. Invariant 1 states that a SURB lives at most `V`.
3. **L5 claimed equality in distribution but rests on PRG/MAC.** L5 is now `≈_c`. Prop. 6 notes that exact equality holds for the carry alone.

**Medium**
4. **L9 did not cover a sender beyond the 64th.** It now takes the quota of the least recently active partition, and that partition's tags stay until their retention ends. Admission rejects while `|R_i| = 64B`. Prop. 9(v) adds the per-hop total bound `64B/V ≈ 1748` cells/s.
5. **The "least b" rule in L3 is circular for stream symbols.** L3 now applies it to fixed-width loops only, with `v₀` included. For stream loops, `b` is the client's choice between throughput and leakage, recorded in a Table II row.
6. **The L5′ proof omitted the carry and arrival time.** It now covers both.

**Low**
7. D9 drops the duplicate `bucket` field.
8. `F` is now defined as the cell framing.
9. D2 now reads "registration includes relay".
10. Algorithm 1 decodes with `enc⁻¹` before writing to the world. Algorithm 2's time parameter is renamed `t₀`.
11. Table III now says "no SURB credit".

**typos**
`typos.toml` gains three entries:
- `Tung`, for the author Viet Tung Hoang, under the proper-name words next to Brin and Pease;
- `SURBs`, as an identifier, because typos split the plural into `SUR`+`Bs`;
- a tex-only ignore for amsmath's `\begin/\end{multline}`.

`typos papers/ typos.toml` is clean locally.

Build: 0 Overfull, no undefined citations, all pages rendered and checked.

RyanKung:
Pushed 99ccf9a0, which renames the paper to *Onion Kleisli Pipelines*. The exposed structure is not a monad: L6 excludes bind, and a circuit is a static composite in Kl(M) (a path in the free category on Σ). `M` is only the effect inside one hop. The files move from `papers/onion-monad.*` to `papers/onion-kleisli-pipeline.*`. The title, the intro, and the sentence that justified the old name are updated, along with the README row, the build command, and the typos comment. The PR title and description now use the new name.

RyanKung:
Pushed c5532474. It addresses the second full re-read, at 99ccf9a0:

1. **Layer size.** `ℓ(s,A)` now counts the 16-byte next-hop MAC `γ` that every layer carries: `ℓ = 409 + A + 48(s+2)`, so `ℓ(2,64) = 665 B` and `|β| = 9310 B`, or about 547 B per layer with raw encodings. Table 1, the note after Prop. 5, and the open parameters are updated to match. The size correction on #834 (issuecomment-5802552904) is edited to include `γ` as well.
2. **Numbering.** The client position is now **D6′**, as in #834. It also states that the tag is padded to `|α|`, so `|χ_{H+1}| = |χ|` and the last cell is padded to `b`.
3. **Mutual citation.** Props. 6 and 7 previously took parts of their arguments from each other. The two arguments now live in Lemma 1 (Header view) and Lemma 2 (Relay carry), and both propositions cite those lemmas.
4. **Abstract.** It now names the onion Kleisli pipeline (not a monad) and says the slot has one width per loop.
5. **Leakage table.** The two class rows are merged into one.
6. **Cleanup.** The unused `corollary`/`obligation` environments and the `\LineComment` macro are removed. `papers/README.md` now lists `stmaryrd` and `multirow`.

Build: 0 Overfull, no undefined citations, all 7 pages rendered and checked. `typos` is clean.

RyanKung:
Pushed 02e5c17d, addressing the third full re-read (at c5532474):

1. **Exact carry claim.** Prop. 6 said the carry is equally distributed at every relay, but that was not true at the last relay of a segment: Assumption 1 only made AEZ ideal for τ = 0, and that relay sees `E^τ_{k_c}(b_k)`. Assumption 1 now models AEZ as an ideal enciphering scheme for τ = 0 and τ = 16 alike (`E^τ_k(m) = Π_k(m‖0^τ)`). Lemma 2 also states that it depends on L9.
2. **Functoriality.** D4b's functor was stated on `Free(Σ)`, but one application placed on two different nodes has two interpretations. The functor now lives on `Free(Σ×N)`, whose edges are placed applications `(f, ā, h)`. With that, `⟦P⋙P′⟧ = ⟦P⟧⋙⟦P′⟧` holds strictly.
3. **Blinded key transport.** D6′ now says where the client tag goes in that case: `λ_H` has no `α_{H+1}` field, so the tag occupies `λ_H`'s next-hop field instead.
4. **Stream chunks.** L3 now says they fill `C₀ − 1` bytes, matching Algorithm 2. The two L3 rows of the obligations table are merged into one.

Three places where the #834 body still differs from the paper have been forwarded to the session that maintains it:
- L6's claim that Selective composition is compatible;
- the `from_i` missing from L5's premise;
- the "least b" rule in L3 applied to stream loops.

Build: 0 Overfull, no undefined citations, pages rendered and checked, `typos` clean.

RyanKung:
Pushed 2fd6f9b0. It addresses the fourth full re-read, done at 02e5c17d.

1. **L5 scope.** L5 was stated unconditionally, but L4 exempts session replies. On a session symbol's reply path, arrival times follow the world and can depend on the session hop's `ā`, for example the target host. L5 now excludes those arrival times from `view_i`. The residual is Table II's session-reply row. Prop. 6's timing step says the same. I have asked the #834 body owner to add this scope to L5.
2. **D6′ sentence.** 02e5c17d attached "padded to `b`" to the blinded-key-transport clause only. Padding now applies to both variants, and the blinded variant has its own sentence.
3. **Introduction.** The name "onion Kleisli pipeline" now attaches to the general structure. Before, it read as naming today's circuit.
4. **Threat table.** The position row uses `C_b`.

Build: 0 Overfull, no undefined citations. The changed pages are rendered and checked, and `typos` is clean.

RyanKung:
Pushed 0617f7c0. It addresses the review of 02e5c17d (issuecomment-5803108280), rebased onto 2fd6f9b0.

**High**
1. **SURB single use.** Algorithm 2 now removes `υ` from the pool before release (`Q_{h,ς} ← Q_{h,ς} ∖ {υ}`, "single use"). Inv. 1 says each reply removes one `υ`, and the credit obligation checks the removal.
2. **Sessions.** New **D2′**: a session symbol (`tcp`, `https`) carries `(ς, target)` in `Args`, where `ς` is a 128-bit session id drawn by the client, and has type `Frame`, with `enc(data(w)) = 00‖w` and `enc(fin) = 01`.
   - `h` opens `ς` at its first forward loop, binds it to the target, and rejects a later loop of `ς` with another target.
   - `data(ε)` is a credit-only loop.
   - `fin` closes one direction. `h` drops `ς` once both directions are closed, or after `V` without a forward loop.

   Pools are now `Q_{h,ς}`: D8, Algorithm 1's session branch, Algorithm 2 (`Reply_{h,ς}`, which sends `fin` at EOF) and Inv. 1 are all per session. A Table II row records that `ς` is seen only by `h`. The credit obligation now requires that two sessions never share a SURB.
3. **L5′.**
   - `x` is added to `π°`.
   - The law now bounds the advantage by `ε_hdr + |D_i − D′_{i′}|/Q`, where `D_i` is the delay from the client to position `i`.
   - Assumption 1 states that the build phase `t_build mod Q` is uniform, and that no hop other than `g` observes `t_build` beyond what `x` reveals.
   - Prop. 7 derives the timing term and shows it is tight: the offset is `u + D_i` with `u` uniform on `[0,Q)`, the statistical distance is `min(1, |ΔD|/Q)`, and a threshold test attains it. A note explains why a uniform departure delay cannot remove the term.
   - Table II gains the row "`x` locates departure within `Q`".

**Medium**
4. **D6′.**
   - `nxt_H` is the client's DID, whose connection `g` already holds, in both key-transport variants.
   - In the blinded variant, the tag gets its own 16-byte field in every layer (uniform elsewhere), so `ℓ` stays uniform.
5. **`W`.** Only fixed-width symbols have `W`. A stream symbol, like `relay`, has none. Its frames carry at most `C₀ − 2` data bytes, and a frame with `|enc| ≥ C₀` fails closed. The L3, threat, and obligation rows are updated to match.

**Low**
6. **Free category.** `Free(G_Σ)` is on the application graph, and D4b uses `Free(G^N_Σ)` over placed applications. L2 is updated to match.
7. **`γ_{H+1}`** is uniform and ignored, because `k_{c_n}` authenticates the carry.
8. **Tag table.** `T` entries are removed on first use or at `x`. The client obligation now includes "expired tag dropped; `T` empty after every `x` passes".
9. **Value index.** The consumed value is `v̂_i` (`v_{k−1}` at `h_k`, `⊥` at a relay), and every other value is a segment value `v_k`. D9, L5, L5′, L6, Props. 6–7 and the L5 obligation are updated.
10. **Guard quota.** The L9 note now states that the guard's per-sender quota caps a client's total forward rate at `B/V` cells/s. Through Inv. 1 it also caps the client's total reply rate at `B·C_b/V`: about 180 KB/s at `b = 16` KiB, counting carry bytes rather than cell bytes. This, not relay-to-relay traffic, is what constrains `B`.

Build: 0 Overfull, no undefined references, 8 pages rendered and checked. The `TU/ptm` font fallback is unchanged. `typos` is clean. D2′ changes the session-symbol `Args` and the Phase 2a wire format. It still needs to go into the #834 body, which the peer session maintains; I have not edited it.

RyanKung:
Pushed b8abcbd3. D2′ now states that `https` frames carry the HTTPS request/response encoding and that TLS tunnels use `tcp`. This is the #839 ruling, which the #834 body already records in its D2′. The #834 body now matches the paper at 0617f7c0 plus this sentence. Build: 0 Overfull, no undefined references, `typos` clean.

RyanKung:
Pushed 36ff1888, which brings the paper in line with the #834 throughput redesign. The target is ≥ 1 MB/s of download per client at the 16 KiB class. It is rebased on 0617f7c0 and b8abcbd3 (session-keyed SURBs, the L5′ timing term, D6′, https scope), and those changes are kept.

- **D6″ header encoding.** Key transport is Sphinx's single blinded group element, DIDs are raw 20-byte, and carry keys travel as 32-byte seeds. This gives `ℓ = 221 B`, `|β| = Ĥℓ = 3094 B`, `|χ| = 3143 B`, and `C_16KiB ≈ 13.2 KB`. Table 1 and the header definition are updated to match, and the per-layer-envelope variant is removed.
- **D7 seeds.** Each segment has one seed `σ_k`. Relay and consumer seeds, and their 48-byte AEZ keys, are derived by domain-separated HKDF-SHA256. A weak derived key forces a re-draw of `σ_k`. Every layer carries exactly one inbound and one outbound seed. The AEZ API still takes raw 48-byte keys only.
- **D2′ and D8.** `Frame` gains `credit(υ₁..υ_k) = 02‖υ₁‖…‖υ_k`. **This constructor is not yet in the #834 body's D2′.** The new Prop. "SURB batching" replaces "No SURB in the carry": `|υ| = 20 + |χ| + 32 + 8 = 3203 B`, and `k = ⌊(C₀−2)/|υ|⌋`. The credit invariant now counts batched SURBs per session.
- **Value of `k`.** With `|υ| = 3203 B`, `k = 4` at 16 KiB for framing `F ≤ 411 B`, and `k ≥ 3` for `F ≤ 3.6 KB`. The issue's `k = 3` uses `|υ| ≈ 3.3 KB`, so it is the conservative bound. Neither value changes the throughput result, which does not depend on `k`.
- **L9 replay store.**
  - `R_i[x]` is a family of Bloom filters keyed by the quantised `x`; at most 6 are live.
  - There are no false negatives, so admission is at most once exactly. False positives are safe drops at ≤ 2⁻²⁰, and the store takes ≈ 3.8 MB.
  - `B = 16384`, with a global budget of `64B`. This admission replaces `MAX_ONION_CRYPTO_*`. Prop. "Replay bounds" is rewritten.
- **New Prop. "Throughput".** A single client's download is at most `min(B/V, (k+1)ρ_f)·(C₀−2)`. At 16 KiB that is about 109 cells/s × 13.2 KB ≈ 1.44 MB/s, which meets the ≥ 1 MB/s target. The ≈ 180 KB/s client-cap paragraph is replaced: with batching, the reply path binds, not the client's credit.
- **Other updates.**
  - The Assumption adds HKDF (as a PRF) and DDH.
  - The obligations table gains rows for the loopback ≥ 1 MB/s benchmark, the replay store (no false negatives across rotation), and seed derivation.
  - The phases row and the parameters are updated. No parameter remains open.

Build: 0 Overfull, no undefined citations, all 9 pages rendered and checked, `typos` clean.

RyanKung:
Pushed 63817ac2, which fixes the findings from the re-read at 36ff1888. The #834 body is updated to match: L9 replay-store memory, the D6′ SURB tag, and the replay-store test.

1. **Replay-store memory.** The old 3.8 MB assumed one pre-sized filter. A filter keyed by `x` can receive up to `64B` tags over its lifetime, so pre-sizing all six live filters for the worst case would take about 23 MB. The replay store now works as follows:
   - Each `R_i[x]` grows in blocks of `B` tags, each block at false-positive rate 2⁻²⁶ (Bloom 1970; Almeida et al. 2007).
   - A filter only receives arrivals in `[x − V, x)`, so under the global cap it holds at most `64B` tags. That bounds it to 64 blocks and a false-positive rate of at most 2⁻²⁰.
   - Memory therefore tracks admitted tags. It is at most about 5.4 MB: 70 blocks of about 76.8 KB, at about 37.5 bits per tag.
2. **Lemma 1.** Hop `i` computes `α_{i+1}` itself. What DDH guarantees is that `α_i` is uniform to `i` and unlinkable to the other hops' `α_j`. The proof now states that.
3. **D6′.** Every SURB ends in its own tag: a forward loop's suffix ends in that loop's tag, and a batched SURB ends in a fresh one.
4. **Throughput and per-neighbour budget.**
   - Prop. "Throughput" is now stated for a client alone on its links.
   - Its proof also covers the forward credit cells: they need about 22–27 cells/s, well under `B/V`.
   - Prop. 9(iii) now says the per-neighbour budget is shared by every loop on that link.

Build: 0 Overfull, no undefined citations, changed pages rendered and checked, `typos` clean.

RyanKung:
Pushed 8e957aff, which fixes the two findings from the re-read at 63817ac2. The #834 body has the matching changes (L9 admission, D8 pool cap, Tests).

1. **Class-weighted admission (L9).** Budgets now count 16 KiB units, and a class-`b` cell costs `b/16 KiB` units, the same way today's cell buckets charge the full visible size.
   - **Why:** the byte-budgeted crypto limiter is being replaced. With a plain cell count, a neighbour could push 12 MiB cells at 109 cells/s.
   - **Bounds:** per neighbour, cell traffic is at most ≈ 1.79 MB/s; the global cap is ≈ 115 MB/s.
   - **Unchanged at 16 KiB:** a cell there is one unit, so throughput is still ≈ 1.44 MB/s.
   - **Larger classes:** they approach 1.79 MB/s as the header's share falls. Prop. "Throughput" states this.
2. **SURB pool cap (D8, Inv. 1).** A session's pool holds at most `Q_max = 256` SURBs (≈ 820 KB).
   - **Why:** without a cap, one client could make `h` hold `(k+1)B` SURBs, about 260 MB.
   - **Enforcement:** credit beyond the cap is dropped, and the pools of all sessions at `h` count against `quota(f)`.
   - **When it binds:** `Q_max` only binds when the loop round trip exceeds `Q_max·V/B ≈ 2.3 s`.

The threat table gains rows for large-class floods and SURB-pool exhaustion. The credit obligation now says reading resumes on new credit.

Build: 0 Overfull, no undefined citations, changed pages rendered and checked, `typos` clean.

Issue #841, opened 2026-09-23
Onion Kleisli Phase 2a-2: x-keyed Bloom replay store and unified admission (#834)


RyanKung:
Part of #834, Phase 2a, sub-PR 2 of 4. The spec is the #834 body (L9) and `papers/onion-kleisli-pipeline.tex` on `develop`.

## Scope

A pure admission module. It is not yet wired into the live data plane.

1. **Replay store (L9).** `R_i` is a set of Bloom filters keyed by the quantised expiry `x`.
   - A tag goes into the filter of its `x`, and a whole filter is dropped when `x` passes; at most `V/Q + 1 = 6` filters are live.
   - Filters grow in blocks of `B` tags at a per-block false-positive rate of 2⁻²⁶, capped by the global admission, so each filter's rate is ≤ 2⁻²⁰.
   - Memory is bounded as stated in #834 L9 (≈ 5.4 MB).
   - There are no false negatives: at-most-once admission is exact.
2. **Admission.**
   - The admission window is `arr < x ≤ arr + V` with `V = 150 s`.
   - The epoch check rejects layers of an earlier process epoch (D2).
   - Per-sender capacity is `B = 16384` over at most 64 senders, with a global cap `64·B/V`.
3. **One admission authority.** The module is designed to replace `circuit/limiter.rs` (`MAX_ONION_CRYPTO_*`, a 60 s window, 8192 ops global). The replacement itself happens in 2a-4. Here, only the pure step and its state are delivered.

## Laws to discharge (tests)

- No replayed tag is ever admitted, including across filter rotation and at `x` boundaries.
- A filter is dropped exactly when its `x` passes.
- The measured false-positive rate is within target, checked with a seeded property test.
- The per-sender and global caps hold, and the least-recently-active partition is recycled as specified.
- Restart: a new epoch rejects every layer of the old one.

Time is injected: the step is pure over `(state, now, tag, x, epoch, from)`.

## Out of scope

Header and carry primitives (2a-1), registration and route (2a-3), and data-plane integration (2a-4).

Issue #843, opened 2026-09-23
Onion Kleisli Phase 2a-4: wire cutover to the loop data plane (#834)


RyanKung:
Part of #834, Phase 2a, sub-PR 4 of 4: **the wire cutover**. It depends on the 2a-1, 2a-2 and 2a-3 PRs. The spec is the #834 body and `papers/onion-kleisli-pipeline.tex` on `develop`.

## Scope

1. The reducer and shell run the loop data plane over the uniform layer (D6), the fixed header (2a-1), and the carry segments (2a-1).
2. Stateless relays: delete `relay_returns`, `RelayReturnKey/Edge/Entry`, `remember_return_hop`, `advance_backward`, `OnionWireMessage::Backward`, `OnionBackwardFrame`, `MAX_ONION_RELAY_CIRCUITS` and `ONION_RELAY_RETURN_TTL_MS`. `OnionCircuitState` holds only the replay store from 2a-2.
3. The admission step from 2a-2 replaces `circuit/limiter.rs`.
4. The client is position `H + 1`, with tag `t_⋄` (D6′).
5. Session symbols (D2′):
   - frames `data`, `fin`, `credit`;
   - per-session SURB pools `Q_{h,ς}`;
   - SURB batching (D8);
   - `tcp` pauses reading on exhausted credit.
   The `data` frame reserves a flags byte for message-oriented symbols (the `wss` follow-up), so that adding them needs no second cutover.
6. Loop class `b` selection by the client (L3).
7. The exit's per-reply signature is dropped, since the AEZ carry under `k_c` authenticates replies (to be confirmed in #834).
8. **Pending decision:** the `https` tunnel question. Phase 1 kept CONNECT tunnels under `https`, dispatched by `⟦fetch⟧ <|> ⟦tcp⟧`, with a known byte overlap. Before this PR, #834 must settle between (a) tunnels move to `tcp` and the `<|>` is deleted (with a `*:443` port wildcard in exit policy), and (b) a wire tag distinguishing fetch from tunnel.
9. Replace the golden set: the Phase 1 golden tests are replaced by one pinned set for the new wire.

## Tests

The #834 Tests section, plus a loopback benchmark: ≥ 1 MB/s single-client `tcp` download at `b = 16 KiB`, with upload ≈ 1/(k+1) of download.

The version bump is **not** in this PR; it lands on the last PR of the cutover (Phase 2b).

PR #844, opened 2026-09-23
Onion Kleisli Phase 2a-2: x-keyed Bloom replay store and unified admission (#834)

RyanKung:
Closes #841
Part of #834

## Summary

This PR adds the pure L9 admission step for #834 Phase 2a-2 in `onion::circuit::admission`. It is not yet wired into the data plane; 2a-4 does that and replaces `circuit/limiter.rs`.

The step is `δ : S × Time × I → S × (1 + Rejection)` over `(state, now, from, epoch, x, ν, units)`. It has no effects, and time and the probe key are injected. It admits a layer iff all of these hold:

- `e = epoch_i`
- `arr < x ≤ arr + V`
- `ν ∉ R_i[x]`
- the sender budget (`B`) and the global budget (`64·B`) hold

It then commits the tag and both charges together. Every rejection commits nothing.

- **Replay store (`bloom.rs`).** `R_i` maps `x` to a sliced Bloom filter (Almeida 2007): `K = 26` slices of `M = 23 680` bits, blocks of `B` tags at a rate `≤ 2⁻²⁶`, and ≈ 76.96 KB per block. A filter is a chain of full sealed blocks plus one open block, so an empty filter cannot be represented. A filter is dropped as a whole once `x ≤ now`.
  - The probe addresses are 26 independent 32-bit words of `Keccak-256(κ ‖ ν ‖ i)`, `i = 0..3`, under a process-secret key `κ`.
  - Double hashing was tried first and rejected. It has only `M² ≈ 2²⁹` probe patterns, which puts a false-positive floor of `n/M² ≈ 2⁻⁹` at `64·B` tags. The seeded saturation test caught this.
- **Budgets (`ledger.rs`).** Unit ledgers over a sliding window of the `V/Q = 5` aligned arrival quanta, with per-sender `B = 16384` units and global `64·B`. There are at most 64 sender partitions. The least recently active partition is recycled with a fresh ledger (ties go to the least DID), and the global cap bounds identity rotation.
- **Expiry and clock.** `OnionExpiry` can only be built on the `Q` grid. The clock is monotone (`now := max(now, clock)`), so a rollback cannot refund budget or bring back a dropped filter.

## Spec readings (confirmed with the #834 dispatcher)

1. The step also takes nonzero `units`, because paper L9 charges a class-`b` cell `b / 16 KiB` units. Tags are counted per cell and budgets per unit, so the tag count is at most the unit count and the memory bound follows. 2a-4 maps class to units.
2. "Per `V`" means a sliding window over 5 aligned quanta. From `arr < x ≤ arr + V` with `x = kQ`, `R_i[x]`'s arrivals fall exactly in quanta `{k−5, …, k−1}`, so `|R_i[x]| ≤ 64·B` holds exactly. A token bucket would allow 2×.
3. Recycling gives the new sender a fresh ledger, and the global cap is the Sybil bound.
4. Because `x` is on the grid, at most `V/Q = 5` filters are live (the paper's bound is 6), and memory is at most `64 + 5` blocks ≈ 5.3 MB.
5. The clock is monotone.

## Tests (`admission/tests.rs`, seeded, injected time)

- A seeded 20 000-step property test (random tags, senders, `x` offsets around both window edges, and clock rollbacks) checks every verdict against the reference model "window ∧ unseen `(x, ν)`". It also checks that at most 5 filters are live and that none is past its `x`.
- A filter is dropped exactly at its `x`, with replays checked at `arr = x − V`, `arr = x − 1` and `arr = x`.
- A forward clock jump followed by a rollback never admits a replay.
- After a restart (new epoch), every layer of the old epoch is rejected and nothing is committed.
- The per-sender budget holds across the sliding window, with the refill at exactly the fifth quantum.
- The global budget bounds identity rotation, and a rejected request recycles no partition.
- LRU partition recycling works, including DID tie-breaks and the fresh ledger.
- A block-geometry test checks `(1 − e^{−B/M})ᴷ ≤ 2⁻²⁶`.
- Seeded saturation: 64 senders × `B` one-unit cells fill one filter through admission.
  - It ends with exactly 64 blocks, and the next cell is rejected by the global budget.
  - The measured rates are ≤ 1.25·2⁻²⁶ per block and ≤ 2⁻²⁰ per filter; the observed filter rate is 0.969·2⁻²⁰.
  - 2¹⁶ fresh probes are also checked against the filter.
- A filter grows by whole blocks, and no inserted tag is missed.

Note: the saturation test hashes 2²⁰ tags and takes about 2 minutes in a local debug build. CI runs native tests under the optimised `ci` profile.

## Notes

- Until 2a-4 wires the module in, it is declared with `#[cfg_attr(not(all(test, rings_native)), expect(dead_code, reason = …))]`. `expect` will fire `unfulfilled_lint_expectations` once the module is used, so the attribute has to be removed then.
- The replay tag reuses `OnionForwardNonce` and the epoch reuses `OnionExitEpoch`. No wire change.


RyanKung:
Addressed in ddd07794, which answers the review of 7721fca6 plus two spec updates now in #834 L9.

### Spec changes (#834 L9)

- **Sender ledgers replace the 64 LRU partitions. This supersedes the reviewed "recycle only idle partitions", which would lock every neighbour out once 64 senders are loaded.**
  - A ledger is keyed by the DID of an authenticated link. `link_opened` creates it, or reopens the existing one without resetting it.
  - A ledger is released only after `link_closed` **and** once its window load is zero. There is no recycling and no LRU.
  - The table holds `2·R` ledgers. `R` (the connection-registry capacity) is a constructor parameter, and 2a-4 wires the real value.
  - A full table fails closed with `SenderTableFull`. A cell from a DID without a ledger is rejected with `UnlinkedSender`.
  - `G = 64·B` is an independent constant.
  - New tests:
    - `test_closed_link_keeps_its_ledger_until_it_drains` is the evict-and-return case: spend `B`, close the link, churn 38 DIDs, reopen, and the cell is still rejected with `SenderBudget`.
    - `test_more_than_64_live_links_are_all_admitted`
    - `test_full_table_rejects_a_new_link_until_a_closed_ledger_drains`
    - `test_rotation_through_many_dids_never_exceeds_the_sender_budget`, a seeded churn property. It checks that no DID is charged more than `B` over any 5 consecutive quanta and that the table stays within `2·R`. It asserts that every rejection kind occurred.
- **Charge once.**
  - Every cell whose key was computed is charged `u(b)` exactly once: in `admit` (charge, then epoch, window and replay checks, then insert), or in `charge_invalid` on a `γ` failure, which leaves the replay store untouched.
  - Replayed, expired and stale-epoch cells pay too.
  - `headroom()` is the pure pre-ECDH query. It shares one rule (`charges`) with the charge itself.
  - A budget rejection charges nothing.
  - Covered by `test_every_computed_cell_is_charged_exactly_once`.

### High

- **H1.** The walk now rolls back with probability 1/64 and a magnitude of `0..=V`, and the test asserts `clock ≥ origin + 100·V`, so it cannot silently degrade again.

### Medium

- **M1.** The doc and `test_block_geometry_meets_the_block_rate` now use the exact bound `(1 − (1 − 1/M)^B)^K ≤ 2⁻²⁶`, which holds iff `M ≥ 23 637.8`. With independent slices, this is the exact expected block rate.
- **M2.** The Clock law now also states the liveness cost: a forward jump of `Δ ≥ V` pins the clock. It specifies the 2a-4 duty: on `now < clock − V`, start a fresh state under a fresh process epoch.
- **M3.**
  - `OnionExpiry::admissible_at` is used by admission and by `reducer.rs`. The reducer therefore now also rejects an off-grid expiry, which no honest client builds.
  - The room predicate is gone with recycling. The recycling decision is now the named `SenderLedger::is_releasable_at`.

### Low

1. Probes store `u16` addresses; the narrowing is const-asserted `M ≤ 2¹⁶`. Blocks are typed `Box<[[u64; W]; K]>`, so the zip over slices cannot truncate. `get` remains only because the index is a runtime value. Both the query and the join now skip the same dead arm, so `insert ; contains = true` holds unconditionally (documented).
2. The ledger picks its cell by a total fold over `[head, tail @ ..]`. The post-check sum is `saturating_add`, proven exact. `None` now means only the domain rejection.
3. `const _: () = assert!(M == 64·W, M ≤ 2¹⁶, 4·K ≤ 32·DIGESTS)` is added.
4. The doc is reworded: an empty filter is unrepresentable, and a non-full sealed block is representable but not reachable.
5. `OnionReplayFilterKey` is no longer `Clone`. It is zeroised on drop, and so are the preimage and stream buffers (`zeroize` is now a rings-node dependency; it was already in the lockfile).
6. The Budgets law now states the window precisely: `≤ B` over any interval of at most `4Q`, and up to `2·B` over an interval just over `4Q`.
7. The doc now says filters are dropped by the first step at or after `x`. An idle hop keeps them within the memory bound, and 2a-4 decides whether to run a timer.
8. The probe sample is reworded as a smoke check. The new `test_all_live_filters_stay_within_the_memory_bound` saturates `G` across all five expiries and checks at most `64 + 5` blocks.
9. Covered by the evict-and-return test above.
10. The doc comment is added.
11. `OnionExpiry` is now a quantum index. `OnionExpiry::of_build` is the single source of the grid, and it replaces `crypto.rs::quantized_forward_expiry` (same values, tested). The epoch and `ν` renames stay with 2a-3 and 2a-1.
12. Items are `pub(super)`, and the key is `pub(in crate::onion::circuit)`.
13. The PR body is corrected: the `expect(dead_code)` stays satisfied while any item is unused, so 2a-4 must remove it explicitly (#843).

Gates run locally:
- clippy `-D warnings` for native, `--all --tests`, browser `--tests`, browser lib and ffi;
- `cargo +nightly fmt --check`;
- rustdoc `-D warnings`;
- `taplo`;
- `typos`;
- core `--features dummy`.

Tests run: `onion` (220 tests, 17 of them admission, under `--profile ci`).

RyanKung:
Round 2 is addressed in b53c13e3. It also includes the charge-on-receipt clarification now in #834 L9.

### High

- **H1: charge before the ECDH, settled by a linear token.**
  - `headroom` and `charge_invalid` are gone.
  - `charge(now, from, u) -> Result<OnionAdmissionCharge, OnionBudgetRejection>` commits the charge before `α` is checked and the ECDH runs.
  - `OnionAdmissionCharge(())` is not `Clone` or `Copy`, is `#[must_use]`, and can only be built by `charge`.
  - `admit(now, token, layer)` consumes the token. Dropping it settles an invalid `α` or a failed `γ`.
  - "Exactly once" is therefore structural: there is no admission without a paid token, and one token cannot be used twice.
  - `test_every_computed_cell_is_charged_exactly_once` now includes the pipelining case: with 1 unit of headroom left, 8 in-flight cells produce one token and 7 `SenderBudget` refusals, so the ledger ends at exactly `B`.
- **H2: a table refusal is a link refusal.**
  - `link_opened` returns `OnionLinkTableFull`. The step doc states that the shell closes the link (fail closed) and the peer redials as a new link event.
  - `test_full_table_refuses_the_link_and_the_peer_redials` checks four things:
    - the shell's `link_closed` on a refused DID is a no-op;
    - that link's cells are `UnlinkedSender`, so they are never decrypted;
    - a redial is still refused while the closed ledger carries load;
    - the redial succeeds after the ledger drains and is swept.

### Medium

1. **Link generations.**
   - The ledger now counts open links (`open_links: u32`). `open(g₁) open(g₂) close(g₁)` leaves it open, and a close without a matching open is a no-op. This makes it correct whether or not core pairs generations. Covered by `test_ledger_counts_interleaved_link_generations`.
   - The shell precondition (`link_opened` on `Connected`, before any cells) is stated. Cells that arrive earlier are `UnlinkedSender` and are never decrypted.
2. **Prop. Ledger as an invariant.**
   - `advance` sweeps releasable ledgers (no open link and zero load) whenever a step enters a new quantum, and `link_closed` releases a drained ledger immediately.
   - Invariant: after every step, no ledger is releasable. A load can only reach zero when the quantum advances.
   - Covered by `test_drained_closed_ledgers_are_swept_on_the_next_quantum`, and asserted after every step of the churn property.
3. **Reducer change reverted.** `reducer.rs` is byte-identical to `develop`; the change is left to 2a-4 (#843). The two exit tests now use on-grid instants:
   - expired: `x = arr = Q`;
   - overlong: `arr = Q`, `x = arr + V + Q`.
4. **Rejection types split.** `OnionBudgetRejection {UnlinkedSender, SenderBudget, GlobalBudget}` comes from `charge`, with nothing charged. `OnionAdmissionRejection {StaleEpoch, OutsideWindow, Replayed}` comes from `admit`, with the charge kept. `OnionLinkTableFull` comes from `link_opened`. Each function returns exactly its own type.

### Low

1. `charge` does one `get_mut` and commits through that reference. The second lookup is gone.
2. The liveness doc is corrected:
   - failures begin at `Δ ≥ X₀ = 4Q`, and all honest layers fail at `Δ ≥ V`;
   - the 2a-4 trigger is `now < clock − X₀`;
   - it states that switching epochs invalidates every loop in flight.
3. The stale "V/Q + 1 = 6" is removed.
4. Reworded: `1 − e^{−n/M}` is a lower bound on `p_n`, so it cannot certify an upper bound on the rate.
5. `ReplayProbe` now has no derives (no `Debug`, `Copy` or `Clone`) and is used by reference.
6. The churn test now opens, sends and closes each churner with load under `R = 8`. It asserts that some links are refused and that the table stays at most `2·R`. After the window, only the spender remains.
7. The memory test now runs sustained saturation over ten quanta: every DID sends B/5 per quantum across the five expiries. It checks at most `64 + live` blocks after each quantum.
8. `2 * R` replaces the literal `32`.
9. `OnionExpiry` (struct, `of_build`, `as_ms`) now lives in `circuit/mod.rs` next to `ONION_FORWARD_EXPIRY_QUANTUM_MS`, so live code no longer depends on the dead-code module. The admission algebra (`from_ms`, `admissible_at`, `has_passed_at`) stays as an inherent `impl` in `admission`. Parsing the layer's `expires_at_ms` into it once changes what the live exit accepts, so that goes with M3 into 2a-4 (#843).
10. Both notes are now in the module laws:
    - **Skew:** there is no tolerance; a hop `δ` off rejects about `δ/Q` of loops.
    - **Budgets:** `G` is one pool shared with `γ`-failing cover cells.
11. `zeroize = "1.8.1"` is now a workspace dependency, used by `core`, `aez` (with `zeroize_derive`) and `node`.

### Gates
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, aez wasm, and ffi.
- `cargo +nightly fmt --check`.
- rustdoc `-D warnings`.
- `taplo` and `typos`.
- core `--features dummy`.
- Tests: `onion` (222, 19 of them admission) under `--profile ci`.

RyanKung:
Round 3 is addressed in e3b870af.

### High

- **H1: the window is judged at the charge's arrival.**
  - `OnionAdmissionCharge` now records `arrival_ms` (the monotone clock at `charge`) and the charging state's `epoch`.
  - `admit` requires `token.arr < x ≤ token.arr + V` and `x > clock`: the second condition keeps the filter live, so a late admission never writes a dropped filter. It also requires `token.epoch = layer.epoch = epoch_i`.
  - The charging quanta of `R_i[x]` are therefore exactly `{k − 5, …, k − 1}` however late the ECDH completes, so `G ≥ |R_i[x]|` holds exactly again.
  - `test_the_window_is_judged_at_the_charge_instant` covers both mismatches:
    - a cell charged with `x = arr + V + Q` is still `OutsideWindow` when admitted one quantum later;
    - a cell charged in window but admitted at `x` is rejected and leaves no filter.
  - `test_a_token_from_another_epoch_admits_nothing` checks the epoch binding.

### Medium

1. **"By construction" is reworded.** The Charging law now claims only what the types give: admission requires a paid token, and one token admits at most once. It states that the token is not bound to its cell or its units and that peeling does not require it. Those are listed as 2a-4 shell obligations: charge every received cell once, with its class's units, before peeling. I kept the rewording rather than adding `Charged<Cell>`, because the cell, class and peel types belong to 2a-1 and 2a-4.
2. **Affine link handle.**
   - `link_opened` returns `OnionAdmissionLink` (affine; only `link_opened` builds it), and `link_closed(now, link)` consumes it. A refused link has no handle and cannot decrement, so the count equals the number of live handles under any interleaving.
   - `test_a_refused_generation_cannot_close_a_live_one` plays the reported trace: `open(g₁)` is refused, the table drains, `open(g₂)` succeeds, and `g₂` is still admitted across three more windows. A late close for `g₁` is not expressible.
   - The "whether or not core pairs" claim is gone.

### Low

1. `OnionExpiry` moves to `circuit/expiry.rs`, and its field is private to that module. It has two entry points, `of_build` and the parser `from_ms`, and `from_ms` is visible to all of `circuit`, including the shell. `from_ms` has a narrow `expect(dead_code)` until 2a-4 uses it. Tests build expiries through `from_ms`. The `admit` flowchart notes the off-grid `None` drop, and `Hash` is removed.
2. The saturation doc is corrected: `as_ms` saturates only past `u128::MAX / Q` quanta, which no millisecond clock reaches, and `from_ms ∘ as_ms = Some` is stated for reachable expiries.
3. "Linear" → "affine" everywhere. Dropping the token is a valid settlement.
4. The memory test now saturates. Every quantum, each DID sends until its charge is refused. The test asserts `load_global = G` after every quantum and exactly 640 refusals, so the `Unpaid` branch is live.
5. Both exit tests are at the boundary, with a control that pins the reason (the reducer stays byte-identical to develop, so the reason is not observable directly):
   - overlong: `arr = Q − 1`, `x = 6Q = arr + V + 1` is rejected, and `x = 5Q` is admitted;
   - expired: `x = arr = Q` is rejected, and `x = 2Q` is admitted.
6. Covered by the new instant-mismatch test above.
7. The redial is stated as a liveness *assumption*: a churner at rate `R/V` can win the freed slots, and there is no fairness.
8. The doc now says "the paper's hop algorithm, with the charge taken on receipt (#834 L9)". Thanks for amending Algorithm 1.

### Gates
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, and ffi.
- `cargo +nightly fmt --check`.
- rustdoc `-D warnings`.
- `taplo` and `typos`.
- core `--features dummy`.
- Tests: `onion`, with 22 admission tests, under `--profile ci`.

RyanKung:
Round 4 is addressed in 4c35aa6f, implementing the #834 L9 decision (links as core generations) and the `renewed` rule.

### Medium 1–3: links are core generations, not a Rust handle

- `OnionAdmissionLink { did, generation: u64 }` is plain `Copy` data. The 2a-4 shell feeds it from core's `Admitted`/`Retired` events. Core does not expose the generation to the application today (`SwarmEvent` carries only the DID), so 2a-4 has to surface it.
- State:
  - one `SenderLedger { ledger, live: BTreeSet<u64> }` per DID with a live link or with load;
  - the table holds at most `2·R` ledgers and at most `2·R` live links (`live_links`).
- `link_opened(now, link) -> Result<(), OnionLinkTableFull>`. Opening a link that is already live is a no-op.
- `link_closed(now, link)` is **idempotent**: closing an unknown, refused or already-closed generation changes nothing. The ledger is released at once if that leaves it releasable, meaning no live generation and zero load. The per-quantum sweep keeps the no-releasable-ledger invariant.
- `charge(now, &link, u)` requires that exact generation to be live. There is one lookup, and its only `None` is the domain case `OnionBudgetRejection::LinkNotLive` (which replaces `UnlinkedSender`). A cell that arrives before its link is live, after it closes, or on a refused link is never charged or decrypted.
- `renewed(epoch, key) -> Self` (the epoch reset), per the follow-up rule:
  - every DID with a live link restarts with a zero-load ledger, so every live link still has a ledger;
  - the clock, the loads, the replay store and the ledgers of closed links are cleared;
  - `B` therefore holds within one epoch;
  - a token charged before the reset is rejected after it with `StaleEpoch`.
- This fixes the drop-leak (there is no handle to drop), cross-state closes (a close is keyed by the link, is idempotent and never goes negative), and charging on links that are not yet live or are closed.
- The doc overclaims are fixed:
  - the Charging law now states what the types guarantee (a charge requires a live link; admission requires a paid token; one token admits at most once) and what the 2a-4 shell owes (charge every received cell once, on its link and with its class's units, and peel only a charged cell);
  - the token is described as "evidence of one charge, not bound to a cell";
  - "never decrypted without a handle" becomes "only a live link is charged".
- Tests:
  - `test_ledger_counts_interleaved_link_generations` (renamed semantics): after `open(g₀) open(g₁) close(g₀)`, `g₀` is `LinkNotLive` and `g₁` is admitted;
  - `test_a_refused_generation_cannot_close_a_live_one`: a late close of a refused `g₀`, a repeated close, and a close of an unknown generation all leave `g₁` admitted across three windows;
  - `test_full_table_refuses_the_link_and_the_peer_redials`: the close of a refused link is a no-op on the table, and the redial comes as a new generation;
  - `test_renewal_keeps_live_links_and_clears_the_rest`;
  - the churn property now uses real generations, including closes of unknown generations.

### Low

1. `admissible_at` and `has_passed_at` move to `expiry.rs` as `pub(super)`, so every hop, including the exit in 2a-4, can call them. Each has a narrow `expect(dead_code)` naming #843.
2. Every `if let Ok(token)` / `if let Some(link)` in the tests is replaced with `expect`, or disappears with the handle.
3. New `test_late_admissions_never_exceed_the_global_bound_per_filter`:
   - 64 DIDs send 768-unit cells as fast as their budgets allow, for 20 quanta;
   - each token names a random expiry around its arrival window, some just outside it;
   - each token is held for a random delay of up to `V` before `admit`;
   - it asserts `max_tags(R_i[x]) · 768 ≤ G`, and that the fullest filter is within a factor of two of the bound, so the check is not vacuous.
4. Paper alignment: the token is `(arr, epoch)`, and `admit` checks both the layer epoch and the token epoch (see paper #851 at 58dd20da).
5. Noted (the #834 body wording).

### Gates
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, ffi.
- `cargo +nightly fmt --check`.
- rustdoc `-D warnings`.
- `taplo` and `typos`.
- core `--features dummy`.
- Tests: `onion`, with 24 admission tests, under `--profile ci`.

RyanKung:
Correction to Low 3 in the round-4 reply above: b516c066 replaces the test I described there.

The random schedule in 4c35aa6f never came close to G per filter: its fullest filter held 169 of 1365 cells. It also still passed with the old admission-time window check, so it did not guard H1. `test_late_admissions_never_exceed_the_global_bound_per_filter` is now a seeded adversarial schedule with a tight bound. Each of 4 rounds uses `u = 1024`, so 64 DIDs × `B/u` saturate `G`:
- **Phase A:** in quantum `k−6`, every DID spends its budget on cells naming `x = kQ`, which is outside the window at arrival. The tokens are held and admitted at random instants in `[(k−5)Q, kQ)`, when `x ≤ clock + V`.
- **Phase B:** in quantum `k−1`, after phase A has left the ledger window, every DID spends its budget again on `x`.

The test asserts that every phase-A token is `OutsideWindow` and that exactly `G/u` phase-B cells are admitted. I verified that it fails with the window judged at `self.clock_ms`: phase A is then admitted, which would give `2·G/u`.

RyanKung:
Round 5 is addressed in 04207087, implementing the two #834 L9 / #843 decisions.

### Medium

1. **A renewal requires a fresh epoch.**
   - `renew(&mut self, epoch, key, live) -> Result<Vec<OnionAdmissionLink>, OnionEpochNotFresh>` rejects `epoch == epoch_i` and then changes nothing.
   - It is `&mut self` rather than `self -> Result<Self, _>`, so a refused renewal cannot consume the state. The error path keeps the replay store, the clock and the ledgers intact.
   - `test_a_same_epoch_renewal_is_refused_and_cannot_readmit_a_replay` shows that after `(x, ν)` is admitted, a same-epoch renewal is refused and the replay is still `Replayed`. I mutation-checked it: removing the guard makes the test fail.
2. **`reconcile` makes the live set agree with core's snapshot.**
   - `reconcile(now, live: impl IntoIterator<Item = OnionAdmissionLink>) -> Vec<OnionAdmissionLink>` (`#[must_use]`) closes every live generation absent from the snapshot first, which frees slots. It then opens every snapshot generation that is not live, with the usual full-table refusal.
   - It returns the refused links, which the shell must close.
   - `renew` takes the same snapshot: it builds a fresh state and reconciles it.
   - The module doc drops "nothing can leak" and states the **event-pairing obligation**: a lost `Retired` pins a slot. Core splits pairs on callback replacement, and a bounded queue may drop events. So the shell calls `reconcile` on callback replacement and on a periodic tick of at most `V`, and a leak then lasts at most one tick.
   - `test_reconcile_recovers_lost_link_events`:
     - with every `Retired` lost, the table refuses a newcomer;
     - reconcile closes the absent generations, opens a never-reported snapshot generation (a lost `Admitted`), and the newcomer then fits;
     - a crowded snapshot shows the returned refusals, with a still-loaded closed ledger keeping its slot until it drains.

### Low

1. The pseudo-code now shows `token.epoch = e = epoch_i`.
2. It now says "within one epoch" for "never reset while it carries load", and for `(iii)` per DID.
3. `is_rolled_back_at(now) -> bool` (`now < clock − X₀`) is now in the pure core, and the Clock law points to it. Tested at the exact threshold.
4. The `live_links` cap is documented as a memory bound only: it is unreachable under core's laws and matters only if the event obligation is broken. `test_live_links_are_capped_as_a_memory_bound` covers it separately (`2·R` generations of one DID, then refused).
5. `live_links` is no longer stored. `live_link_count()` derives `Σ_d |live(d)|`, and the churn property asserts that it equals the model's live count after every step.
6. **API nits:**
   - `charge` now takes the `Copy` link by value;
   - `OnionAdmissionLink` derives no `Ord`/`PartialOrd`; `reconcile` groups the snapshot by DID instead;
   - `link_opened` does a single `entry` lookup;
   - `link_closed` now also uses `entry`;
   - `OnionBudgetRejection` is renamed `OnionChargeRejection`.
7. The late-admission test now admits each phase-A token at its own release instant in `[(k−5)Q, kQ)`, in time order, interleaved with phase B, so the doc is now accurate. It still fails under the admission-time window, which I re-verified by mutation.
8. The churn property now closes and charges a random live generation, or with probability ¼ a random generation that may be stale, refused or unknown. Old-generation closes, closes of refused links and stale charges therefore all occur in the property run.
9. Noted: the reducer switch is deferred to 2a-4 (#843).

### Gates
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, ffi.
- `cargo +nightly fmt --check`.
- rustdoc `-D warnings`.
- `taplo` and `typos`.
- core `--features dummy`.
- Tests: `onion`, with 28 admission tests, under `--profile ci`.

RyanKung:
Round 6 is addressed in b2bea934.

### Medium

1. **Linearisation obligation.**
   - The module laws now state it as a separate 2a-4 obligation, and the `reconcile` doc points to it: the snapshot must be linearised with the `Admitted`/`Retired` stream it repairs. That means delivering it through the same ordered channel, or reading and applying it atomically at the shell's queue-drain point.
   - The doc spells out both failure directions: a snapshot that predates a processed `Admitted(g)` closes the live `g`, whose cells become `LinkNotLive` for up to one tick, and the converse reopens a retired `g`.
   - "Agrees with core" is now stated only under this obligation. After a reconciliation the live set is `L \ refused`.
2. **The laws are now witnessed.**
   - **Renew with `L ≠ 𝓛ᵢ`.** In `test_renewal_keeps_live_links_and_clears_the_rest`, DIDs 1 and 2 are live and the snapshot is `{1, 3}`: it drops live DID 2 and adds non-live DID 3. After the renewal, the ledgers are exactly {1, 3} at zero load, 2 is `LinkNotLive`, and 3 is admitted in the new epoch.
   - **New `test_reconcile_restores_core_truth_under_lost_events`** (links/, seeded, 6000 steps):
     - Core's truth `L` gains a generation on each admit and loses one on each retire. About 1 in 6 events is lost before reaching admission.
     - Reconciliation runs about 1 step in 16.
     - Two states run in lockstep on identical inputs, one given each snapshot sorted and the other shuffled.
     - After every reconciliation it asserts:
       - `𝓛ᵢ = L \ refused`;
       - no link that was live before and is in `L` is refused;
       - a second reconciliation refuses the same links and leaves every ledger unchanged;
       - both states refuse the same links and hold identical ledgers.
     - After every step it asserts that `live(state)` equals the model's set, and that `LinkNotLive ⟺ link ∉ model`.
     - The shell's close of each refused link is modelled by removing it from `L`.
     - It also asserts more than 100 reconciliations and more than 50 lost events.
   - **Mutation checks:**
     - dropping `reconcile`'s close phase fails both reconcile tests;
     - a `renew` that ignores its snapshot fails the renewal test.

### Low

1. The fresh-epoch obligation is stated: the state checks only `e′ ≠ epoch_i`, and the shell draws `e′` independently and uniformly from `2¹²⁸`, so `A→B→A` has probability `≤ k·2⁻¹²⁸` over `k` resets.
2. `#[must_use] OnionRefusedLinks` is now returned by both `reconcile` and `renew` (`Result<OnionRefusedLinks, OnionEpochNotFresh>`), so `renew(..)?;` warns.
3. `reconcile` now applies opens in an explicit loop, with no mutation inside `filter`. It uses `&did` patterns and `.copied()` instead of `to_owned()` on `Copy` values.
4. The "no fairness" note now covers ordering: the event path is first come, first served, while `reconcile` opens in DID order, so small DIDs win a full table and DID prefixes can be ground cheaply.
5. Doc drift is fixed:
   - the liveness law says `renew` rebuilds from the snapshot;
   - the event-pairing law says the live set shrinks through `link_closed` and `reconcile` between resets;
   - the alphabet separates `δ` (timed, five summands) from the reset `ρ : S × Epoch × Key × 𝒫 Link → S × (Refused + NotFresh)`, which takes no time.
6. The churn property now asserts `LinkNotLive ⟺ ¬ modelled_live` for every charge, and so does the new reconcile property.
7. `admission/tests.rs` is split by proposition into `tests/{mod, window, budget, links, bloom}.rs`. `mod.rs` holds the shared pipeline and constructors, and every file uses explicit imports.

### Gates
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, ffi.
- `cargo +nightly fmt --check`.
- rustdoc `-D warnings`.
- `taplo` and `typos`.
- core `--features dummy`.
- Tests: `onion`, with 29 admission tests, under `--profile ci`.

RyanKung:
Round 7's Lows are addressed in b4d257b9, together with the epoch-reuse bound from paper #854 (5f4e266a).

1. **Refusals are justified.** `test_reconcile_restores_core_truth_under_lost_events` now asserts `is_full_for(state, link)` for every refused link on both the event path and the reconcile path. A refusal is justified only when the live-link set is full, or when the link's DID has no ledger and the ledger table is full. The test also counts delivered opens, delivered closes, event refusals and reconcile refusals, and asserts that each one occurs. Mutation-checked: a `reconcile` that opens nothing and reports every missing link as refused now fails the property.
2. **The linearisation obligation now names both snapshots and is stated positively.** `reconcile` and `renew` both treat their snapshot as authoritative. Under the obligation (the same ordered channel, or atomic application at the queue-drain point), after either call the live set is exactly core's registry minus the refused links. The two failure directions follow as the consequence of breaking it, and the quoted "agrees with core" is gone.
3. `reconcile` now starts with `self.advance(now_ms)`, so the clock law holds even for an empty snapshot with no live links, and an idle node's periodic reconciliation drops expired filters.
4. The `renew` doc now reads "the reset `ρ(epoch, key, live)`", and the test is renamed `test_renewal_rebuilds_live_links_from_the_snapshot`.
5. Every doc paragraph with a line longer than 100 columns is rewrapped (`mod.rs:19`, `tests/links.rs:68` and others in `admission/` and `expiry.rs`), so all lines are now at most 100 characters. `tests/mod.rs` `live()` is renamed `live_filters()`.

**Epoch reuse, per #854.** The fresh-epoch paragraph now cites both bounds. The state rejects `e′ = epoch_i`, and the shell redraws `e′` uniformly from `2¹²⁸` on that rejection. Over `k` resets, reusing one *fixed* earlier epoch has probability `≤ k·2⁻¹²⁸`. Reusing *any* earlier epoch (`A → B → A`) is at most the birthday bound `k²·2⁻¹²⁹`.

Gates:
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, ffi.
- `cargo +nightly fmt --check`, rustdoc `-D warnings`, `taplo`, `typos`.
- core `--features dummy`.
- Tests: `onion` (29 admission tests) under `--profile ci`.

RyanKung:
All four round-8 items are addressed in c0acf35f.

1. **Line width.** `admission/mod.rs:153` and `circuit/tests/test_circuit_protocol.rs:942` are rewrapped. Every line under `circuit/admission/`, `circuit/expiry.rs` and `test_circuit_protocol.rs` is now at most 100 characters, checked by character count because `wrap_comments = false`.
2. **Function length.**
   - `test_reconcile_restores_core_truth_under_lost_events` is now a 25-line driver over a `Lockstep` model struct (rng, the two states, `L`, the modelled set and the counters).
   - It has one step method per branch: `open`, `retire`, `reconcile` and `charge`. `reconcile` carries the four laws and the refusal justification.
   - No function in the admission module or its tests exceeds 100 lines.
   - `is_full_for` now takes the live-link count from the model (`modelled.len()` on the event path, `|L \ refused|` on the reconcile path) rather than restating the guard.
3. **Refusal order is pinned.** `test_reconcile_recovers_lost_link_events` asserts `refused.links() == link(13..20)`: `reconcile` opens in DID order, so DIDs 10–12 take the three free slots.
4. **Wording.**
   - The Clock law now reads `ρ = reconcile ∘ clear`: clear the ledgers, the clock and the replay store, then reconcile with the snapshot, so the live links are the snapshot up to refusals.
   - `renew`'s doc notes that the cleared table is empty and a snapshot has no more DIDs than links, so nothing is refused when `|L| ≤ 2·R`.

Gates:
- clippy `-D warnings`: native `--all --tests`, node lib, browser `--tests`, browser lib, ffi.
- `cargo +nightly fmt --check`, rustdoc `-D warnings`, `taplo`, `typos`.
- core `--features dummy`.
- Tests: `onion` (232 tests, 29 of them admission) under `--profile ci`.

Issue #840, opened 2026-09-23
Onion Kleisli Phase 2a-1: Sphinx header, carry seeds and AEZ segments as pure primitives (#834)


RyanKung:
Part of #834, Phase 2a, sub-PR 1 of 4. The spec is the #834 body and `papers/onion-kleisli-pipeline.tex` on `develop`.

## Scope

This PR adds pure primitives only. Nothing touches the live data plane or any wire message.

1. **Fixed-length header (Sphinx).**
   - Build and peel the header of `MAX_ONION_LOOP_HOPS` layers with per-hop PRG filler and a MAC `γ`, following the paper's header definition and its correctness proposition (#834 L5′, D6″).
   - Key transport is a **single blinded secp256k1 group element**: ECDH against the hop's delegatee key, and the blinding factor is derived from the shared secret.
   - DIDs are raw 20 bytes.
2. **Carry seeds (D7).**
   - Segment seed `σ`; `k = HKDF-SHA256(σ, domain tag, index) → [u8; 48]`.
   - A weak derived key (zero subkey `I`, `J`, `L`) is detected so that the client re-draws `σ`.
3. **Carry segments (D7).**
   - The producer wraps `s + 1` AEZ layers with `rings-aez` (#838): `τ = 16` innermost, then `τ = 0` per relay.
   - Relays remove one layer. The consumer checks the authenticator.
   - `()` is encoded as `ε`.
4. **Layer encoding (D6, D6″).**
   - The uniform layer struct and its fixed-width encoding, with `ℓ` computed as a `const` from its fields (≈ 221 B).
   - `C_b` is computed for every class `b`.

## Laws to discharge (tests)

- Header correctness for every `H ≤ MAX_ONION_LOOP_HOPS` and every position, `i = H` included (client position, D6′).
- L5′ lengths: the header and carry lengths are equal at every position and for every `H`.
- L8: bytes differ on every edge. A one-bit flip anywhere in a carry segment makes the consumer reject the whole value.
- Seeds: the derivation matches its definition, and a weak key is detected.

## Benchmark gate

Measure per-cell peel cost (ECDH, blinding, PRG, MAC, one AEZ layer) natively and on `wasm32` (headless Chrome). The target is at least 109 cells/s per neighbour (#834 L9) on wasm. Report the numbers in the PR; if the target is missed, stop and report.

## Out of scope

Admission and replay (2a-2), registration and route (2a-3), and data-plane integration (2a-4).

PR #845, opened 2026-09-23
Onion Kleisli Phase 2a-1: Sphinx header, carry seeds and AEZ segments as pure primitives (#840)

RyanKung:
Closes #840
Part of #834

Phase 2a, sub-PR 1 of 4. This PR adds the pure primitives of the onion Kleisli loop as a new module, `crates/node/src/onion/sphinx/`. Nothing is wired to the data plane, and no live wire message changes.

## What is added

| module | spec | content |
|---|---|---|
| `sphinx::layer` | D6, D6″ | `OnionLayer`, the uniform layer `λ = (f, ā, next, e, x, ν, σ_in, σ_out, γ_{i+1}, t)`, with its fixed-width encoding. `ONION_LAYER_BYTES = ℓ` is a `const` sum of the field widths (= 221). `relay` has no argument field, so a relay layer with arguments cannot be represented, and decoding rejects any non-canonical string. |
| `sphinx::header` | Def. fixed-length header, L5′, D6″ | The Sphinx header `χ = (α, β, γ)` of `MAX_ONION_LOOP_HOPS = 14` layers (`|χ| = 3143`), `OnionHeader::build` (client) and `OnionHeader::peel` (hop). |
| `sphinx::seed` | D7 | Segment seed `σ_k` → `σ_{k,j} = KDF₃₂(σ_k, relay j)` and `σ_{k,c} = KDF₃₂(σ_k, consumer)`; each seed → `k = KDF₄₈(σ)` → AEZ. `OnionSegmentSeed::draw` re-draws `σ_k` until every derived key is strong. |
| `sphinx::carry` | D7, L3, L8 | `OnionCarry`: the producer seals `Enc⁰_{r_1} ∘ … ∘ Enc⁰_{r_s} ∘ Enc¹⁶_c (pad v)` on `rings-aez`; a relay peels one `τ = 0` layer; the consumer opens and checks the authenticator. `()` is encoded as `ε`. |
| `sphinx::class` | D6, L3 | `OnionLoopClass` covers the cell buckets from 16 KiB to 12 MiB. It defines `C_b = b − |χ| − F`, `C₀ = C_b − 16` and `from_cell_bytes`. |

Header key schedule, per position:
- **ECDH:** `K_i = d_i·α_i`, computed with the new `DelegateeKey::diffie_hellman`, a constant-time k256 scalar action.
- **Key derivation:** `HKDF-SHA256(salt, SEC1(α_i) ‖ x(K_i))` expands to three keys:
  - `prg` feeds `ρ_i`, a ChaCha20 keystream of `(Ĥ+1)ℓ` bytes;
  - `mac` feeds `γ`, HMAC-SHA256 truncated to 16 bytes;
  - `blind` is 64 bytes, wide-reduced mod `n`, and zero is rejected.
- **Blinding:** `α_{i+1} = b_i·α_i`.

`peel` checks in this order:
1. `α` must be a non-identity curve point, checked before any ECDH.
2. `γ` must verify, checked before any decoding.
3. The peeled bytes must decode as a layer.
4. `α_{i+1} ≠ O`.

Changes outside the new module, each needed by it:
- `rings-core`: `DelegateeKey::diffie_hellman`. The delegatee secret stays private.
- `onion::signature`: each symbol gets a one-byte layer code (`relay = 0`, `tcp = 1`, `https = 2`), with an injectivity test.
- `OnionExitEpoch::to_bytes` and `OnionForwardNonce::to_bytes`, for the layer encoding.
- `rings-node` dependencies: `rings-aez`, `k256`, `hkdf`, `hmac`, `sha2`, `chacha20`, `zeroize`. All were already in the lockfile.

## Assumptions for 2a-4, agreed with the #834 dispatcher

- **No discriminant (`F = 0`).** A cell is the fixed-width string `α ‖ β ‖ γ ‖ y` with `|cell| = b`, and its length is its class. So `C_16KiB = 13241` and `k = 4` (D8).
  - Consequence: link cover traffic becomes a uniformly random `b`-byte cell, which fails the `γ` check at the receiving hop. The `γ` check must therefore come before any replay-store insertion or admission charge, and the ECDH of a cell that fails `γ` is charged to the sender's crypto budget.
- **Integrity and neighbour.** With the Sphinx header, the per-edge `seal_message` cell AEAD goes away:
  - header integrity is `γ`;
  - carry integrity is the consumer's AEZ authenticator;
  - the neighbour `from` is the authenticated transport link.

Both assumptions are also stated in the `sphinx` module docs.

## Laws discharged (`sphinx::tests`)

- **Header correctness:** `test_every_position_of_every_loop_length_peels_its_own_layer`. For every `H ∈ 1..=14` and every position `i ≤ H`, `i = H` included, the peel under hop `i`'s key yields exactly `λ_i`, so every `γ_i` verifies.
- **L5′ lengths:**
  - `|χ|` is 3143 at every position and for every `H`; `test_widths_are_the_specified_constants` pins `Ĥ = 14`, `ℓ = 221`, `|β| = 3094`, `|χ| = 3143`.
  - `C_b = b − 3143` for every class and `C_16KiB = 13241` (`test_carry_width_per_class`).
  - The carry is `C_b` wide on every edge.
  - No forwarded header shows an all-zero layer block, because the filler hides where the loop ends.
- **L8:**
  - `χ_i ≠ χ_{i+1}` (both `α` and `β`) and `y_{j−1} ≠ y_j` on every edge.
  - A one-bit flip at 70 sampled offsets on each of the `s + 1` edges makes the consumer reject the whole value with `Inauthentic` (`test_one_bit_flip_on_any_edge_rejects_the_whole_value`).
- **Seeds:**
  - The derivation matches HKDF written out independently (`test_derivation_matches_its_definition`).
  - The seeds are pairwise distinct.
  - A zero `I`/`J`/`L` subkey is detected by `OnionCarryKey::new`, which every derivation goes through (`test_weak_key_is_detected`). A weak HKDF output itself (≈ 2⁻¹²⁸) cannot be exhibited, so re-drawing is covered by `draw` returning only strong segments.
- **Rejection:**
  - a wrong key or a flipped `β`/`γ` bit gives `Mac`;
  - an invalid `α` gives `GroupElement`;
  - random headers (the F = 0 cover) are rejected at `α` or `γ`;
  - `H = 0` and `H = 15` give `HopCount`.

All tests are deterministic: seeded `StdRng`, fixed hop scalars.

## Benchmark gate (#834 L9: ≥ 109 cells/s per neighbour on wasm32)

One cell is the header peel (α check, ECDH, key schedule with blinding, ChaCha20 `ρ`, HMAC `γ`, layer decode, `α′ = b·α`) plus the relay carry step (`KDF₄₈(σ_in)`, AEZ key setup, one AEZ decipherment of 13241 B). The measurement uses 1000 cells, a release build, and an Apple M1 Max.

| target | header only | carry only | **whole cell** |
|---|---|---|---|
| native (aarch64) | 9615 cells/s | 1727 cells/s | **1460 cells/s (0.69 ms)** |
| wasm32, headless Chrome 153 | 4202 cells/s | 2597 cells/s | **1626 cells/s (0.61 ms)** |

wasm32 clears the 109 cells/s target by about 14.9×. On aarch64 the AES rounds run as fixsliced software, because the build has no `--cfg aes_armv8`. That is why the native carry is slower than wasm here. On x86 the rounds use AES-NI.

The benchmarks are `#[ignore]` measurements, not assertions:

```
cargo test --release -p rings-node --lib sphinx::tests::test_peel_cost -- --ignored --nocapture
CHROMEDRIVER=/usr/local/bin/chromedriver cargo test --release -p rings-node --lib --target wasm32-unknown-unknown \
  --features browser_default --no-default-features -- --include-ignored --nocapture bench_peel_cost
```

## Gates run locally

- `cargo +nightly fmt --all -- --check`, `taplo format --check`, `typos`
- `cargo clippy -p rings-node -p rings-core --all-targets -- -D warnings` (native)
- `cargo clippy -p rings-node --features browser_default --no-default-features --target wasm32-unknown-unknown --tests -- -D warnings`
- `cargo clippy -p rings-node --features ffi -- -D warnings`
- `cargo clippy -p rings-core --features dummy --no-default-features --tests -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p rings-node -p rings-core --no-deps`
- `cargo test -p rings-node --lib onion::` (212 passed, Phase 1 golden bytes unchanged), `cargo test -p rings-core --lib delegation`


RyanKung:
7d5622fa re-checks the benchmark, because native (1460 cells/s) came out slower than wasm32 (1626 cells/s).

Measurement checks, all confirmed:
- The native run was `--release` with `--test-threads=1`.
- Both targets run the same `measure()`: same `H`, class and number of passes, one warm-up each, and every operand built outside the timed regions.
- Both clocks are the millisecond wall clock, and every row spans at least 15 ms.
- k256 has its default features, including precomputed tables.

The new per-primitive breakdown (see the updated **Benchmark gate** section) puts the whole gap in AES:
- **AEZ decipher of 13241 B:** 593 µs native vs 383 µs on wasm32. On aarch64, `aes 0.8.4` uses the ARMv8 AES instructions only under `--cfg aes_armv8`. Otherwise its soft single-block `hazmat::cipher_round` bitslices four copies of the block on every round; wasm32 bitslices two.
- **With `--cfg aes_armv8`:** AEZ takes 7.5 µs and the native cell is 129.5 µs (7722 cells/s).
- **All other primitives are faster natively:** ECDH 34 vs 94.5 µs, blind 33 vs 93.5 µs, PRG 5.5 vs 7 µs, MAC 10 vs 12 µs.

Corrected whole-cell numbers:

| target | whole cell |
|---|---|
| native, default build | 689.5 µs (1450 cells/s) |
| native, `--cfg aes_armv8` | 129.5 µs (7722 cells/s) |
| wasm32 | 618 µs (1618 cells/s) |

The gate still passes at about 14.8× the 109 cells/s target. Enabling `aes_armv8` for aarch64 builds is a separate build-configuration decision.

RyanKung:
f67ca901 addresses the review (issuecomment-5809041877). Every finding and what it became:

**High (spec changes, paper 800ea69f)**
- **H1.** The MAC is now `γ_i = HMAC-SHA256(b ‖ β_i)[0, 16)`, where `b` is the class's cell length as 8 bytes big-endian (`OnionLoopClass::mac_label`). `build` and `peel` take the class. `test_relabelled_cell_dies_at_the_next_honest_hop` checks two cases: peeling under the wrong class fails `γ`, and a cell relabelled to 12 MiB after an honest peel dies at the next hop.
- **H2.** The tag is now `t_⋄ := γ_{H+1}`. `build` takes the `OnionLoopTag`, writes it as `λ_H`'s next-MAC field, and the guard forwards it; `OnionHeader::loop_tag` reads it at the client. The per-layer tag field is removed, so **ℓ = 205, |β| = 2870, |χ| = 2919, C_16KiB = 13465**, all pinned in `test_widths_are_the_specified_constants` and `test_carry_width_per_class`. The header-correctness test now also checks, for every `H ∈ 1..=14`, that the tag reaches the client.

**Medium**
- **M1.** `onion::sphinx` and everything in it is `pub(crate)`, with one scoped attribute on the module declaration: `#[cfg_attr(not(test), expect(dead_code, reason = "… #843 …"))]`. The raw-bytes `OnionCarryKey::new` is gone; `OnionCarrySeed::key` (`KDF₄₈` of a seed) is the only constructor.
- **M2.** New `rings_core::ecc::prime_order`:
  - `NonIdentityPoint<Secp256k1>` and `NonZeroScalar<Secp256k1>`, for `G ∖ {O}` and `Z_n^*`, which are closed under the module action;
  - generator multiplication, `P·k`, `Z_n^*` products, SEC1 in both directions, wide reduction, and the DH x-coordinate;
  - law tests.

  The node crate goes through this module and no longer depends on `k256` directly.
- **M3.** `DelegateeKey::diffie_hellman(&NonIdentityPoint<Secp256k1>) -> Zeroizing<[u8; 32]>`. The secret scalar only ever exists inside a zeroizing `NonZeroScalar`.
- **M4.** `OnionSegmentKeys.relays` is now `[OnionCarryKey; ONION_SEGMENT_RELAYS]`. The `[first, second]` pattern pins `s = 2` at compile time.
- **M5.** `OnionLoopClass(OnionCellBucket)`, with `TryFrom` rejecting `KiB4`. `OnionCarry` stores only its bytes.
- **M6.** Removed:
  - `GroupElement` for x·G, the identity filter, and the b·α check (all excluded by type now);
  - `OnionCarryError::{Expansion, Decrypt}`: AEZ decryption failures map to `Inauthentic`, and seal's impossible short-slot case maps to `Width`.

  The `peel` doc now lists the checks in code order.

**Low**
- **L1.** `OnionSignature::code` is the table position. `by_code` returns the coproduct `OnionSymbol::{Relay, WorldFacing}`, which the layer decoder matches on directly, and its inverse law is tested.
- **L2.** `layer_record!` declares the fields once and derives ℓ, encode and decode. `take_array` is removed. `β`/`ρ` are `[[u8; ℓ]; Ĥ]` / `[[u8; ℓ]; Ĥ+1]`, split by irrefutable array patterns.
- **L3.** Zeroized: the U512 blind, the HMAC key, `K_i` and the blinding. Seeds compare in constant time and are no longer `Clone`, and neither are layers. `OnionHeaderMac` implements only `ConstantTimeEq`.
- **L4.**
  - `ρ_i` is computed once per hop.
  - Relay indices are a const `[u8; s]`, so there is no `as u64`.
  - `draw` fails closed after four weak draws.
  - The hop count is checked once, in `OnionHeaderRoute::new`.
- **L5.** New tests for `−α`, the guard key at positions 1 and H, H1, and H2.

**Benchmark**
- Now uses `web_time::Instant` (`performance.now()` in the browser), 200 warm-up runs, and `black_box` on every result.
- The module doc and the PR body use the same `browser_default` target.
- Whole cell: **wasm32 617 µs (1621/s)**, native 683 µs (1464/s). The gate passes at about 14.9×. The per-primitive table is in the PR body.
- `aes_armv8` is not enabled here.

All local gates pass: fmt, taplo, typos, clippy (native, node lib, wasm tests and lib, ffi, core dummy, core wasm), `check --all-features`, rustdoc, and the node `onion::` tests (214). In `rings-core`, the full lib run had one failure: `test_handle_connect_node`, a networking test, hit its deadline under full-suite load. It passes 3/3 when run alone, and this PR doesn't touch it.

The branch was not rebased. `develop` only moved `papers/` (800ea69f), which does not conflict.

RyanKung:
fbe61157 addresses the round-2 review (issuecomment-5809717330). What each finding became:

**Medium**
- **M1.** New `sphinx::cell::OnionCell`, the one parser of a received cell: `parse(w) = (b, χ, y)` with `b = |w|`, and `to_bytes` as its inverse. `OnionHeader::peel` is now `pub(super)`, so a hop only peels through `OnionCell::peel`, which uses the parsed class. The class flows along `parse → peel → relay`, or `parse → peel → consume → produce`, and is never a parameter at a hop.
  - The relabel test is now cell-level: an honest peel at hop 1, then the forwarded 16 KiB cell re-padded to 12 MiB, parses as a 12 MiB cell and is `Invalid` at hop 2, while the honest bytes pass.
  - New `test_cell.rs` runs one whole loop through parsed cells: client → r₀₁ → r₀₂ → h₁ (consume v₀, produce v₁) → r₁₁ → g → client.
- **M2.** `ecc::prime_order` is rebuilt as subtypes of `ecc::group`:
  - `NonIdentityPoint<C>(Point<C>)` and `NonZeroScalar<C>(Scalar<C>)`, with inclusions `From` into the carriers and `TryFrom` partial inverses rejecting `O` and `0`, generic over a `PrimeOrder` marker (secp256k1 and secp256r1, both cofactor 1);
  - sampling is the carrier's `random_scalar_with_rng`;
  - SEC1 decoding is `Point`'s codec followed by the subtype check;
  - `TryFrom<Point<Secp256k1>> for PublicKey<33>` now factors through the total `From<&NonIdentityPoint>`, so the codec is shared rather than duplicated;
  - the secret scalar comes from the existing `SecretKey::secp256k1_scalar`, and `Scalar<C>` gains `Zeroize`;
  - tests cover the inclusions, the action commuting with `ι`, SEC1, wide reduction and DH.
- **M3.** `code`/`from_code`, `OnionSymbol`, and `OnionExitEpoch::to_bytes`/`OnionForwardNonce::to_bytes` are `pub(crate)`, each under `#[cfg_attr(not(test), expect(dead_code, reason = "… #843"))]`. No new `pub` item remains outside core and aez.

**Low**
1. The bench doc now says `C_16KiB = 13465`.
2. The probability is `≈ 2^−499`, with the per-draw weakness `≈ 9·2^−128` stated.
3. The docs say a random `α` is rejected at decoding (≈ 99.6 %) before any ECDH. `GroupElement` and `Mac` are now the single outcome `OnionHeaderError::Invalid`, charged alike per #834.
4. Zeroization:
   - the wide bytes are built in a `Zeroizing<WideBytes>` and reduced by reference;
   - ChaCha20 builds with its `zeroize` feature;
   - **not fixed: the HKDF PRK and HMAC ipad/opad states.** hkdf 0.12 and hmac 0.12 expose neither `Zeroize` nor a wipe, and I don't want `unsafe` or a pre-release crate bump in this PR. The state is one per hop and per cell, and dropped at once. If you want this fixed, I'd open a separate issue for moving to zeroizing versions.
5. `build` draws `t_⋄` itself and returns `(OnionHeader, OnionLoopTag)`. `OnionLoopTag` and `OnionHeaderMac` have only `ConstantTimeEq`.
6. `OnionHeaderRoute::new` validates each key into `NonIdentityPoint` once; `build` has no key error arm left.
7. rings-aez gains `Expanded<τ>` (constructors `new`, which rejects `|w| < τ`, and the total `with_authenticator_slot`), plus `encrypt_expanded`/`decrypt_expanded`. `OnionCarry` holds an `Expanded<16>`, so `seal` has no error mapping and `open` has one real outcome, `Inauthentic`. Law tests show the typed path agrees with `encrypt`/`decrypt` and zeroizes on rejection.
8. The const assertion now states and checks what it claims (`DEFAULT` = `KiB16`, and `C₀ ≥ 32`). `mac_label` is the bucket's `repr(u8)` discriminant, so there is no widening cast. The MAC input is fixed-width `1 + 2870` bytes.
9. `Σ_W` is a private `repr(u8)` enum inside `OnionServiceName`, and the layer code is the coproduct `OnionSymbol = 1 + Σ_W`:
   - `code(inl) = 0` and `code(inr w) = 1 + w`;
   - `from_code` is its left inverse, returning `None` past `Σ`;
   - there is no table search and no silent `3`.

   `OnionLayerApplication = 1 + Σ_W × A` is the application type, i.e. the same coproduct with the arguments added. `OnionSymbolPosition` stays the D1 attribute of a spec. Service names still order by canonical string, so every existing ordering is unchanged.
10. Tests and bench:
    - `draw`'s retry is `first_strong(draw, derive)`, a pure combinator. It is tested with an injected derivation that fails k times, for k = 0…5, including the fail-closed path after four failures.
    - The `prime_order` tests use a fixed key.
    - "Whole cell" is now `parse → peel → relay → encode` on `black_box` input, with no clone in the timed region.
    - The 139-character line is wrapped.
    - `diffie_hellman`'s doc no longer names Sphinx.

**Benchmark:** wasm32 runs a whole cell in 633.8 µs (1578 cells/s, about 14.5× the target). Native is 758.3 µs; its AEZ row uses soft AES without `aes_armv8`. The table is in the PR body.

**Gates:**
- fmt, taplo and typos pass;
- clippy passes for native (node, core, aez), node lib, wasm (node tests and lib), ffi, core dummy and core wasm;
- `check --all-features` and rustdoc pass;
- node `onion::` passes 216 tests, and `rings-aez` passes.
- In the full `rings-core` lib run, `test_handle_connect_node` again hit its deadline under load. It passes alone (2/2 this round) and is untouched by this PR.

**#846:** noted in the PR body. `onion::loop_shape` is canonical, and whichever PR merges second deletes the sphinx constants and adapts to `OnionProcessEpoch`.

RyanKung:
6e7ee8f6 addresses the round-3 review. What each finding became:

**Medium**
- **M1.** `onion_signature!` declares `Σ` once:
  ```
  relay: "relay",
  world_facing: [Tcp tcp "tcp", Https https "https"]
  ```
  From these rows it generates:
  - `OnionSignature`'s fields, `ONION_SIGNATURE`, and `symbols()` (table order = layer code);
  - the `Σ_W` enum, `ALL` and `spec()`;
  - `OnionServiceName::{tcp, https}`.

  `OnionSymbol::{code, from_code}` read only the enum and `ALL`, so a new row cannot desynchronise them. `test_from_code_inverts_code` enumerates all of `u8` and checks contiguity, `code ∘ from_code`, the spec of each position, and `|Σ_W| + 1 = |Σ|`, with no hand-written list. `test_layer`'s code range comes from `symbols().len()`.
- **M2.** `prime_order` is now the one secp256k1 SEC1 encoder:
  - `From<&NonIdentityPoint>` is the only encoder; `From<K256PublicKey>`, `From<VerifyingKey>`, `TryFrom<AffinePoint>` and `TryFrom<Point>` all factor through it, and the zero-fallback copies are deleted from `ecc/mod.rs`;
  - `TryFrom<PublicKey<33>>` decodes `k256::PublicKey` straight into `G ∖ {O}` via `From<K256PublicKey>`, with no affine round trip or re-check;
  - `group.rs` no longer imports `prime_order`, since the `Point → PublicKey` codec moved there;
  - `PrimeOrder: … + sealed::Sealed`, implemented for secp256k1 only, and the `Secp256r1` impl is dropped;
  - nits: `Mul` goes through `C::mul`/`C::scalar_mul` on references with no clones; `From<&NonZeroScalar> for Scalar` is removed; errors are now `IdentityElement`/`ZeroScalar`; `random_with_rng` requires `CryptoRng + RngCore`; the `from_secret_key` comment is corrected.
- **M3.** rings-aez:
  - one private `encrypt_in_place`/`decrypt_in_place` core under both APIs; the slice API only adds its `|buffer| ≥ τ` check, via `checked_sub`, so nothing is masked;
  - the typestate is `Plaintext<τ>` (no `Clone`/`Debug`/`Eq`, `Zeroizing`, capacity reserved for the slot so `seal` never reallocates) → `Aez::seal` → `Ciphertext<τ>` (`|C| ≥ τ`) → `Aez::open` → `Plaintext<τ>`. Only `open` yields a message;
  - `DecryptError::Inauthentic(Inauthentic)`;
  - `Expanded` is removed; the README and laws are updated (typed = checked, bit flip ⇒ `Inauthentic`, `|C| < τ` rejected).
- **M4.** Per-stage errors: `OnionCellWidth` (parse), `OnionHeaderRouteError { HopCount, PublicKey }`, `OnionBlindingError` (build), `OnionPeelError { Invalid, Blinding, Layer }`, `OnionStepError { Key, Open }`, `OnionClientError { Blinding, ValueTooWide }`, `OnionValueTooWide` (produce), and `OnionOpenError { Inauthentic, Padding }` (client open). Every variant is constructible at its stage.
  - `OnionPeeledCell::step()` dispatches on `λ_i.application` to `Relayed { layer, cell }` or `Consumed { layer, value, producer }`.
  - `OnionProducer` holds the `OnionSegmentKeys` of `σ_out` from `λ_i`, and `produce(v′)` takes only the value.

**Low**
1. The PRK is wiped: `hkdf_extract` zeroizes `finalize()`'s PRK and is shared by the header and seed schedules. The remaining HMAC/HKDF state is left to your issue.
2. `OnionCell::client(route, class, keys, v, rng) → (cell, t_⋄)` is the only client constructor. `OnionHeader::build` and `carry::seal` are sphinx-internal, so hop code cannot seal.
3. `parse(Vec<u8>)` splits the slot off with `split_off`, keeping the header allocation (capacity `b`), and `into_bytes` re-encodes into that allocation. A relay now does one allocation and one slot copy per cell, instead of three allocations. Fully zero-copy would need the slot before the header in the cell (`y ‖ χ`), which is a layout decision I can raise for 2a-4 if you want.
4. The three `b`s: `b` is only the class now; the blinding factor is `z_i` and the padded message `m_k`.
5. The `prime_order` nits are covered under M2.
6. `OnionLoopTag` derives `Eq`/`Hash` again, and its doc gives the rationale (the client maps `t_⋄`; only the guard also sees it).
7. Tests:
   - `test_cell` now puts the guard key at positions 1 and H;
   - the label injectivity check covers all pairs;
   - `test_seed` uses `SEGMENT_SEED_DRAWS`;
   - `test_cell` also checks `layer()` before `step` and the consumed layer's application.
8. `signature.rs` now notes that 2b's intermediate `invoke` shapes make the codec `1 + Σ_I + Σ_W`, a change to this table at the cutover.

**Bench:** a whole cell (parse → peel → step → encode) takes 631.7 µs on wasm32 (1583 cells/s, about 14.5× the target) and 709.9 µs natively. Each run's buffer is prepared before timing starts, since `parse` consumes it.

**Gates:** all green. That covers fmt, taplo, typos; clippy for native (node/core/aez), node lib, wasm (node tests and lib, aez), ffi, core dummy and core wasm; `check --all-features`; rustdoc; `rings-aez` tests; and node `onion::` (216). In the full `rings-core` lib run, `test_handle_connect_node` flaked again under load. CI's wasm failure is #853, unrelated, and I have not rerun CI.

**#846 checklist:** unchanged, and in the PR body.

RyanKung:
ec1d9a31 addresses the round-4 review:

**Medium (SURB size).** A symbol hop's `step()` now yields `Consumed { layer, value, surb }` with `OnionSurb = (class, χ_{i+1}, σ_out)`: one byte of class, a 2919 B header and a 32 B seed, whatever the class. With the next DID and `x` that the D8 pool keeps beside it, that is your ≈ 2979 B. The SURB holds neither the received buffer nor derived keys; `OnionSurb::produce(v′)` derives the keys of `σ_out` when spent and seals into a fresh buffer of the class, failing with `OnionProduceError { Key, ValueTooWide }`. `OnionProducer` is gone. `test_cell` spends the SURB.

**Lows**
1. The `cell.rs` docs now say what happens: a relay allocates once and copies the slot twice. Zero-copy over a slice view is #843, as you recorded, and the layout stays `α‖β‖γ‖y`.
2. `Plaintext::new` keeps a buffer that already has spare capacity for τ, and otherwise copies once and wipes the original. `carry::seal` builds `pad(v)` with capacity `C_b`, so producing allocates once. A law test checks that the buffer's address survives `seal`.
3. The client seed supplied twice is left to 2a-4 (#843), as you suggested; the loop builder that owns σ₀ belongs with route construction.
4. The single encoder is now `encode_sec1(&AffinePoint)` on the affine carrier. `From<K256PublicKey>` and `From<VerifyingKey>` pass `as_affine()` with no inversion, so `recover_hash` pays nothing extra; `NonIdentityPoint` normalises once. The SEC1 test also checks the `k256::PublicKey` path.
5. Doc drift fixed: `m_k` in `sphinx/mod.rs`, `z·α`/"blinding factor z" in the bench, and `hkdf_extract` cites #852. The PR body no longer mentions `Expanded`.
6. `world_facing` is now `OnionServiceName::world_facing()`, generated with the table. Its two call sites are updated.
7. `Ciphertext` records `|M|` from the one `checked_sub` in `new` (or from `seal`), so `open` truncates to it with no subtraction. `into_slotted` extends with `repeat_n(0, τ)`, and `Plaintext::new` uses `spare_capacity_mut().len()`, so no `+`/`−` remains in `typed.rs`. The slice `decrypt` returns its prefix through the same checked `split_authenticator` (`checked_sub` + `split_at_mut_checked`) as its precondition.
8. `parse`: every failure point is a width condition (`|w|` is not a class length, `|w| < |χ|`, or `|w| − |χ| < τ`), and the last two cannot hold once the first passes. So `OnionCellWidth` is the exact reason, and adding variants for unreachable cases would bring back the round-1 M6 problem. The doc says this now. If you prefer a total header decode from a fixed array, stable Rust has no checked array split to express it without the same `Option`.

**Bench:** wasm32 whole cell 700.8 µs (1427 cells/s), native 841.6 µs. Every row is about 15–20 % slower than in round 3 because the machine was under load average ≈ 50 during this run; the gate still passes at about 13×.

**Gates:** fmt, taplo, typos, all clippy targets (native, node lib, wasm, aez wasm, ffi, core dummy, core wasm), `check --all-features` and rustdoc pass. `rings-aez` passes 15 tests, node `onion::` 216 and core `ecc` 59. In the full `rings-core` lib run under that load, a storage-repair networking wait flaked; it passes 2/2 alone.

**#846 checklist:** unchanged; the line numbers are in your review.

RyanKung:
971172e5 addresses the round-5 review:

**Medium (υ).**
- `OnionSurb` is now D8's `υ = (next, χ_υ, σ_υ, x_υ)` with the class of its cell: `{ class, next: Did, header, outbound: OnionSegmentSeed, expires_at_ms }`. `produce(v′)` returns `(next, OnionCell)`, and `expires_at_ms()` exposes `x` for the pool's drop and least-`x` order.
- `const _: () = assert!(size_of::<OnionSurb>() <= 3 * 1024)` pins its class independence, and the doc says 2980 B (1 + 2919 + 32 + 20 + 8). No constructor beyond `step` yet; the batched-credit codec lands in 2a-4 and decodes into this type.
- `OnionLayer` is now `{ head: OnionLayerHead, inbound, outbound }`, where the head is key-free (`f, ā, next, e, x, ν`). The layer no longer derives `ZeroizeOnDrop`, since each seed wipes itself, so `step` can destructure it.
  - `step` turns σ_in into the key and moves σ_out into the SURB, then yields only the head: `Relayed { head, cell }` and `Consumed { head, value, surb }`. Each seed lives in exactly one place.
  - `OnionPeeledCell::head()` exposes the head for admission before the step.
- `test_cell` checks that the SURB's `next` and `x` are the symbol layer's.

**Lows**
1. The lines over 100 characters are wrapped: `header.rs`, the `vectors.rs` assertion, the `onion/mod.rs` expect reason, and the `signature.rs` matcher. No line of this PR's diff exceeds 100 characters now; the longer lines that remain in touched files are pre-existing.
2. `measure()` is split into `component_rows(key, σ_in, rng)` and `real_path_rows(key, header, cells)`.
3. `TryFrom<K256AffinePoint> for PublicKey<33>` checks `point == AffinePoint::IDENTITY` and calls `encode_sec1` directly, with no projective round trip. The SEC1 test covers the identity rejection and the affine path.
4. The SURB doc now says "a class byte" and 2980 B.

**Bench** (load average ≈ 12): whole cell, parse → peel → step → encode, takes 628.0 µs on wasm32 (1592 cells/s, about 14.6× the target) and 714.0 µs natively.

**Gates:** all green. That covers fmt, taplo, typos; all clippy targets; `check --all-features`; rustdoc; `rings-aez` (15 tests); node `onion::` (216); and the `rings-core` lib suite, 749/749 with the two filed load-sensitive networking tests skipped.

RyanKung:
da7c5c02 addresses the five items of the acceptance check:

1. **Unspent SURB on a too-wide value.** `OnionSurb::produce` now returns `OnionProduceError::ValueTooWide { surb: Box<OnionSurb>, width }`: the block comes back unspent. `OnionSurb::capacity()` (the class's new `value_capacity() = C₀ − 1`, shared with `carry::seal`) lets a caller check first. A weak `σ_out` key is still `Key(KeyError)` and drops the block, since it can never produce.
2. **Provenance.** Every `test_cell` layer now has `next = 100 + i` and `x = 1000 + i`. The test asserts three things: admission (`peeled.head()`) reads position i's own `x`; the SURB carries position 2's `x = 1002` and `next = 102`; and a SURB handed back by a too-wide value is then spent.
3. **Wording.** The module and type docs now say "D8's 2979 B (`next`, `χ`, `σ`, `x`) plus the class byte, 2980 B".
4. **Diagram.** The module diagram shows `Relayed(head, OnionCell)` / `Consumed(head, v, OnionSurb)`, and the text says the step consumes both seeds and yields the key-free head.
5. **`encode_sec1`.** Its precondition now names every caller: the normalised point of a `NonIdentityPoint`, the affine point of a k256 public or verifying key, or an affine point just checked against `O`. It also states that `O` would encode as 33 zero bytes and that no caller can pass it.

**Gates:** fmt, typos, clippy (native for node/core/aez, node lib, wasm tests, ffi, core dummy), rustdoc, node `onion::` (216) and core `ecc` (59) all pass. No line added by this PR exceeds 100 characters.

I will rebase onto develop for #846 (dedupe `loop_shape`, `OnionExitEpoch` → `OnionProcessEpoch`, move `to_bytes`) only when the dispatcher says to.

RyanKung:
f5324b7d brings #845 up to date with `develop` (808f9ea9, which includes #844), as a signed merge commit. I merged instead of rebasing so that no force-push was needed; a squash merge keeps `develop` linear.

- **Conflicts.** The only conflicts were in `crates/node/Cargo.toml` and `Cargo.lock`.
  - `crates/node/Cargo.toml`: `zeroize = { workspace = true, features = ["zeroize_derive"] }`. This uses the workspace dependency #844 introduced; the derive feature is named explicitly because sphinx derives `Zeroize`/`ZeroizeOnDrop`.
  - `Cargo.lock`: taken from `develop` and regenerated, so the only additions are this PR's node dependencies.
- **Silent breakage check.** Nothing breaks.
  - #844's `circuit/admission`, `circuit/expiry.rs` and `circuit/mod.rs` changes share no item with `onion::sphinx`: no duplicate accessor on `OnionForwardNonce`/`OnionExitEpoch`, and no use of `OnionCellBucket::ALL`, `world_facing`, the Σ codec or the loop constants.
  - Both scoped `expect(dead_code)` attributes still hold.
- **#846 checklist not applied.** As agreed, the constants dedupe, the `OnionProcessEpoch` rename and moving `to_bytes` stay with #846, which merges after this.

**Gates on the merged tree:** all green.
- fmt, taplo, typos;
- clippy: native (node/core/aez), node lib, wasm (node tests and lib, aez), ffi, core dummy, core wasm;
- `check --all-features` and rustdoc;
- tests: `rings-aez` 15; node `onion::` 245 (now including #844's admission tests); `rings-core` lib 749/749, with the two filed load-sensitive networking tests skipped.

**wasm32 bench:** a whole cell (parse → peel → step → encode) takes 621.1 µs, 1610 cells/s.

CI has not been rerun manually; the push started the regular run.

Issue #842, opened 2026-09-23
Onion Kleisli Phase 2a-3: process epochs for every node and guard-closed loop selection (#834)


RyanKung:
Part of #834, Phase 2a, sub-PR 3 of 4. The spec is the #834 body (D2, D4, D5, L7) and `papers/onion-kleisli-pipeline.tex` on `develop`.

## Scope

1. **Process epoch for every node (D2).**
   - The relay capability in `OnlineNodeDescriptor` carries the process epoch `e_n`, drawn afresh at each start. Symbol descriptors carry the same `e_n`.
   - Registering any symbol implies registering `relay`.
2. **Loop selection (D4, D5, L7).**
   - A pipeline of `n` symbols gets `H(n, s) = (s + 1)·n + s` positions, with `s = ONION_SEGMENT_RELAYS = 2`.
   - The guard is at positions 1 and `H` and nowhere else. `has_duplicate_dids` admits exactly the guard twice.
   - `MAX_ONION_LOOP_SYMBOLS = 4`, and `MAX_ONION_LOOP_HOPS` is derived (14).
   - Selection fails closed below `H − 1` distinct eligible DIDs.
   - Each position is taken from nodes registering its symbol.
3. **Removals.** Delete `DEFAULT_ONION_ROUTE_HOPS`, `OnionRouteRequest::{hop_count, allow_short_paths}` and `MAX_ONION_CIRCUIT_HOPS`: the loop length is fixed by the pipeline, and a short path would violate D5. Update callers (proxy config, CLI, config docs) in the same PR.

Descriptor encodings change, so this is part of the breaking cutover and there is no compatibility with pre-change descriptors. The data plane still consumes the selected route only through the Phase 1 API until 2a-4 switches it.

## Laws to discharge (tests)

- Loop shape: `H = (s+1)n + s`, exactly `s` relays between consecutive symbol hops, and the guard at 1 and `H` only.
- `n > 4` is rejected, and selection fails closed when too few DIDs exist.
- Epoch: a descriptor without an epoch, or with a stale one, is rejected.
- The symbol ⇒ relay implication is enforced at registration.

## Out of scope

Header and carry (2a-1), admission (2a-2), and data-plane integration (2a-4).

PR #846, opened 2026-09-23
Onion Kleisli Phase 2a-3: process epochs for every node and guard-closed loop selection (#842)

RyanKung:
Closes #842
Part of #834

## Summary

**Registration (D2)**
- `OnionExitEpoch` → `OnionProcessEpoch`, drawn afresh at every process start.
- `OnlineNodeDescriptor::capabilities` is now typed (`OnlineNodeCapabilities { onion_relay: Option<OnionProcessEpoch> }`). A relay registration without an epoch cannot be represented, and the `"onion-relay"` label is removed.
- Symbol descriptors carry the same `e_n` (one epoch per process, by construction in `ProcessorBuilder`).
- Symbol ⇒ relay is enforced in two places:
  - At registration, `advertise_onion_exit ⇒ advertise_onion_relay ⇒ advertise_presence` is rejected at config time. `enable_*_onion_exit` also registers relay.
  - At the reader, a symbol descriptor is eligible only while its `(DID, session key, e_n)` equals a live relay registration. This rejects descriptors whose epoch is missing or stale.

**Loop selection (D4, D5, L7)**
- New `onion::loop_shape`:
  - Constants: `ONION_SEGMENT_RELAYS = 2`, `MAX_ONION_LOOP_SYMBOLS = 4`, and `MAX_ONION_LOOP_HOPS = H(4, 2) = 14`, which is derived.
  - `OnionLoopShape` (n ∈ [1, 4]) computes the role of each position.
  - `OnionLoop<P>` stores the guard once, at positions 1 and H.
- `select_loop` draws in this order:
  1. the guard (under the entry-guard policy);
  2. each symbol hop, from its registrants — a Hall's-condition guard keeps the greedy draw complete;
  3. the relays.

  It fails closed below `H − 1` distinct eligible relays and rejects `n > 4`.
- `has_duplicate_dids` admits exactly the guard twice.
- `OnionRoute` holds the loop. Until 2a-4, the data plane still consumes the Phase 1 view `positions()` = `g, r₀,₂, h₁` (today's three-hop path).

**Removals**
- `DEFAULT_ONION_ROUTE_HOPS`, `OnionRouteRequest::{hop_count, allow_short_paths}` and `MAX_ONION_CIRCUIT_HOPS`.
- The matching config fields in proxy, HTTP CONNECT, native gateway and CLI, the browser `onionHttpsProxy` arguments, and the frontend controls/settings.
- Docs, `SECURITY.md` and the docker cluster entrypoint check are updated.

The descriptor encoding changes. This is part of the breaking #834 cutover, with no compatibility with pre-change descriptors. There is no version bump; that happens on the last PR of the cutover.

## Laws tested
- Loop shape for every n ∈ [1, 4]:
  - `H = (s+1)n + s`;
  - exactly `s` relay positions per segment, the guard counted;
  - the guard at 1 and H only;
  - `H − 1` distinct hops.
- `n > 4` is rejected by both the shape and selection. Selection fails closed with `NotEnoughLoopHops { required: 4, eligible: 3 }`, and a node without the relay capability does not count.
- Epoch:
  - At the directory boundary, descriptors with missing or stale relay epochs are rejected.
  - A route whose symbol hop epoch differs from its descriptor's is rejected.
  - An exit from an earlier process is not selected, end to end through `Processor`.
- Symbol ⇒ relay is rejected at config time; the relay capability carries a fresh epoch per start.
- The guard policy, quality weighting, the Hall-steered symbol draws and `has_duplicate_dids` are covered.

## Verification
- Native:
  - `cargo clippy --all --tests -D warnings`, the strict node lib, `--all-features` check and ffi clippy;
  - wasm clippy for the browser node and the frontend;
  - `cargo +nightly fmt`, rustdoc `-D warnings`, `taplo`, `typos`.
- Tests:
  - `rings-node` onion/processor/online/config tests.
  - The wasm browser tests for node onion and the frontend. The frontend's browser onion WebView e2e now runs a 5-node loop (client → guard → relay → exit).
- The ignored public-network gateway e2e tests were reworked to a 3-relay mesh but not run here, since they need public network access.


RyanKung:
Addressed the review in 08355db1, on top of 9fc034cb. The branch merges cleanly with develop; 800ea69f only touches `papers/`.

## Medium

**M1 — the browser API reports used hops only.**
- New `OnionRoute::circuit_hops()` returns the forward prefix `g, r₀,₂, exit` that the data plane uses until 2a-4. It is built on `OnionLoop::prefix_to_symbol` and does not allocate.
- `BrowserOnionRouteInfo.hops` and the frontend `display_route` now report it, so WorkBench shows 3 hops.
- The docs state that the return segment is selected but unused.

**M2 — visibility.**
- `OnionRoute::new` is back to `pub(crate)`.
- `loop_shape` is a private module. `OnionLoop`, `OnionLoopShape`, `OnionLoopStep` and the constants have a single public path: the `onion` root re-exports.

**M3 — one sum type for the role** (`onion/role.rs`).
- `OnionRole<X> = Client | Relay | Exit(X)` is a functor in `X`. The same rung is carried through each layer:
  - `OnionRole<OnionExitOffer>` in `ProcessorConfig`, `ProcessorBuilder` and `Processor`. `OnionExitOffer` is non-empty services under an open policy.
  - `OnionRole<NativeOnionTcpExitConfig>` at `NativeOnionCircuitHandle::install`.
  - `OnionCircuitCapabilities = OnionRole<OnionProcessEpoch>` in the reducer.
- The textual flags (yaml and CLI) are parsed exactly once, in `OnionRole::from_flags`. "Exit without relay" is rejected there and has no typed inhabitant.
- `accepts_forward_layers` is gone, because `registers_relay` covers it. `permits_exit_epoch` is now `permits_exit_layer`.
- In the browser, the circuit capabilities and the HTTPS exit policy both come from the processor role. `install_onion_https_exit()` takes no policy and rejects a provider that is not configured as an HTTPS exit. This closes the reachable exit-without-relay state you pointed out.

**M4 — removed keys and env vars.**
- `gateway.onion_hop_count` and `gateway.onion_allow_short_paths` are rejected by name when the gateway section is deserialized. The error points at the loop shape.
- `rings run` rejects a set `ONION_HTTP_PROXY_HOP_COUNT` or `ONION_HTTP_PROXY_ALLOW_SHORT_PATHS`. Both cases have tests.

**M5 — no parallel array, no lookup errors.**
- `OnionLoop::try_unfold` is now the only loop constructor. It unfolds the pipeline along its list zipper and gives each position its pending symbols.
- `select_loop` draws every position directly as an `OnionRouteHop`, with Hall's filter over the pending symbols.
- The route draws the exit first, among exits that some permitted guard can precede. It then unfolds the loop around that exit, so there are no `hops.symbol(1)` or `exits.find` lookups.
- Draw failures are total by position:
  - a guard draw fails with `NoPermittedFirstHop`, or `NoDistinctSymbolHops` when Hall's condition fails;
  - a relay draw fails with `NotEnoughLoopHops`, which is the fail-closed case;
  - a symbol draw fails with `NoDistinctSymbolHops`.
- `RouteHasNoHops` and `HopCountOutOfBounds` (and its bound in `edge_circuit_ids_with`) are deleted.

## Low
1. New `StaleExitRegistration` and `ExitWithoutRelayRegistration` errors, each tested. The D2 match moved from candidate collection into selection so that each rejection names its cause.
2. Selection tests use seeded keys (`[seed; 32]`), so DID-ordered draws are reproducible.
   - New pinned test: the sole symbol registrant is first in DID order and a permitted guard, and the Hall filter still keeps it out of the guard.
   - The frontend e2e drops its fixed 1 s sleep and uses `OnionRole<&str>` instead of its own enum.
3. The Hall check folds subfamilies as (size, union) pairs, so there is no `1 << len`, no `count_ones() as usize` and no per-candidate `taken.clone()`. Exclusion is a predicate.
4. `circuit_hops` does not allocate. `positions()` is built once per circuit in `assign_edges`. `route_first_hop` reads `hops().guard()`. `word()` is derived from `circuit_hops`.
5. New typed plain-mirror DTO: `OnlineNodeCapabilitiesInfo { onion_relay: Option<Value> }`.
6. Removed the no-op `BTreeMap` collect in `eligible_relays`. I kept the re-validation of signature, liveness, local protocol mode and newest-per-DID: the browser directory can be a remote node, whose filtering is not trusted locally. The candidates doc now says so.
7. Not fixed here, as you asked; I'll file it as its own issue.

## Gates
- clippy `-D warnings`:
  - native (`--all --tests` and the strict `node` lib);
  - wasm (`browser_default` tests and the `browser` lib);
  - ffi;
  - the frontend.
- `cargo check --all-features`, nightly fmt (workspace and frontend), rustdoc `-Dwarnings`, taplo, typos, and rings-core `dummy` clippy.
- Tests:
  - `rings-node` onion/processor/online/native/rpc: 301 passed. CLI: 8 passed.
  - wasm node onion tests.
  - The frontend onion tests, including the 5-node browser loop e2e (11 passed).

RyanKung:
Round 2 is addressed in c947f498, a fast-forward on 08355db1.

## Medium

**N1 — Hall's condition counts only drawable DIDs.**
- `select_loop` is generic over symbol registrations `X` with a projection `X → OnionRouteHop`.
- At entry, it keeps only candidates whose hop is one of `relays` (DID, session key and epoch), so every DID that Hall's condition counts can actually be drawn.
- The new test `test_hall_filter_ignores_registrants_outside_the_relays` pins your example: `S = {x, a}` with `x ∉ R`, where `a` is first in DID order and a permitted guard. `a` is kept for the symbol, and `x` never appears.

**N2 — the native install derives its role from the processor.**
- `NativeOnionCircuitHandle::install(extensions, key, network_id)` no longer takes a role.
- It reads `Core::onion_circuit_capabilities()` (the processor role at its epoch; this is now the only place `role.as_ref().map(|_| epoch)` is written) and, on the exit rung, builds `NativeOnionTcpExitConfig::from_offer`.
- The gateway fixture now builds its processors with their roles via `prepare_processor_with_onion_role`.

**N3 — browser roles.**
- Installing the browser runtime rejects any exit offer containing a service other than `https`.
- `new_browser_provider_with_storage` installs the circuit protocol whenever `registers_relay()`; the HTTPS-exit constructor already did.
- `install_onion_https_exit` is replaced by `install_onion_runtime()` for providers built around an existing processor.
- A new wasm test covers the three roles. The frontend e2e workaround is gone. With this fix, round-1 L7 is fixed in this PR.

## Low

1. **One mechanism.** The exit is no longer pre-filtered: it is drawn at its own loop position by the same Hall-filtered draw as every other position.
   - This works because `OnionLoop<P, T>` now stores the terminal symbol as a field (`lead`, `stages`, `terminal`, `tail`), and `try_unfold` takes a non-empty `OnionPipelineSymbols` and threads the draw state through two label functions.
   - The route reads the drawn exit back totally, with `project_symbols` and no lookup.
2. **Naming.** `StaleExitRegistration` is renamed to `ExitRelayRegistrationMismatch`; its doc and message now state the whole-hop mismatch (session key or process epoch).
3. **Structural views.**
   - `OnionRoute::positions` is renamed to `forward_path`, and is built from `forward_prefix()` plus `terminal()` with no DID search.
   - `circuit_hops` is structural.
   - `word()` counts `forward_prefix()` and no longer uses `saturating_sub`.
4. **Service set.** `OnionExitOffer` stores a `BTreeSet<OnionServiceName>`. `from_offer` copies that set, so it no longer repeats `new`'s dedupe.
5. **Unknown gateway keys.** Every key outside `GATEWAY_KEYS` is rejected. `GATEWAY_KEYS` is test-pinned to equal the keys `rings init` writes. The removed keys keep their loop-shape message.
6. **Traversal.** Added `OnionRole::try_map`, the traversal in `Result`, and the frontend fixture uses it.
7. **Browser e2e.** The fixture now waits for its precondition: the client directory must hold the relay registrations of all 4 relays and the `https` exit. After that it navigates exactly once, and any route error fails the test.
   - The DHT gives a browser no change notification, so the directory itself is re-read every 250 ms.
   - The fixed 1 s sleep and the retried navigation are both gone.
8. **Phase 2b.** `role.rs` now notes that invoke-only (intermediate-symbol) registrants will need a new rung in 2b.
9. **PR body.** Refreshed.
10. **Small items.**
    - The shape is computed only inside the unfold; the `required` count comes from the cursor.
    - Hall's check is now an augmenting-path matching (Kuhn), with no 2ⁿ set clones.
    - Relays are sorted by DID at entry, so the weight test's DID-order claim is now true and not a seed coincidence.

## Gates
- clippy `-D warnings`:
  - native (`--all --tests` and the strict `node` lib);
  - wasm (`browser_default` tests and the `browser` lib);
  - ffi;
  - the frontend.
- `cargo check --all-features`, nightly fmt (workspace and frontend), rustdoc `-Dwarnings`, taplo, typos, and rings-core `dummy` clippy.
- Tests:
  - rings-node onion/processor/online/native/rpc: 304 passed. CLI: 8 passed.
  - wasm node onion tests: 4 passed, including the new role test.
  - frontend onion tests: 11 passed, including the 5-node browser loop e2e.

RyanKung:
Round 3 and the red CI run 35974765853 are addressed in 61acd4d2, a fast-forward on c947f498. I have not rerun CI.

## Medium — draw order (#834 L7)
`select_loop` now separates the draw order from the position order:
1. It draws the symbol positions first, in pipeline order, each from its admitted registrants.
2. Then it draws the guard, from the permitted relays.
3. The relays are drawn last, while `OnionLoop::try_unfold` places the drawn labels.

Every draw is filtered by a Kuhn matching over all still-pending positions, and the guard is one of those positions, so drawing symbols first can never strand the guard. `try_unfold` now only places the labels: the pending/cursor types are gone, and the drawn symbols are passed in as the pipeline.

`test_exit_and_guard_marginals_follow_the_draw_order` pins your example exactly: with R = {x (Healthy), a, b, c}, E′ = {x, a}, and both rolls enumerated over [0, 48)², it asserts P(exit = x) = 1536/2304 = 2/3 and P(guard = x) = 384/2304 = 1/6.

## Low
1. `NativeOnionCircuitHandle::install(extensions)` reads the delegatee key, network, capabilities and exit offer from `Core`, and takes no other arguments.
2. `NativeOnionTcpExitConfig` is deleted. The TCP runtime holds the `OnionExitOffer` (a set of services), and the dead `set_native_proxy` branch and setter are removed.
   - One pre-existing gap remains: `OnionHttpsRuntime::native_proxy` now has no writer, so its egress proxy path is unreachable. It had no configuration path before this PR either. I'm flagging it as an issue candidate rather than widening this PR.
3. **Browser exits.**
   - A named predicate, `OnionExitOffer::uninterpretable_service` (with `runtime_interprets`), feeds a precise `Error::UninterpretableOnionService { service }` (code 813).
   - It is checked in `validate_onion_role_config` at build time, which covers every entrance, and in `new_browser_provider_with_storage` before any IndexedDB or processor side effect.
   - `test_browser_constructor_admits_only_roles_it_can_interpret` asserts the error variant on the production constructor. It also checks that the constructor installs the protocol at start for relay and `https`-exit roles.
   - The `from_processor` window is closed structurally: browser `listen()` installs the runtime of any relay-registering role before its first registration (`test_browser_listen_installs_the_runtime_of_a_relay`). `install_onion_runtime` is removed.
4. **E2E wait** (this was also the Firefox timeout). The listeners were started before any edge existed, so every node's first registration missed the overlay and the directory only converged on the next 30 s heartbeat. On top of that, five serial WebRTC handshakes pushed Firefox past the 120 s budget. The fixture now:
   - opens all edges concurrently;
   - waits until every edge reports `Connected`;
   - starts the listeners only after that, so each first registration publishes over live edges;
   - waits on the directory precondition, then navigates once.

   Both waits share one documented, bounded read-back (`poll_until`, upper bound 20 s), because the browser has no notification for an edge opening or a DHT entry converging. Chrome now takes 13 s for this test (was 20–40 s), and the full frontend suite passes 48/48.

   I could not run Firefox here: geckodriver fails under the macOS app sandbox (`sandbox_extension_issue_file_to_process … Operation not permitted`), so Firefox is confirmed only by CI.
5. `OnionLoop` no longer stores `shape`; it is derived from `stages.len() + 1`. The overclaim is corrected: the only constructor emits the segment lengths.
   - `lead`, `tail` and the stage relays stay `Vec`. Filling a `[P; N]` from a fallible labelling needs `array::try_from_fn`, which is unstable, so every stable route adds an unreachable length-mismatch branch.
6. `OnionLoopRelay`, `OnionPipelineSymbols` and `try_unfold` are `pub(crate)`. The pending and cursor types are deleted. `OnionLoop`/`OnionLoopShape` stay public because `OnionRoute::hops()` returns them.
7. D2 admission runs once: `RelayRegistrants::admit` is the only constructor of `SymbolRegistrants`, and it uses a binary search over the DID-ordered, deduplicated relays. Draws happen on references, and only the winner is copied.
8. D5 is checked before any guard logic: `test_fail_closed_is_checked_before_the_guard_policy` gets `NotEnoughLoopHops` even with a deny-all guard policy.
9. `role.rs` now cites #834 D1′: `invoke⟨In, Out, W, L⟩` shapes paired with an open backend name.

## CI
- `qaci.yml` now names `captured_tcp_reaches_public_http_only_through_onion_loop` and `real_linux_tun_tcp_reaches_public_http_through_onion_loop`, and the step titles are updated.
- I ran the captured-TCP test locally against the public exit. It first exposed a fixture bug: stabilization auto-connects the relay mesh, so a later explicit offer failed with `AlreadyConnected`. `connect_gateway_edge` now accepts an edge the overlay already opened. The test then passed through a real 5-processor loop in 17 s.
- The Linux TUN variant shares the same fixture but needs Linux and root, so only CI can confirm it.

## Gates
- clippy `-D warnings`:
  - native (`--all --tests` and the strict `node` lib);
  - wasm (`browser_default` tests and the `browser` lib);
  - ffi;
  - the frontend.
- `cargo check --all-features`, nightly fmt (workspace and frontend), rustdoc `-Dwarnings`, taplo, typos, and rings-core `dummy` clippy.
- Tests:
  - rings-node onion/processor/online/native/rpc: 304 passed. CLI: 8 passed.
  - wasm node onion tests: 5 passed.
  - frontend: 48 passed in Chrome.
  - The captured-TCP gateway e2e passed.

## Merge coordination with #845
Noted: whichever of #845 and #846 merges second applies your checklist:
- `sphinx/mod.rs` imports the loop constants from `loop_shape`;
- the `OnionExitEpoch` uses are renamed to `OnionProcessEpoch`;
- the textual conflict in `onion/mod.rs` is resolved.

RyanKung:
Round 4 addressed in 01868d16.

1. **Exit marginal**: the route module doc, the `select_loop` doc and SECURITY.md now state the quality-share marginal only "when every registered exit extends to a loop", as the paper does. They also say that the Hall exclusion conditions the marginal otherwise.
2. **Unreachable `ok_or`s**: these are replaced by `Error::LoopDrawInvariant`. An empty draw after the D5/Hall admission is an invariant breach, and it is typed as one.
3. **Guard source**: the guard is drawn from a single `guards ∖ T` source.
4. **Clones and cursor coupling**:
   - `OnionLoop::try_unfold(intermediate, terminal, label_relay)` takes the pre-drawn symbols and labels only the relays.
   - The symbols are drawn by `OnionPipelineSymbols::try_map_with_later`, which hands each symbol the symbols after it (the Hall pending set) through `iter::successors` over suffixes.
   - No shared cursor and no `LoopDraw` state are left.
5. **One registration per DID**: `RelayRegistrants::admit` dedupes by DID, so the directory holds at most one registration per DID.
6. **Installs**:
   - Browser `listen()` installs the onion runtime synchronously before spawning, and the task moves only the `Result`, with no `Provider` clone.
   - Native `Provider::listen`/`listen_with` install `NativeOnionCircuitHandle` for a role that registers `relay` when no runtime is installed yet, and now return `Result<()>`. The callers are updated: FFI, the tests and `examples/native`.
   - New test: `test_provider_listen_installs_the_onion_runtime_of_relay_roles`.
7. **`runtime_interprets`** is `pub(crate)`.
8. **E2E waits**: the frontend e2e waits for each edge to be `Connected` on both sides (`edge_connected`).
9. **`native_proxy`**: the field, the getter, the proxy egress branch and their tests are deleted.
   - `NativeHttpsEgress` is now the pinned `{ host, addresses }` struct.
   - Every non-public resolution is denied, including the fake-IP range.
10. **Vec segments**: kept, as documented.

The pre-existing "first registration before any edge" issue is left for its own issue, as agreed.

Local gates:
- clippy passes on native (`--all --tests`), wasm, ffi and the frontend;
- nightly fmt, native rustdoc `-D warnings`, taplo, typos and the core dummy build pass.

Local tests:
- native lib onion/processor/online/native/provider tests pass, as do the CLI tests, the wasm node lib tests (Chrome) and the frontend (Chrome, 48 tests);
- one unrelated test failed once in the full run and passed when rerun alone: `test_processor_e2e_message_streams_and_decrypts_with_receiver_identity_key` (E2E stream frames, untouched here).

RyanKung:
Round 5 addressed in 73cf68af.

**Rule violations**
1. **Lines over 100 characters**: every Rust line this PR added over 100 characters is split. The string literals use `concat!` or `\`-continuation, the error-mapping diagram in the `select_route` doc is narrowed, and one doc line is rewrapped. The files are `bin/rings.rs`, `native/config.rs`, `failure.rs`, `loop_shape.rs`, `role.rs` and `route/mod.rs`.
2. **Functions over 100 lines**:
   - `run_browser_onion_webview_flow`: the fixture topology is extracted into `browser_onion_loop()`, which builds the nodes, connects the edges, starts the listeners and waits for the loop directory. It returns the client with its listeners.
   - `OnionRouteError`'s Display: instead of hand-splitting `fmt` by family, the enum now derives it with `thiserror::Error` (already a dependency), one `#[error]` per variant. The match stays exhaustive without a catch-all, and no `fmt` body remains. `OnionExitFailure` is unchanged.
   - The only other touched functions over 100 lines (`foreground_run`, `run`, `test_two_relays_peel_…`) were already over 100 lines on develop; this PR shortens `foreground_run`.

**Optional Lows**
3. **Marginal claim**: the quality-share marginal is claimed for `h₁` only (in the module doc and `select_loop`). The later symbols are drawn conditionally on `T`, so their marginals are not claimed.
4. **Unreachability proof**: `select_loop` now states the premises and the steps of the no-empty-draw argument:
   - P1: `Sₖ, G ⊆ R`, one hop per DID (D2 admission);
   - P2: the matching of `(S₁ … Sₙ, G; ∅)`;
   - P3: `|R| ≥ H − 1` (D5);
   - I, the matching invariant carried through each symbol draw, and from it the guard step and the relay count `|R ∖ T| ≥ H − 2 − n`.
5. **Borrowed DIDs**: `SymbolRegistrants` carries its DIDs, projected once in `admit`. `PendingPositions<'_, X>` views the registrants directly, so no `dids()` vectors are rebuilt per draw.
6. **Dedupe test**: `test_registrants_keep_one_registration_per_did` covers `RelayRegistrants::new` and `admit`: the first registration of each DID is kept and a stale hop is not admitted.
7. **Native install**: the check-and-install now runs under a lock shared by the provider's clones, so concurrent listens install once. The doc states the scope: this provider's registry only; another provider over the same processor has its own registry, and ordering a direct `NativeOnionCircuitHandle::install` against a concurrent listen is the caller's job.
8. **Listener flowchart**: the browser listener flowchart now shows the synchronous install and the rejected task when it fails.

Local gates:
- clippy passes on native, wasm, ffi and the frontend;
- nightly fmt, native rustdoc `-D warnings`, taplo, typos and the core dummy build pass.

Local tests:
- the native lib tests (onion/processor/online/native/provider, 326) and the CLI tests pass;
- the wasm node lib tests pass in Chrome, as does the frontend (48 tests).

RyanKung:
Merged develop (with #844) in 6ac71876:
- The textual conflict in `test_circuit_protocol.rs` is resolved in favour of develop's exit-window boundary laws (`exit_layer_is_admitted`). The helper's reducer is built from `OnionCircuitCapabilities::Exit(TEST_PROCESS_EPOCH)`, since `from_registration` no longer exists here.
- The admission module and its tests from #844 now use `OnionProcessEpoch` (`admission/mod.rs`, `tests/{mod,links,window}.rs`). No `OnionExitEpoch` or `exit_epoch` remains.

Local gates:
- clippy passes on native, wasm, ffi and the frontend;
- nightly fmt, native rustdoc `-D warnings`, taplo, typos and the core dummy build pass.

Local tests:
- the native lib tests (onion/processor/online/native/provider, 355) and the CLI tests pass;
- the wasm node lib tests pass in Chrome, as does the frontend (48 tests).

**Firefox failure at 73cf68af.** The failure is not in the explicit connect handshake. The error prefix `entry action error:` is `Error::EntryError`, which only DHT storage operations produce. It comes from the client-directory poll (`lookupOnlineNodes`/`lookupOnionExits` → `Swarm::storage_fetch`): the fetch was sent on a connection generation that stabilization or glare replaced.

`ConnectionAttemptSuperseded` is classified as a deferrable data-plane send in `core/error/policy.rs`, and storage sync and repair defer on it, but the user-facing `storage_fetch` returns it to the caller. This is a product-level question, and the fix is pending a decision; it is not fixed in this push.

RyanKung:
Merged develop (with #845) in 08d946d8 and applied the #845 merge checklist:
- **`onion/mod.rs` conflict**: resolved. `to_bytes` now lives on `OnionProcessEpoch`.
- **Loop constants**: the duplicate `ONION_SEGMENT_RELAYS`, `MAX_ONION_LOOP_SYMBOLS` and `MAX_ONION_LOOP_HOPS` are deleted from `sphinx/mod.rs`. `sphinx` (header, seed and its tests) imports them from `onion::loop_shape`. The L5′ note ("exactly `Ĥ` layer slots whatever the loop's own `H`") moved onto `ONION_HEADER_ROUTING_BYTES`.
- **Rename**: `OnionExitEpoch` is now `OnionProcessEpoch` in `sphinx/layer.rs`, `sphinx/tests/mod.rs` and `test_cell.rs`, with no alias. No `OnionExitEpoch`, `exit_epoch` or `onion_exit_epoch` remains anywhere; `registry.rs` has no such name after the merge.

Local gates:
- clippy passes on native, wasm, ffi and the frontend;
- nightly fmt, native rustdoc `-D warnings`, taplo, typos and the core dummy build pass.

Local tests:
- the native lib tests (onion/processor/online/native/provider, 377) and the CLI tests pass;
- the wasm node lib tests pass in Chrome.

**Browser e2e is flaky in Chrome, and this predates #845.** `test_webview_node_fetches_page_resources_through_browser_onion_exit` does not pass reliably:

| Commit | Mode | Result | Observed failure |
|---|---|---|---|
| `08d946d8` (head) | full frontend suite | 2 fails in 4 runs | `OnionRequestTimedOut` during the WebView fetch |
| `08d946d8` (head) | test alone | 5/5 pass | none |
| `6ac71876` (pre-#845) | test alone | 1 fail in 5 runs | `NotEnoughLoopHops { required: 4, eligible: 3 }` right after the fixture's directory poll had counted 4 relay registrants |

Besides the `storage_fetch` superseded-send failure tracked in #859/#860, this gives two further failure modes:
- The fixture's directory poll and the route selection each fetch the online-node registry. The second fetch can see fewer registrants than the first, so the directory is eventually consistent across fetches.
- An onion request times out mid-flow.

I have not changed the fixture; the details are in the dispatcher thread.

RyanKung:
Pushed 532d1e1b and 05554e83.

**532d1e1b: #864, the NotEnoughLoopHops flake, fixed at the root.**

*Cause.* A directory read returns whichever replica answered last: `fetch_storage_entry` answers from the cache, and every `FoundEntry` reply replaces the cache. So two reads of the online-node or exit registry can disagree while ownership moves, and a loop precondition observed once (`|R| ≥ H − 1`) could vanish before selection.

*Fix.*
- `descriptor::DescriptorView` joins every read into a per-process view: the newest live descriptor per key under the heartbeat order, dropped at expiry. The join is commutative, associative and idempotent, so reads are monotone up to expiry.
- `Processor::lookup_online_nodes` / `lookup_onion_exits` answer from this view, which every clone shares. The fixture's readiness poll and the WebView's route selection therefore observe one growing snapshot. The fixture's predicate is unchanged and is now sound.
- There is no retry on `NotEnoughLoopHops`, and D5 stays fail-closed (reasoning in #864).
- `latest_valid_by_key` generalizes `latest_valid_by_did`. The exit registry keys by `OnionExitDescriptor::registration_key` through it, which removes a copy of the newest-per-key loop.
- New law tests cover: monotone reads, the newest heartbeat winning in either read order, and expiry.
- SECURITY.md states the view, and that a withdrawn registration stays a candidate until its TTL.

**05554e83: merged develop with #863** (the #856 IndexedDB LRU fix).

Local gates:
- clippy passes on native, wasm, ffi and the frontend;
- nightly fmt, native rustdoc `-D warnings`, taplo, typos and the core dummy build pass.

Local tests:
- native lib tests (380), the CLI tests, the wasm node lib tests and the core IndexedDB tests pass (Chrome);
- the frontend full suite passes, 48/48;
- with the view, the browser e2e passed 8 of 8 single runs.

**Still open.** In one of 6 full-suite runs before the develop merge, the fixture's readiness poll timed out after 20 s: the view never reached all four relay registrations plus the exit. This is a different mechanism: a registration was not visible on any replica the client read within the window.

This and the earlier `OnionRequestTimedOut` are being diagnosed with debug logs; neither is fixed here. The `storage_fetch` superseded failure stays with #860.

RyanKung:
Review of 05554e83 addressed in bb294cda. Per the user's decision on H1, the join now lives in core; the #864 body was rewritten to that design before the code.

**H1: join in core.**
- `local_cache_put` now does `cache[k] ← cache[k] ⊔ reply` with `Entry::join`, under the ring's `storage_transition`. Two concurrent `FoundEntry` replies therefore cannot lose one side.
- It shares `join_live_entry` with `join_storage_entry`, so both paths use one read-join-write. `local_cache_get` now takes the transition too, because its lazy retirement is a write.
- The laws are stated on `ChordStorageCache`:
  - the join is commutative, associative and idempotent, so the cached value does not depend on reply order;
  - tombstones are kept;
  - every reply is admitted like a replicated write, carriers are capped at `max_data_len` (oldest dropped), and the cache is bounded at `LOCAL_CACHE_CAPACITY`;
  - monotonicity holds only per resident key: capacity eviction, carrier expiry and a restart un-observe a key.
- `DescriptorView` is deleted, and the lookups read the joined carrier. The view's unbounded growth, signer-chosen eviction clocks and `u128::MAX` pinning are gone with it.

**M1: equal-heartbeat ties.**
- The carrier is order-independent by the CRDT laws. `test_local_cache_joins_replies_in_any_order` feeds both reply orders, including two writers stamped at the same instant.
- At the node, `latest_valid_by_key` now breaks an equal heartbeat by `(expiry, signed body, signature)`, a total order. `test_latest_valid_by_did_breaks_an_equal_heartbeat_by_content` checks both read orders, at a different expiry and for two session keys of one account at the same millisecond.

**M2: failed fetch.** `fetch_storage_entry_with_stop` treats a failed fetch as the empty read: it answers from the live cache and returns the error only when there is none. A processor-level test would need a fault-injection seam into `storage_fetch`, and none exists; I did not add a one-use seam for it.

**M3, L2, L3.** These were properties of the deleted view.
- No per-view re-verification remains.
- What is left is the pre-#864 pattern: the lookup verifies once, and `eligible_*` or the directory verifies again. That is unchanged by this PR.
- Delegation liveness at wall-clock time is likewise the pre-existing `latest_valid` behaviour, and no view law depends on it now.

**M4: fixture waits.**
- Edges are now event-driven: the fixture installs a `BackendObserver` per node and awaits `peer_admitted` on both ends of every edge. The `listPeers` polling is gone.
- A registration publish has no completion signal, so the registry wait stays a bounded poll. #866 records the missing signal and proposes a store report with a first-publish future.

**Lows.**
- L1: SECURITY.md now describes the cache join, what un-observes a key (eviction, bounded retention, restart), the failed-fetch rule and the remote-directory case, and the long lines are re-wrapped.
- L4: `OnionPipeline::hop_count` returns `NonZeroUsize`, and `edge_circuit_ids_with` takes it, so a zero count is unrepresentable.
- L5: new cache law tests, plus the tie tests above. The exit `(DID, service)` key keeps its existing test.
- L6: the Docker README says 4 nodes for an outside client and 5 when a cluster node is the client.

**Sweep of cache readers.** The registration observation, `fetchTopicMessages`, the browser cache reads and the exit refresh path have no replace-semantics assumption. One core test asserted cache == the raw reply; it now expects the storage-normalized carrier.

Local gates: clippy passes on native, core wasm, node wasm, ffi and the frontend; nightly fmt, rustdoc `-D warnings` (core and node), taplo, typos and the core dummy build pass.

Local tests:
- core `dht::` and storage handlers (270), node lib (385) and CLI pass;
- node wasm passes in Chrome;
- the frontend passed 48/48 in 3 full-suite runs.

In the full core lib run, `test_handle_connect_node` failed once on its wait deadline and passed alone; it does not touch the cache.

**Pre-existing, not from this PR.** Core wasm `test_storage_repair_load_does_not_starve_three_node_stabilization` exceeds its 60 s budget locally in Chrome, both at this head and at the previous head 05554e83 (without the cache join).

#846 still waits for #865.

RyanKung:
Round 2 (bb294cda) addressed in 275a6cf3.

**H1: compaction floors.** I verified the mechanism in code:
- `should_compact = … || !stale_values.is_empty()` in the registry publisher;
- `compact_data` stamps a floor above `max_observed_version`;
- `DataTopicBuffer::new` retains only `dot ≥ register`, and the join takes `register = max`.

What changed:
1. **The laws are now stated precisely**, in `ChordStorageCache`, SECURITY.md and #864:
   - The carrier ascends in ⊑ while it stays resident.
   - The *visible* elements shrink under tombstones, overwrite and compaction floors, and the element cap.
   - The registry publisher compacts on every heartbeat. So a floor from an owner that never held a descriptor erases that descriptor from the join, stickily, until the node heartbeats again.
   - "readiness and selection share one growing snapshot" is no longer claimed.
2. **Filed #867 before any code**: "Data-topic compaction floors erase concurrent adds the compacting owner never saw". It carries the mechanism, the trace and two directions: causally stable compaction, or registration that does not compact on every heartbeat. Its placement is with the user; it is core DHT semantics and likely affects master.
3. **#864 is no longer claimed closed.** The #864 body has a Status section saying it is not closed at the root until #867 lands, and the PR body now reads `Part of #864`.
   - Note for merge: commit 532d1e1b says `Closes #864` and is already pushed. Drop that line from the squash message.

**M1: cache lock.** The cache has its own `cache_transition`. The transition law is now stated per store: one lock per store, never nested, no path holds both. A directory read no longer waits behind an IDB scan or a repair pass.

**M2: tie tests.** New tests, each run in both reply orders with repeats:
- `test_local_cache_breaks_a_same_instant_add_tie_by_dot`: the same payload added by two writers at an equal `now`, with distinct dots asserted; the cache keeps the greater dot.
- `test_local_cache_breaks_a_same_instant_overwrite_tie_by_register`: two same-instant `Overwrite`s from different writers, with distinct registers asserted; the cache keeps the greater register's value.
- The distinct-payload test is renamed in its doc as a union test, not a tie test.

**Lows.**
- L1: the empty-read rule now goes through one helper, `answer_failed_fetch`. The exit refresh path (`fetch_storage_entry_after_cache_refresh`) uses it too, and the doc on `fetch_storage_entry_with_stop` states that the rule also serves the registration publisher, and why that is sound.
- L2: `storage.rs` and the SECURITY.md paragraphs are re-wrapped. No new line exceeds 100 columns.
- L3: not done. Sharing `Rendezvous` with the frontend, or adding `Provider::set_backend_observed_by`, would widen the public API for a test fixture. Under the issue-first rule that needs its own issue; I can file it if you want the fixture on a shared rendezvous.
- L4: `OnionSymbolWord::hop_count` now returns `NonZeroUsize` too.
- L5: `test_found_entry_replies_join_in_the_cache` delivers two `FoundEntry` replies with disjoint adds, in both orders, through `MessageHandler::handle`, and reads `storage_check_cache`. The `dummy`-gated `test_found_entry_read_repair_backpressure_is_deferred` had the same replace-semantics assertion and now expects the normalized carrier.

Local gates:
- clippy passes on native, core wasm, node wasm, ffi and the frontend;
- nightly fmt, rustdoc `-D warnings` (core and node), taplo, typos and core dummy clippy pass.

Local tests:
- core `dht::`/storage (273), the full core `dummy` lib (956), node lib (385) and the CLI pass;
- node wasm passes in Chrome;
- the frontend passes 48/48.

#846 still waits for #865.

